Issue #23, opened 2021-01-06
Create other higher level constructs


jayair:
Create higher level constructs to make it easier to build backends. Things like:

- DynamoDB
- SNS (events)
- SQS (queues)
- Cron

jayair:
Closed in #76

PR #76, opened 2021-01-29
Add Table, Topic, Queue and Cron Constructs

fwang:
### Docs
- [x] Added timeout and memorySize to sst.FunctionProps, with default timeout 10s and memorySize 1024MB


jayair:
Closes #23

PR #82, opened 2021-02-03
Change SST to not use the forked CDK

fwang:
The following SST/internal commands need to be migrated over:
- [x] sst build
- [x] sst deploy
- [x] sst remove
- [x] sst add-cdk
- [x] bootstrap


jayair:
Hi all, help us test this version so we can merge it into master soon.

In your `package.json`, update to this version using:

```
"@serverless-stack/cli": "0.8.1-next.15",
"@serverless-stack/resources": "0.8.1-next.15",
```

Give it a try and post any issues that you run into.

Thank you!

deini:
This may be a silly question but could `aws-cdk` and friends be `peerDependencies` and let the user pick the versions?

fwang:
> This may be a silly question but could `aws-cdk` and friends be `peerDependencies` and let the user pick the versions?

Hey @deini that's a totally legit question. CDK actually is going to change the way they structure the modules in v2, and we will be able to move to the `peerDependencies` approach.

olup:
Yep, just tryed `-next.15`, works great with the flag ! Thanks a lot.

justindra:
Does this also allow the SST to not be tied to a particular CDK version for the resources?

fwang:
@justindra Currently, it will still be. But after CDK v2 comes out, this should be possible.

That said, with forked CDK removed, it is now much easier for SST to be kept up to date with the newest CDK versions.

PR #144, opened 2021-02-24
Fix attachPermission instanceof error

### LINKED ISSUE #139: Fails to attachPermission with error 'The specified permissions is not a supported.'
This code inside `permission.js` is flawed:
```
  else if (permission instanceof iam.PolicyStatement) {
    role.addToPolicy(permission);
  }
```
This check fails if the `iam` modules imported by SST and by user are not the same module. Ie. one version of `@aws-cdk/aws-iam` found inside `node_modules`, and one version found inside `node_modules/@serverless-stack/serverless-stack/node_modules/@aws-cdk/aws-iam`.

Could check `permission.constructor.name`, but even that is not perfect.

jayair:
Closes #139

@fwang Review the changes

### DIFF
--- a/packages/resources/src/util/permission.ts
+++ b/packages/resources/src/util/permission.ts
@@ -55,8 +55,19 @@ export function attachPermissionsToRole(
       }
 
       // Case: iam.PolicyStatement
-      else if (permission instanceof iam.PolicyStatement) {
-        role.addToPolicy(permission);
+      else if (permission.constructor?.name === "PolicyStatement") {
+        const policyPermission = permission as iam.PolicyStatement;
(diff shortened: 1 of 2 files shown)


packages/resources/src/util/permission.ts
```diff
@@ -104,6 +115,7 @@ export function attachPermissionsToRole(
       }
       // Case: grant method
       else if (
+        Array.isArray(permission) &&
         permission.length === 2 &&
         permission[0] instanceof cdk.Construct &&
```
fwang, line 120: @jayair This `instanceof` needs to be changed as well

PR #195, opened 2021-03-05
Support for JSII

jayair:
- [x] Globally install SST `$ npm install -g sst@npm:@serverless-stack/cli`
- [ ] Create Python Lambda sandbox
- [ ] Support other runtimes in `sst.Function`
- [ ] Setup jsii for resources
- [ ] Setup build script to publish to other packages
- [ ] Wrap around the `app.py`


jayair:
For Go, let's look at the build process used in this Serverless Framework plugin — https://github.com/mthenw/serverless-go-plugin

PR #205, opened 2021-03-06
Proposal: Support including extra files with final bundle

thdxr:
Decided to mock it up in actual code to demonstrate how it could work.  Basically I need to add some supporting files with my bundles (native libraries for example) for my function to work properly so I thought SST could support an `includes` parameter.

Usage:

```
    const api = new sst.Api(this, "Api", {
      defaultFunctionProps: {
        include: [
          {from: "node_modules/.prisma/client/prisma.schema", to: "prisma.schema"},
        ],
      }
    })
```

### DIFF
--- a/packages/resources/src/Function.ts
+++ b/packages/resources/src/Function.ts
@@ -23,6 +23,10 @@ const runtimeTargetMap = {
 
 export type HandlerProps = FunctionHandlerProps;
 export type FunctionDefinition = string | Function | FunctionProps;
+export type Include = {
+  from: string;
+  to: string;
+};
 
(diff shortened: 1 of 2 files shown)


fwang:
Thanks @thdxr this is great! I'm going to take a look at it tomorrow.

thdxr:
Thought about this more and another approach would be exposing a similar API to `aws-lambda-nodejs` https://github.com/ryands17/prisma-lambda/blob/main/lib/prisma-lambda-stack.ts#L24

fwang:
@thdxr your fork has been merged in to #225 and is work in progress

Issue #208, opened 2021-03-08
Additional loader when transpiling Lambda code


claudioc:
Is it possible to add an additional ~babel (?)~ esbuild loader when transpiling? (I am using TypeScript).

Use case: I need to import HTML files, as in `import html from './template.html';`; it's easy to tell Typecript that it's OK to import html (see [this post](https://medium.com/@sampsonjoliver/importing-html-files-from-typescript-bd1c50909992)), but then the error is `error: No loader is configured for ".html" files: ./template.html`.

claudioc:
OK, I realize that what it's needed is a loader for `esbuild`. The closest one is the [txt loader](https://esbuild.github.io/content-types/#text).

I leave it here in case someone knows how to have esbuild use the 'txt' loader for files with a 'html' extension :)

fwang:
Yup, this is on our immediate roadmap. We want to support most if not all of the bundling options for esbuild.

@jayair Can you take a look and see how much work is required to support all of these build options?
https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-lambda-nodejs.BundlingOptions.html

fwang:
@claudioc The `loader` option has been added in v0.9.17

Here's an example on how to set it - https://docs.serverless-stack.com/constructs/Function#configuring-bundling

claudioc:
Oh yes, thanks. Is the `copyFile` new as well? Love it!

fwang:
Yuup!

claudioc:
And this solved my use case! Thanks :)

```
    const api = new sst.Api(this, 'Api', {
      defaultFunctionProps: {
        bundle: {
          loader: {
            '.html': 'text'
          }
        }
      },
      routes: {
        'GET /': 'src/index.handler',
```

fwang:
Case closed ;)

PR #225, opened 2021-03-11
Support bundle options

fwang:
Adding the following bundling option:
- loader
- externalModules
- nodeModules
- copyFiles


fwang:
Closes #216 
Closes #208

Issue #256, opened 2021-03-17
Support attaching permissions to instance of sst.Function 


SethThomas:
The current implementation of `attachPermissionsToRole` lets us specify SST constructs when adding permissions.
For example
```Javascript
    let workerLambda = new sst.Function(this, "Worker", {
      handler: "src/lambdas/worker.main"
    });
    
    let table = new sst.Table(this, "mydb", {
      fields: {
        PK: sst.TableFieldType.STRING,
        SK: sst.TableFieldType.STRING
      },
      primaryIndex: {partitionKey: "PK", sortKey: "SK"}
    });
    
    workerLambda.attachPermissions([table]);
    
```

The currently supported SST constructs for the `attachPermissions` function are [Api, Table, Topic, Queue](https://github.com/serverless-stack/serverless-stack/blob/c27af489bb24f089f169a4c32033606acb02ccda/packages/resources/src/util/permission.ts#L102).  I've come across a use case where I want to call a lambda from another lambda, which requires that I add the appropriate permissions to an SST.Function construct.  

What I want to do

```Javascript
    let lambdaA = new sst.Function(this, "LambdaA", {
      handler: "src/lambdas/lambdaA.main"
    });

    let lambdaB = new sst.Function(this, "LambdaB", {
      handler: "src/lambdas/lambdaB.main"
    });
    
    let table = new sst.Table(this, "mydb", {
      fields: {
        PK: sst.TableFieldType.STRING,
        SK: sst.TableFieldType.STRING
      },
      primaryIndex: {partitionKey: "PK", sortKey: "SK"}
    });
    
    lambdaA.attachPermissions([table,lambdaB]);   // <--- Give lambdaA permission to call lambdaB

```  

However, this functionality is not yet supported.  Instead, I have to define the IAM policy myself:

```Javascript
    lambdaA.attachPermissions([
      table,
      new iam.PolicyStatement({
        actions: ["lambda:InvokeFunction"],
        effect: iam.Effect.ALLOW,
        resources: [
          <THE ARN OF lambdaB>
        ]
      })
    ]);

```

This is not terribly problematic, but it would be nice if the Function construct was supported when attaching permissions.

fwang:
Cutting a release with your PR 🎉

PR #259, opened 2021-03-18
Support attaching Function invoke permissions

SethThomas:
This is one way to add support for issue #256. 

This change would make it possible to give Functions permissions to invoke other Functions:

```js
const sns = new sns.Topic(this, "Topic");
const table = new sst.Table(this, "Table");
const moreFun =  new sst.Function(this, "MyLambda", {
      handler: "src/myLambda.main"
    });

fun.attachPermissions([sns, table, moreFun]);
```


fwang:
Closes #256

fwang:
Thanks, @SethThomas! That's awesome you updated the tests and the docs too.

Merging this.

PR #260, opened 2021-03-18
Support passing jest configs to sst test

kpamaran:
Motivation: My team recently migrated a repo from Serverless Framework to SST and naturally our TS compiler started complaining about our test files. This tiny change would allow developers to pass their jest-config to SST's jest-cli process via `sst test`, otherwise we would simply need to set up a direct jest dependency and test script ourselves.

### DIFF
--- a/packages/cli/scripts/test.js
+++ b/packages/cli/scripts/test.js
@@ -20,15 +20,20 @@ let argv = process.argv.slice(2);
 const createJestConfig = require("./util/createJestConfig");
 const path = require("path");
 const paths = require("./util/paths");
-argv.push(
-  "--config",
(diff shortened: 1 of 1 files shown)


fwang:
Hi @kpamaran, thanks for the contribution. Let me just pull in @jayair to take a quick look in the morning.

brinsleylogic:
@fwang and @jayair, just a casual bump on this being reviewed/merged, would be handy for something I'm looking at currently :)

fwang:
Looping back to this PR, the `sst test` command is removed in SST v1. Users are responsible for setting up `jest`, and have full control over the config.

Issue #363, opened 2021-05-12
Respect Permission Limitations Locally when Running in Debug Mode


simonireilly:
When running in debug mode the ServiceRole of the lambda is not the source of the credentials. The credentials used are those of the deployer.

## Story

Given my lambda has permissions to only read from a service, then when running `npx sst start` I should not be able to write to the service.

## Details

Operating system: Ubuntu 20.04.2 LTS
Node version: v12.16.3

## Minimal Recreation

See this PR: https://github.com/simonireilly/sst-tester/pull/1

![assumed-credentials-bug-recreation](https://user-images.githubusercontent.com/30017294/118230566-c8f3c100-b485-11eb-80d7-1ae5705c9de7.gif)

simonireilly:
This issue is recreated with the following repository: https://github.com/simonireilly/sst-tester

fwang:
Hmm.. just tried out the repo and I'm getting the AccessDenied error.

I would try the following to dig deeper:
- in Lambda, make an aws sdk call to STS's `getCallerIdentity` or `getAccessKeyInfo` to verify the IAM credentials the sdk is using
- does the Lambda have other non-dynamodb permissions? (ie. make an aws sdk call to list out all SNS topics)

<img width="1072" alt="Screen Shot 2021-05-13 at 3 29 05 PM" src="https://user-images.githubusercontent.com/83515/118177100-3bba5900-b400-11eb-99b8-62b824f7d394.png">

<img width="1432" alt="Screen Shot 2021-05-13 at 3 29 41 PM" src="https://user-images.githubusercontent.com/83515/118177111-3d841c80-b400-11eb-9395-8c9adee06042.png">

simonireilly:
Thanks Frank, what a great suggestion. Get caller identity did indeed show, that the caller is myself, locally:

MR here with a GIF for recreation https://github.com/simonireilly/sst-tester/pull/1


```
[1620972564267] INFO (LambdaLogger/34329 on simon-XPS-15-9500): Caller Identity
    callerIdentity: {
      "$metadata": {
        "httpStatusCode": 200,
        "requestId": "d52cdc62-b7c2-4e65-9fcd-207546f30205",
        "attempts": 1,
        "totalRetryDelay": 0
      },
      "UserId": "AIDAUWGULNAJUAPI2KCBK",
      "Account": "322567890963",
      "Arn": "arn:aws:iam::322567890963:user/simonireilly+cloudguru+user"
    }
[...]

PR #369, opened 2021-05-17
sst start: Fix AWS_PROFILE used over AWS_ACCESS_KEY_ID when aws sdk v3 is used inside Lambda functions

fwang:


fwang:
Closes #363

PR #375, opened 2021-05-20
sst start: Feature/experimental xray support

simonireilly:
WIP on Xray supports using manual segments to locally mimic the Xray context existing inside the environment lambda is running in.

May close: https://github.com/serverless-stack/serverless-stack/issues/123

### TODO:

Design abstraction:

- [x] The `bootstrap.js` should not error when segments fail.
- [x] The XRay cls hooked wrapper on the lambda client should only be enabled when XRay tracing is enabled.
-  [x] ~The XRay segments being drained to AWS should be non-blocking~. Going to have to allow these PUTs to xray synchronously
    -  [x] Write the collected segments to a buffer and then drain them over HTTP.
- [x] Remove test stack from POC after adding tests and perform TD refactor.

### Design

- The websocket forwards the env, this contains the `_X_AMZN_TRACE_ID` env var
- The spawned child process inherits the env therefore has access to this
- The AWSXRay node library establishes a Container Local Storage (https://www.npmjs.com/package/cls-hooked) which has an AWSXray namespace
- The namespace is fetched, and the handler is wrapped as an async function being executed within the contained namespace.
- Lambdas with AWSCapture setup inherit the automated context from the cls/namespace binding
- After execution the segments are drained and sent to Xray using the aws-sdk, this means the local user does not need to run an AWS-XRay daemon locally.

#### When stack is deployed for debug 'npx sst start`.

- Local nodes appear as being called by the `SST:Function` stub lambdas, which gives the user a clear route to debug
- Additional trace context to SQS, DynamoDB etc. are captured

_Service lense shows traces to local file system_

image

_Service Map has context of calling order_

image

#### When stack is deployed in normal mode `npx sst deploy`

_Service lense shows traces with **NO** local file system_

image

_Service Map has context of calling order_

image

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -42,6 +42,7 @@
     "@aws-cdk/aws-iam": "1.111.0",
     "@aws-cdk/aws-lambda": "1.111.0",
     "@aws-cdk/core": "1.111.0",
+    "@aws-sdk/client-xray": "^3.16.0",
     "@babel/core": "^7.10.5",
     "@babel/eslint-parser": "^7.11.3",
     "@babel/eslint-plugin": "^7.11.3",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -56,6 +56,7 @@
     "@typescript-eslint/parser": "^4.15.1",
     "aws-cdk": "1.111.0",
     "aws-sdk": "^2.761.0",
+    "aws-xray-sdk": "^3.3.3",
     "body-parser": "^1.19.0",
     "chalk": "^4.1.0",
     "chokidar": "^3.4.3",
--- a/packages/cli/scripts/util/bootstrap.js
+++ b/packages/cli/scripts/util/bootstrap.js
@@ -13,12 +13,19 @@ const path = require("path");
 const fetch = require("node-fetch");
 const { getChildLogger, initializeLogger } = require("@serverless-stack/core");
 const { serializeError } = require("../../lib/serializeError");
+const { wrapWithLocalXray, drain } = require("./xray");
+
+// Handle draining xray segments on exit
+process.on("exit", async () => {
+  console.info("Goodbye old friend");
+  return drain();
+});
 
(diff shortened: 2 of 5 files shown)


simonireilly:
Thanks for running the tests @fwang, I added the missing dependency, I think yarn workspaces locally had this for me hoisted into the parent node_modules so it didn't fail.

I have ran `yarn run test` locally and also used [act](https://github.com/nektos/act) to test out the various versions. I had an issue locally with act not able to secure port 3000, I think that might be a docker specific issue with the replication of github actions locally. If it occurs in ci I will debug locally.

fwang:
Thanks @simonireilly! Tests are looking good.

simonireilly:
@fwang I think this feature is complete, some questions I had though:

1. This is more or less a Proof of Concept so let me know what your thoughts are on supporting local XRay?
2. Would an example in the examples folder be beneficial, I notice there isn't one already, and instrumenting code with XRay will likely be a go-to for most users?
3. I think logging the trace ID, or Url locally would help developers tighten the feedback loop if using XRay to profile their code; something like:

```
Debug session started. Listening for requests...
1676aedd-1fee-4730-b3c5-aefee84a232b REQUEST dev-cache-stack-my-stack-ApiLambdaGETB1714EF3-nGmMAX3pQP3M [src/lambda.handler] invoked by API GET /
1676aedd-1fee-4730-b3c5-aefee84a232b RESPONSE {"statusCode":200,"headers":{"Content-Type":"text/plain"},"body":"Hello, World! Your request was received at 03/Aug/2021:09:59:41 +0000."}
1676aedd-1fee-4730-b3c5-aefee84a232b Trace available at: https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#servicelens:traces/1-6108e0a1-7d2424a86c17223b023a4003
```

Given that this feature wraps the default handler, and adds new node_module dependencies, it is probably a canary release candidate.

PR #384, opened 2021-05-24
KinesisStream: Add construct for Kinesis streams

fwang:


fwang:
Cutting a release later today.

PR #423, opened 2021-06-09
Replace dynamodbTable property with s3Bucket

mmccall10:
Under the "Configure the internally created CDK Bucket instance" section the dynamdbTable property appears to be a typo and should be s3Bucket.


fwang:
Thanks @mmccall10! Nice catch!

PR #433, opened 2021-06-11
StaticSite: update typo tableName to bucketName

mmccall10:
Looks like the property tableName is supposed to be bucketName.


fwang:
Ahh.. of course!

PR #517, opened 2021-07-05
Improving docs for bundling props

jayair:


pabacham:
My second Stack accepts following args 
`constructor(scope: sst.App, id: string, props?: sst.StackProps)`

But seems like sst.StackProps does not accept api prop, it throws a TS error 

Argument of type '{ api: sst.Api; }' is not assignable to parameter of type 'StackProps'.
  Object literal may only specify known properties, and 'api' does not exist in type 'StackProps'.ts(2345)

Rohithgilla12:
@pabacham hey I am getting this too, how did you solve it?

fwang:
Note that you don't have to pass it through `StackProps`. You can pass it directly to the constructor. Here's an example in TypeScript - https://docs.serverless-stack.com/advanced/cross-stack-references

Rohithgilla12:
@fwang Something like this
```
interface StorageStackProps extends sst.StackProps {
  table: sst.Table;
}

```

fwang:
Yup, right on!

PR #522, opened 2021-07-08
Allow using Code param

erikm-icario:
We are somewhat limited by what we can do with the current builders.  By allowing the developer to pass in a Code object, they can create their lambda however they want.

This is a somewhat naive implementation that I'm happy to talk about.

### DIFF
--- a/packages/resources/src/Function.ts
+++ b/packages/resources/src/Function.ts
@@ -58,6 +58,15 @@ export interface FunctionProps
    */
   readonly runtime?: string | lambda.Runtime;
   /**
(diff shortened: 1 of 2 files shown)


erikm-icario:
While this addresses the deploy side of things, we're having a discussion here https://github.com/serverless-stack/serverless-stack/discussions/219 about the "sst start" scenario.  I believe that we can continue to use the existing start watchers and then build using the local bundling settings.

erikm-icario:
Closing this in favor of https://github.com/serverless-stack/serverless-stack/pull/551

PR #551, opened 2021-07-15
Function: support custom esbuild config

erikm-icario:
This is an implementation of supporting esbuild configuration in nodejs lambda building for both start and deploy.

There are a number of potential improvements:
* Right now the esbuild config files are discovered by convention ('esbuild.dev.js' and 'esbuild.prod.js').  We could make it a configuration option so that each lambda could be built with its own options.
* I have positive tests for the configuration.  I didn't put in tests for if the configuration had errors in it (though I did check it manually)
* esbuild.js could probably be cleaned up.


fwang:
@erikm-icario Thanks for the high-quality PR!

I made two minor changes:
1. Instead of hardcoding to `esbuild.dev.js` and `esbuild.prod.js`, I added an `esbuildConfig` option to the `bundle` prop. It takes the path to the config, and ppl can conditionally pass in different files.
2. I added the restriction that only esbuild `plugins` can be overridden. SST will fail if `outdir`, `entryPoints` and `metadata` are changed. We will allow overriding other esbuild params on request.

Thanks again for taking this on!

PR #552, opened 2021-07-15
Function: support granting permissions to RDS ServerlessCluster

pyoung213:


packages/resources/src/util/permission.ts
```diff
@@ -112,6 +113,22 @@ export function attachPermissionsToRole(
...
+      )
+    ) {
+      // @ts-expect-error We do not want to import the cdk modules, just cast to any
+      const clusterArn = permission.clusterArn;
+      role.addToPolicy(
+        buildPolicy("rds-data:*", [clusterArn, `${clusterArn}/*`])
+      );
+      // @ts-expect-error We do not want to import the cdk modules, just cast to any
+      const secret = permission.secret;
+      if(secret) {
```
pyoung213, line 129: This needs read access to the secret manager.  From docs of the function I was using (plus it didn't work without it)

`/**
     * Grant the given identity to access to the Data API, including read access to the secret attached to the cluster if present.
     *
     * @param grantee The principal to grant access to.
     * @stability stable
     */
    grantDataApiAccess(grantee: iam.IGrantable): iam.Grant;`

PR #574, opened 2021-07-20
Doc: explain the design choice behind using esbuild

fwang:


pabacham:
@fwang  The only problem I am facing is that it drains the memory when the app is growing. Currently, I have 5 stacks running on my local and it costs around 9GB of memory to be used. It will be good if we can limit watching the files for specific stacks (or any other options) then in order to decrease the memory usage.

fwang:
> Currently, I have 5 stacks running on my local and it costs around 9GB of memory to be used.

Oh no! That's terrible news. Do you mind providing some details:
- How many Lambda functions do you have roughly?
- Does `sst start` take up 9GB memory right away when running it? Or does it grow over time?
- Windows or Mac?
- Can you run `sst start` and send me the `.build/sst-debug/log` at frank@serverless-stack.com? 
- Lastly, may I bother you with a sample repo that can reproduce the issue? Or temporary clone permission to your repo if it doesn't contain sensitive data. No worries if not, I can start investigating with the info above.

I want to check if it is indeed the watcher that's taking up the memory. If it is, we'd need to look for a more efficient watcher or not watch all the stacks as u suggested.

Let me know, and I will start investiaging right away!

fwang:
> Currently, I have 5 stacks running on my local and it costs around 9GB of memory to be used.

@pabacham Oh no! That's terrible news. Do you mind providing some details:
- How many Lambda functions do you have roughly?
- Does `sst start` take up 9GB memory right away when running it? Or does it grow over time?
- Windows or Mac?
- Can you run `sst start` and send me the `.build/sst-debug/log` at frank@serverless-stack.com? 
- Lastly, may I bother you with a sample repo that can reproduce the issue? Or temporary clone permission to your repo if it doesn't contain sensitive data. No worries if not, I can start investigating with the info above.

I want to check if it is indeed the watcher that's taking up the memory. If it is, we'd need to look for a more efficient watcher or not watch all the stacks as u suggested.

Let me know, and I will start investiaging right away!

pabacham:
@fwang 
1) 76 functions. Every API endpoint has its own function. All these functions splitted between 2 API (admin, public). 7 for the public API, 69 for the admin API.
2) I am using "sst start --stage $(whoami)". It takes up 9GB of memory right away.
3) Mac
4) Will do
5) It is a private organization repo, so I can not do here anything

Screenshot https://monosnap.com/file/8dFPbP1zZAS9jpE7DG0ZUwFmLjZBxm

fwang:
Hey @pabacham I managed to reproduce the issue.

### What is going on?
The `esbuild` process takes 9GB of memory.

### Why is it taking 9GB?
Normally, `esbuild` doesn't not try to build JS files inside `node_modules` b/c `esbuild` is configured to exclude all dependencies inside the `package.json`. Looking at ur `.build/sst-debug.log`, dependencies like `aws-sdk` were not excluded. And that caused `esbuild` to take up a lot of memory.

### Why were the dependencies not excluded?
Again by looking at ur `.build/sst-debug.log`, I gathered ur folder structure looks like this:
```
simple-events/
  package.json    <- outside ur SST app
  node_modules/
  packages/
    api/
      sst.json
      package.json    <- inside ur SST app
      node_modules/
      src/
        entities/
          events/
            lists.ts
            ...
```
`esbuild` tried to exclude the dependencies from the `package.json` that's inside ur SST app, but the dependencies like `aws-sdk` are installed in the `node_modules` that's outside of ur SST app. Hence they were not excluded.

Can you confirm if this is indeed the case?

pabacham:
@fwang That's correct. We use Lerna in the project and we have several apps inside the "packages" folder. Every app has its own package.json, meanwhile, all the node_modules inside every package are hoisted to the root node_modules folder. We started the project when there was no example of how to build it with Lerna, SST and React, so the structure looks like in the description above.

fwang:
@pabacham Sorry for the delayed response. I will continue the conversation here https://github.com/serverless-stack/serverless-stack/issues/579#issuecomment-887349274

PR #575, opened 2021-07-20
Support esbuild config in sst config

erikm-icario:
We use esbuild to build the entire app, not just the NodeJS functions.  Some constructs can benefit from the use of things like decorators (generators for example).

After discussing with @fwang, it was decided to add an optional esbuildConfig setting to sst.json.

This PR adds that setting to sst.json and it is used in cdkhelpers.  It does not affect or override what is used for building the nodejs functions.

A future PR can look at somehow merging esbuildConfig from sst.json with defaultFunctionProps if that is desired.


fwang:
Thanks @erikm-icario!

Btw, I put in the change for the `bundle` option to inherit the `esbuildConfig` setting from sst.json.

PR #591, opened 2021-07-23
ApolloApi: fix CORS not working if authorization is enabled

fwang:


riccardogiorato:
Hey @fwang with this change every ApolloApi has been split into 2 different lambda, 1 for the GET and one for the POST.

Is there a reason to have 2 lambdas?

fwang:
@riccardogiorato, I will follow up in the new issue thread.

PR #596, opened 2021-07-23
Fix for specifying handler with period in path

### LINKED ISSUE #583: Function: cannot specify handler with dot in the path
Currently, defining the handler path like this:
```
handler: `./src/lambda.handler`,
```
causes a build error:
```
Transpiling Lambda code...
 > error: Could not resolve "../.js"
```

---
Notify ppl who ran into this:
- Dax (Slack thread https://serverless-stack.slack.com/archives/C01HQQVC8TH/p1626895089465300)
- Brent (GH issue https://github.com/serverless-stack/serverless-stack/issues/589)

thdxr:
Fixes #583

### DIFF
--- a/packages/cli/scripts/start.js
+++ b/packages/cli/scripts/start.js
@@ -476,14 +476,10 @@ async function handleRemoveWatchedFiles(files) {
 // Lambda Reloader functions - NodeJS //
 ////////////////////////////////////////
 
-async function handleTranspileNode({
-  srcPath,
-  handler,
-  bundle,
-  esbuilder,
-  onSuccess,
-  onFailure,
-}, config) {
+async function handleTranspileNode(
+  { srcPath, handler, bundle, esbuilder, onSuccess, onFailure },
+  config
+) {
   // Sample input:
   //  srcPath     'service'
   //  handler     'src/lambda.handler'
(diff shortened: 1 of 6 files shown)


packages/cli/scripts/start.js
```diff
@@ -476,14 +476,10 @@ async function handleRemoveWatchedFiles(files) {
...
 
-async function handleTranspileNode({
-  srcPath,
-  handler,
-  bundle,
-  esbuilder,
-  onSuccess,
-  onFailure,
-}, config) {
+async function handleTranspileNode(
```
thdxr, line 479: Seems like my prettier is formatting differently but I am on latest

packages/cli/scripts/start.js
```diff
@@ -693,18 +689,19 @@ function handleRunTypeCheck(srcPath, inputFiles, tsconfig, config) {
...
   for (const ext of extensions) {
-    const file = path.join(appPath, srcPath, `${name}${ext}`);
+    const file = path.join(appPath, srcPath, dir, `${name}${ext}`);
     if (await checkFileExists(file)) {
       return file;
     }
   }
 
-  return path.join(appPath, srcPath, `${name}.js`);
+  return path.join(appPath, srcPath, dir, `${name}.js`);
```
thdxr, line 704: Curious why it defaults to this without verifying it exists

fwang: Yeah we should. Let me add a check.

packages/cli/scripts/start.js
```diff
@@ -693,18 +689,19 @@ function handleRunTypeCheck(srcPath, inputFiles, tsconfig, config) {
 }
 
 async function getHandlerFilePath(appPath, srcPath, handler) {
-  const parts = handler.split(".");
+  const parts = path.basename(handler).split(".");
```
fwang, line 692: It doesn't handle this case, as suggested here - https://github.com/serverless-stack/serverless-stack/issues/600
```
handler: 'src/users/index.controller.main'
```

We actually do something similar on `sst deploy`. Let's re-use the same logic - https://github.com/serverless-stack/serverless-stack/blob/master/packages/resources/src/util/nodeBuilder.ts#L130

fwang:
Hey @thdxr, I'm not able to push changes to your branch. So I created forked your branch and created a new PR https://github.com/serverless-stack/serverless-stack/pull/611

PR #615, opened 2021-07-28
helpful error added to nodebuilder copyfiles

myufa:
Added check to see that copyFiles options on bundle is array and return a helpful error message specifying the expected type and given type


fwang:
Thanks @myufa! We are planning to put in validation across all input props very soon. I will merge this in with the rest of the changes.
https://github.com/serverless-stack/serverless-stack/issues/486

PR #623, opened 2021-07-30
Update esbuild to 0.12.17 

radu-c:
This patch updates esbuild to version 0.12.17.
Due to some API changes in esbuild I had to do some small modifications to the way the metafile is created.
The way esbuild is used could benefit from a later refactor but this fix should hold hopefully until esbuild's API is stable and a refactor effort is justified.


thdxr:
Had some rebase issue with this so reopened the PR at #703

PR #632, opened 2021-08-02
Support environment property in StaticSite

### LINKED ISSUE #629: StaticSite: support environment prop
---
_Requested by [Dax on Slack](https://serverless-stack.slack.com/archives/C01HQQVC8TH/p1627172631001200)_

thdxr:
Fixes #629

### DIFF
--- a/packages/resources/src/ReactStaticSite.ts
+++ b/packages/resources/src/ReactStaticSite.ts
@@ -12,16 +12,17 @@ import {
 // Interfaces
 /////////////////////
 
-export interface ReactStaticSiteProps extends StaticSiteProps {
-  readonly environment?: { [key: string]: string };
-}
+/**
+ * @deprecated This type has been replaced by the standard StaticSiteProps
+ */
+export type ReactStaticSiteProps = StaticSiteProps;
 
 /////////////////////
 // Construct
 /////////////////////
 
 export class ReactStaticSite extends StaticSite {
-  constructor(scope: cdk.Construct, id: string, props: ReactStaticSiteProps) {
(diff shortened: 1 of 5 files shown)


fwang:
Looks good! A couple of things to wrap this up:
- ReactStaticSite: we can remove the `ReactStaticSiteProps` interface from as it doesn't do much now
- StaticSite: add tests for the `environment` prop
- doc: move the `environment` props from the `ReactStaticSite` doc to the `StaticSite` doc
- doc: add an example using the `environment` props in the `StaticSite` doc
- as a side note, as I was looking at the `StaticSite` doc, I see we still have a `Creating a React site` example of using the `StaticSite`. Maybe remove the code snippet, and just tell ppl to use the `ReactStaticSite` construct

thdxr:
I made the changes.  Thinking more about docs though this section: https://docs.serverless-stack.com/constructs/ReactStaticSite#configuring-environment-variables is probably going to be useful in the core StaticSite section as well.  Do we want to duplicate it?  Also this PR makes the ReactStaticSite construct less useful so not sure what we want to encourage users to use.

fwang:
Yup I agree. In `ReactStaticSite`'s doc, we just need to explain to ppl on how to use the environment variable in ur React's HTML files and JS files; and how to set them through `environment`.

The goal is so that when a React person reads the `ReactStaticSite` doc, he should be able to set everything up without relying on both the `ReactStaticSite` and `StaticSite` doc.

thdxr:
Merged this by accident without doc updates and reverted it.  Going to open a new PR

PR #636, opened 2021-08-03
Support copyFiles regardless of runtime

### LINKED ISSUE #635: Support copyFiles regardless of runtime
CopyFiles is a useful feature regardless of the runtime so we should pull this out of the runtime specific props.

thdxr:
Fixes #635 

I started to write a test but wasn't sure if there was a recommended way to check what's in `./build` for the specific handler

### DIFF
--- a/packages/resources/src/Function.ts
+++ b/packages/resources/src/Function.ts
@@ -3,6 +3,7 @@
 
 import path from "path";
 import * as esbuild from "esbuild";
+import * as fs from "fs-extra";
 import * as cdk from "@aws-cdk/core";
 import * as iam from "@aws-cdk/aws-iam";
 import * as lambda from "@aws-cdk/aws-lambda";
@@ -81,10 +82,7 @@ export interface FunctionProps
    *
    * @default - Defaults to true
    */
-  readonly bundle?:
-    | boolean
-    | FunctionBundleNodejsProps
-    | FunctionBundlePythonProps;
+  readonly bundle?: FunctionBundleProp;
   readonly permissions?: Permissions;
   readonly layers?: lambda.ILayerVersion[];
 }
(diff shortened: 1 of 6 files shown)


packages/resources/src/Function.ts
```diff
@@ -320,15 +328,37 @@ export class Function extends lambda.Function {
...
-    return layers.map(layer => {
+  static copyFiles(
+    bundle: FunctionBundleProp | undefined,
+    srcPath: string,
+    buildPath: string
+  ) {
+    if (!bundle) return;
+    if (typeof bundle === "boolean") return;
+    if (!bundle.copyFiles) return;
+
```
thdxr, line 339: This is using Typescript type narrowing to allow it to infer the right type

packages/resources/src/Function.ts
```diff
@@ -320,15 +328,37 @@ export class Function extends lambda.Function {
...
+    bundle.copyFiles.forEach(({ from, to }) => {
+      const fromPath = path.join(srcPath, from);
+      const toPath = path.join(buildPath, to);
+      console.log(process.cwd(), fromPath, toPath);
```
fwang, line 343: Debugging log?

thdxr: Good catch :sweat_smile:

fwang:
> I started to write a test but wasn't sure if there was a recommended way to check what's in `./build` for the specific handler

We don't have a standard way of checking it. We have an existing test for `copyFiles` that just calls `fs.existsSync` to test for specific files.
https://github.com/serverless-stack/serverless-stack/blob/master/packages/cli/test/nodejs-build-bundle-copyfiles/nodejs-build-bundle-copyfiles.test.js

fwang (approved):

PR #644, opened 2021-08-04
Queue: Append .fifo to queue name when fifo is true

### LINKED ISSUE #639: Queue: Creating a FIFO queue fails if a `queueName` is not configured
### Issue
```
  const queue = new Queue(stack, "Queue", {
    sqsQueue: {
      fifo: true,
    }
  });
```
The above code fails with error:
```
FIFO queue names must end in '.fifo'
```

### Solution
Suffix the queue name with `.fifo`

thdxr:
I decided not to override a name that the user configured so that nothing unexpected happens

fixes #639

### DIFF
--- a/packages/resources/src/Queue.ts
+++ b/packages/resources/src/Queue.ts
@@ -38,7 +38,7 @@ export class Queue extends cdk.Construct {
     if (cdk.Construct.isConstruct(sqsQueue)) {
       this.sqsQueue = sqsQueue as sqs.Queue;
     } else {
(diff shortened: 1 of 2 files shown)


fwang:
Let's add another test for fifo queue and user specified the queueName, and test that we don't append `.fifo` to the queue name.

thdxr:
Ah I ned to remove `test.only` will fix that and add the additional test

fwang (approved):

Issue #658, opened 2021-08-06
StaticSite should skip build on removal


thdxr:
### Requests
- https://serverless-stack.slack.com/archives/C01HQQVC8TH/p1628276552055200

PR #660, opened 2021-08-07
Fix static site looking for build directory when skipping build

thdxr:
Refactored a bit to remove nesting as well


thdxr:
Fixes #658

PR #699, opened 2021-08-16
StaticSite: Add support for external domains

warwickgrigg:
Enhancement.  As discussed on Slack. With documentation , a new test case,  and two test cases updated.  I've tested this e2e for my own hostname.

In brief, for external domains omit customDomain, and include:

```
cfDistribution: {
  domainNames: ["www.domain.com"],
  certificate: Certificate.fromCertificateArn(this, "MyCert", certArn)
}
```
Then set external domain's CNAME record to the cloudfront distribution's URL.

### DIFF
--- a/packages/resources/src/StaticSite.ts
+++ b/packages/resources/src/StaticSite.ts
@@ -459,26 +459,26 @@ export class StaticSite extends cdk.Construct {
 
     const cfDistributionProps = cfDistribution || {};
 
-    const domainNames = [];
+    // Validate input
+    if (cfDistributionProps.certificate) {
+      throw new Error(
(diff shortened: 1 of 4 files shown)


warwickgrigg:
@fwang Thank you for moving this forward. Is the failed check a false negative, or do I need to fix something?

fwang:
Hey @warwickgrigg, I moved the code around a bit to rely on using the `customDomain` to configure external domains:
```
const site = new StaticSite(this, "Site", {
  path: "path/to/src",
  customDomain: {
    isExternalDomain: true,
    domainName: "domain.com",
    certificate: Certificate.fromCertificateArn(this, "MyCert", certArn),
  },
});
```

Behind the scene, it does exactly as what you proposed. This way, `site.customDomainUrl` still works.

Thanks again for working on the PR. This will go into the next release!

PR #702, opened 2021-08-16
CLI: Add update command to update SST + related CDK versions

### LINKED ISSUE #588: sst update: add a command to update SST and CDK versions
Link to discussion - https://github.com/serverless-stack/serverless-stack/discussions/581

thdxr:
Fixes #588


packages/core/src/update/update.ts
```diff
@@ -0,0 +1,93 @@
...
+          "node_modules/@serverless-stack/resources/package.json"
+        )
+      )
+      .toString()
+  );
+
+  function cdk(type: "dependencies" | "devDependencies") {
+    const updates = Object.keys(json[type] || {})
+      .filter((key) => key.startsWith("@aws-cdk"))
+      .filter((key) => json.dependencies[key] || json.devDependencies[key])
```
fwang, line 40: Curious if this filter is required as this will always be true?

packages/core/src/update/update.ts
```diff
@@ -0,0 +1,93 @@
...
+      .filter((key) => compare.dependencies[key])
```
fwang, line 41: Does this mean we are only updating the @aws-cdk packages that exist in resources? (We should update all @aws-cdk packages, CDK doesn't handle mixed package versions very well)

packages/core/src/update/update.ts
thdxr: This was leftover from a previous version of the code, removing

packages/core/src/update/update.ts
thdxr: Yeah good point.  I updated to use whatever `@serverless-stack/core` is using

packages/core/src/update/update.ts
```diff
@@ -0,0 +1,93 @@
...
+          rootDir,
+          "node_modules/@serverless-stack/resources/package.json"
+        )
+      )
+      .toString()
+  );
+
+  function cdk(type: "dependencies" | "devDependencies") {
+    const updates = Object.keys(json[type] || {})
+      .filter((key) => key.startsWith("@aws-cdk"))
```
fwang, line 37: Users might also include the `aws-cdk` package. The regex used in `packages/resources/bin/aws-cdk-version-check.js` to check the dependency version is `/^@?aws-cdk/.test(key)`.

thdxr: Fixed

PR #714, opened 2021-08-18
Feature: Configurable build directory

cfraz89:
Adds a flag `--build-dir` / config file option `buildDir` which allows to override the build output folder from `.build` to a specified directory.

Note that I adjusted the builder to build *all* output into this path. So where previously, if there was a srcPath it would have gone into `srcPath/.build/hash-$ts`, it now builds into `.build/srcPath/hash-$ts`. Although I it behaves fine in my testing, and all tests pass (except the dotnet one which times out for me), I am wary that there was a reason it was done that way in the first place. So if this is an issue let me know.


cfraz89:
Actually I'll need to expand the scope of this one. I'm trying to set up sst in an nx monorepo. In order for this to week, at least I'll need a different build folder, but also the ability for the app to live in a different folder to the package.json and node_modules.

I'm thinking of adding a -c / --config-file option that allows you to specify the path to sst.json. How does that sound?

fwang:
Thanks @cfraz89! This has been implemented. All build artifacts are now in `.sst`.

PR #718, opened 2021-08-18
App: Support setting defaultRemovalPolicy

### LINKED ISSUE #634: Set all resources' removalPolicy in the entire App/Stack to DESTROY
### Issue
For short-lived stages like a PR stage, it makes sense to have all resources removed on stack removal. We can add a high level flag on the `sst.App` or `sst.Stack`, to set removalPolicy for all resources.

### Workaroud
For now, you can loop through each CFN resource and set the flag. Something like:
```
stack.node.children.forEach(resource => {
  resource = resource as cdk.CfnResource;
  resource.applyRemovalPolicy(cdk.RemovalPolicy.DESTROY);
});
```

---
_Requested by [Sam Hulick on Slack](https://serverless-stack.slack.com/archives/C01JG3B20RY/p1627945769078900)_

thdxr:
Closes #634

### DIFF
--- a/packages/resources/src/App.ts
+++ b/packages/resources/src/App.ts
@@ -128,6 +128,10 @@ export class App extends cdk.App {
     | FunctionProps
     | ((stack: cdk.Stack) => FunctionProps)
   )[];
+  private _defaultRemovalPolicy?: cdk.RemovalPolicy;
+  public get defaultRemovalPolicy() {
+    return this._defaultRemovalPolicy;
+  }
 
   /**
    * The callback after synth completes.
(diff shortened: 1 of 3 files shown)


fwang:
I think S3 buckets might be trouble b/c they need to be emptied before removed. You'd have to set [autoDeleteObjects](https://docs.aws.amazon.com/cdk/api/latest/docs/aws-s3-readme.html#bucket-deletion):
```js
new s3.Bucket(this, 'MyBucket', {
  removalPolicy: RemovalPolicy.DESTROY,
  autoDeleteObjects: true,
});
```

And when `autoDeleteObjects` is true, CDK creates a custom resource to empty the bucket on stack remove.

I can't think of a good solution off the bat. We might not be able to set this flag right before synth. Not sure, maybe give it a try.

PR #770, opened 2021-08-30
Making Windows compatible

robertobasile84:
Without this change, hitting the root drive (whatever letter) would case a RangeError: Maximum call stack size exceeded

### DIFF
--- a/packages/core/src/packager/packager.ts
+++ b/packages/core/src/packager/packager.ts
@@ -57,10 +57,11 @@ const Yarn = Implementation({
 
 type Manager = typeof NPM | typeof Yarn;
 
-export function getManager(dir: string): Manager {
-  const topPath = /^[a-z]+:\\$/gi;
-  
-  var isTopPath = topPath.test(dir);
+const WINDOWS_TOP_PATH = /^[a-z]+:\\$/gi;
+
+export function getManager(dir: string): Manager 
+{ 
+  const isTopPath = dir === "/" || WINDOWS_TOP_PATH.test(dir);
   
   const lock = path.join(dir, "yarn.lock");
   if (fs.existsSync(lock)) return Yarn;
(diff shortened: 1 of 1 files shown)


packages/core/src/packager/packager.ts
```diff
@@ -58,8 +58,12 @@ const Yarn = Implementation({
 type Manager = typeof NPM | typeof Yarn;
 
 export function getManager(dir: string): Manager {
+  const topPath = /^[a-z]+:\\$/gi;
```
thdxr, line 61: Can you hoist this outside this function and put it in a `const WINDOWS_TOP_PATH`

packages/core/src/packager/packager.ts
```diff
@@ -58,8 +58,12 @@ const Yarn = Implementation({
...
+  
+  var isTopPath = topPath.test(dir);
+  
   const lock = path.join(dir, "yarn.lock");
   if (fs.existsSync(lock)) return Yarn;
-  if (dir === "/") return NPM;
+  if (dir === "/"  || isTopPath) return NPM;
```
thdxr, line 67:
```suggestion
if (isTopPath) return NPM;
```

packages/core/src/packager/packager.ts
```diff
@@ -58,8 +58,12 @@ const Yarn = Implementation({
 type Manager = typeof NPM | typeof Yarn;
 
 export function getManager(dir: string): Manager {
+  const topPath = /^[a-z]+:\\$/gi;
+  
+  var isTopPath = topPath.test(dir);
```
thdxr, line 63:
```suggestion
const isTopPath = dir === "/" || WINDOWS_TOP_PATH.test(dir)
```

thdxr (changes requested):

thdxr:
@robertobasile84 Thanks for finding this! Left some minor style comments but can get this merged in once those are updated

jfriend00:
The last solution shown in this stackoverflow post https://stackoverflow.com/a/68977855/816620 is a bit more elegant and would also handle Windows UNC paths (paths that start with \\) or paths that don't contain a drive letter, but start with \.  That might be better than the regex.

PR #772, opened 2021-08-30
Prompt for default stage name

### LINKED ISSUE #691: CLI: Support automatically setting stage to sts identity
For customers that share a single AWS account with multiple developers for development we often get questions on how to make sure people's stacks don't conflict with each other.  The way to do this is to set `stage` to a unique value for each developer.

We should make it easy to set the stage value to a unique valued based on `sts` identity.  Here are some examples of what that looks like for various setups:

#### IAM
```
$ aws sts get-caller-identity
{
    "UserId": "AIDAIHW56K6VNFT4XPI2W",
    "Account": "112245769880",
    "Arn": "arn:aws:iam::112245769880:user/frank"
}
```

#### SSO
```
{
    "UserId": "AROAUCYUUGB2U7OZTDVN3:thdxr",
    "Account": "280826753141",
    "Arn": "arn:aws:sts::280826753141:assumed-role/AWSReservedSSO_AdministratorAccess_6475a72f96e8505e/thdxr"
}
```

Seems like it makes sense to use the `Arn` and pull out the last value past the `/`

In terms of how to enable this feature some options below

#### Explicit CLI Flag
```
sst start --self-stage
```
#### Special stage value
This might need to be more differentiated
```
sst start --stage=self
```

Would like some more ideas on how this can be defined

#### Requests
- https://serverless-stack.slack.com/archives/C01JG3B20RY/p1628855080492500

thdxr:
Closes #691

### DIFF
--- a/packages/cli/.gitignore
+++ b/packages/cli/.gitignore
@@ -1,3 +1,4 @@
 # Build outputs in test
 /test/**/.build/**
+/test/**/.sst/**
 /test/**/cdk.out/**
--- a/packages/core/src/state/state.ts
+++ b/packages/core/src/state/state.ts
@@ -1,0 +1,41 @@
+import fs from "fs";
+import path from "path";
+import { STS } from "aws-sdk";
+
+export function resolve(root: string, relative: string) {
+  return path.resolve(root, ".sst", relative);
+}
+
+export function init(root: string) {
+  const dir = directory(root);
+  if (fs.existsSync(dir)) return;
+  fs.mkdirSync(dir, { recursive: true });
+}
+
+export function directory(root: string) {
+  return resolve(root, "");
+}
+
+function resolveStage(root: string) {
+  return resolve(root, "stage");
+}
+
+export function getStage(root: string) {
+  const file = resolveStage(root);
+  try {
+    return fs.readFileSync(file).toString();
+  } catch {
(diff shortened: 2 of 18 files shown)


packages/core/src/state/state.ts
```diff
@@ -0,0 +1,41 @@
...
+    return null;
+  }
+}
+
+export function setStage(root: string, stage: string) {
+  const file = resolveStage(root);
+  fs.writeFileSync(file, stage);
+}
+
+export async function suggestStage() {
```
fwang, line 37: Should we default the stage name to something in case sts call failed?

packages/core/src/state/state.ts
```diff
@@ -0,0 +1,41 @@
...
+  const client = new STS();
+  const result = await client.getCallerIdentity().promise();
+  return path.basename(result.Arn!);
```
fwang, line 40: Do we have to allow "non-null" assertion in our eslint rules?

packages/cli/.gitignore
```diff
@@ -1,3 +1,4 @@
 # Build outputs in test
 /test/**/.build/**
+/test/**/.sst/**
```
fwang, line 3: Let's also add `.sst` to our starter templates.

PR #788, opened 2021-09-07
Preserve existing log group logical ids

thdxr:


packages/resources/src/util/apiGatewayV2AccessLog.ts
```diff
@@ -73,7 +74,9 @@ export function buildAccessLogData(
   } else {
     const root = scope.node.root as App;
     const apiName = root.logicalPrefixedName(scope.node.id);
-    logGroup = new logs.LogGroup(scope, "LogGroup" + apiStage.stageName, {
+    // Backwards compatibility, only suffix if not default stage
+    const logGroupName = "LogGroup" + (!multiStage ? "" : apiStage.stageName);
```
fwang, line 78: This is trivial, I feel naming this option `isDefaultStage` is more intuitive than `multiStage`. Also matches what the comment is saying. What do you think?

PR #789, opened 2021-09-07
Function: New api for default props

### LINKED ISSUE #726: Implement stack.setDefaultFunctionProps
#### Requests
- https://serverless-stack.slack.com/archives/C01JG3B20RY/p1629401113194700?thread_ts=1629392231.183700&cid=C01JG3B20RY
- And a million others

thdxr:
Closes #726

### DIFF
--- a/packages/resources/src/App.ts
+++ b/packages/resources/src/App.ts
@@ -18,6 +18,8 @@ import {
   CustomResourceProvider,
   CustomResourceProviderRuntime,
 } from "@aws-cdk/core";
+import { ILayerVersion } from "@aws-cdk/aws-lambda";
+import { Permissions } from "./util/permission";
 
 const appPath = process.cwd();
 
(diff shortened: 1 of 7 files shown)


packages/resources/src/App.ts
```diff
@@ -204,9 +206,25 @@ export class App extends cdk.App {
   setDefaultFunctionProps(
     props: FunctionProps | ((stack: cdk.Stack) => FunctionProps)
   ): void {
+    if (this.node.children.some((node) => node instanceof cdk.Stack))
+      throw new Error(
+        "Default function props for the app are already set. Please use app.addDefaultFunctionEnv or app.addDefaultFunctionPermissions to add more default properties"
+      );
```
fwang, line 212: This is checking if `setDefaultFunctionProps()` is called after a Stack has been added. But the error message "Default function props for the app are already set" sounds like it's implying `setDefaultFunctionProps()` has been called multiple time?

Do we need 2 checks for this:
- if `setDefaultFunctionProps()` has been called => throw "Default function props for the app are already set..."
- if `this.node.children.some(...)` => throw "Cannot call setDefaultFunctionProps after a stack has been created"

Not sure if this makes sense?

thdxr: I didn't update this copy from when I was checking to see if they were calling it only once.  I think we can just keep the second copy - I'm not sure if preventing them from calling it multiple times is something we need to worry about.

PR #807, opened 2021-09-11
Allow keepNames esbuild config option

KinnersleyM:
This PR allows for the [`keepNames`](https://esbuild.github.io/api/#keep-names) esbuild option.

It also allows for some extensibility for future esbuild options this project might want to support.

Error message for using an invalid option:
image

Without `keepNames`
```js
var RandomType = class RandomType2 {
};
```
With `keepNames`
```js
var RandomType = /* @__PURE__ */ __name(class RandomType2 {
}, "RandomType");
```

This fixes an issue I have been having with using the `@anatine/esbuild-decorators` plugin to use `"emitDecoratorMetadata": true` in my tsconfig. Prior to this, my `type-graphql` types were coming back as:
image
but with keepNames:
image


thdxr:
@KinnersleyM thanks for the PR - it looks good! There's a conflict that needs to be resolved but I'll merge it after then

KinnersleyM:
@thdxr Thank you! Just rebased now 👍

packages/cli/scripts/util/cdkHelpers.js
```diff
@@ -193,13 +193,17 @@ async function loadEsbuildConfigOverrides(configPath) {
   }
   const configOverrides = require(configFullPath);
 
-  // validate only "plugins" can be overrid
+  const allowedOptions = ["plugins", "keepNames"];
+
+  // validate only allowed options can be overridden
   const nonPluginsKey = Object.keys(configOverrides).find(
```
fwang, line 199: Let's rename `nonPluginsKey` to something more descriptive ie. `unallowedKey`

KinnersleyM: Sorted

fwang:
Thanks @KinnersleyM, will take a look at it today!

fwang:
@KinnersleyM with v0.34.7, you can now configure `keepNames` inline like this:
```
  bundle: {
    esbuildConfig: {
      keepNames: true,
    },
  },
```

Issue #805, opened 2021-09-10
Api: add `httpApiArn` property


fwang:
The CDK HttpApi construct does not provide an `arn` construct. It would need to be built like this:
```
arn:partition:apigateway:region::/apis/api-id
```

Let's add this property to `sst.Api`.

---
Requested by Adrian https://serverless-stack.slack.com/archives/C01HQQVC8TH/p1631258358391400

PR #809, opened 2021-09-12
Add ARN property to API stack

phausken:
This PR adds the public property `httpApiArn` to `sst.API` so that it can be referenced as needed.

Format is `arn:partition:apigateway:region::/apis/api-id`

Closes issue #805 

See Slack thread for more context:  https://serverless-stack.slack.com/archives/C01HQQVC8TH/p1631258358391400


thdxr:
Closes #805 

Looks good will merge this tomorrow

Issue #817, opened 2021-09-14
Allow deploying StaticSite during sst start


thdxr:
Add a `disablePlaceholder: boolean` option to StaticSite props. If turned on it should behave normally as though it's a deployment.

PR #820, opened 2021-09-15
StaticSitic: Add 'disablePlaceholder' boolean to StaticSite props

phausken:
When `sst start` is run, a placeholder static site is used. Depending on their particular use case or workflow, some users may not want this behavior, and want the static site to deploy like normally.

This resolves issue #817


fwang:
Hey @phausken, thanks for taking a crack at this :fire:

One quick thought on the naming of this flag. We currently have the `enableLiveDebug` flag for the `Function` construct, would it be more consistent if we named this `enableLiveDebugPlaceholder`? Or maybe `liveDebug.enablePlaceholder` if we are going to have more `sst start` options.

Curious to hear @thdxr 's thoughts as well.

PR #843, opened 2021-09-21
Function: Preserve original path when bundling js

thdxr:


packages/resources/src/util/nodeBuilder.ts
```diff
@@ -220,7 +222,8 @@ export function builder(builderProps: BuilderProps): BuilderOutput {
   let outCode, outHandler;
   if (bundle) {
     outCode = lambda.Code.fromAsset(buildPath);
-    outHandler = path.basename(handler);
+    outHandler = path.join(path.dirname(entryPath), path.basename(handler));
+    console.log(outHandler);
```
fwang, line 226: console.log monster

packages/resources/src/util/nodeBuilder.ts
```diff
@@ -220,7 +222,8 @@ export function builder(builderProps: BuilderProps): BuilderOutput {
   let outCode, outHandler;
   if (bundle) {
     outCode = lambda.Code.fromAsset(buildPath);
-    outHandler = path.basename(handler);
+    outHandler = path.join(path.dirname(entryPath), path.basename(handler));
```
fwang, line 225: Lambda needs the handler path in posix format ie. "/" this slash. We need to test deploying on Windows environment.

packages/resources/src/util/nodeBuilder.ts
thdxr: lol oops

PR #859, opened 2021-09-27
fix wrong path in template

miyamonz:
I found this when I created the ts template. I don't check another language but I think this fix seems right.


thdxr:
Thanks for catching this!

PR #890, opened 2021-10-05
Api: Remove Redundant Info from doc

NoahDavey:
Noticed while looking at API docs that this info seems no longer relevant. As I have been able to change the `path` attribute for my custom domain both in local dev and in my seed deployed environment


jayair:
Oh interesting, let me get @fwang to look at this.

PR #900, opened 2021-10-06
NextjsSite: allow using existing CloudFront cache policies

### LINKED ISSUE #899: NextJsSite: Share Cache Policies?
I have an account with about 5 Next.JS sites, and am in process of converting them to SST.  On the fifth site, due to multiple stages for each, I hit the limit of Cache Policies per account (20).  This is a hard limit, no requesting an increase, so my only option is to have the NextJsSite's share policies.

**At present, you can have no more than 6 NextjsSites per AWS Account, and this includes any preview style branching.**

I am looking for a recommendation, or possibly a feature, to make this easier than what I am currently seeing as a possibility: I would essentially have to recreate all additional behaviors from scratch and overwrite all of the ones built in the construct.  This wouldn't be a big deal until something changed, but I would definitely be wanting to avoid having to re-create every change simply to avoid this. Even in this case, though, the cache policies would still be created, causing the issue.

Thank you.

dkershner6:
Resolves #899 

- Allows for access to the `CachePolicyProps` used in all three cases for the `Distribution` through static props on the `NextjsSite` class.
  - This is how you would create the cache policies outside the construct.
- Created a new prop named `cachePolicies` on `NextjsSiteCdkDistributionProps` that allows for the use of cachePolicies that are created outside the construct.
- The construct will now use the outside policies if they exist, or otherwise use existing behavior, ensuring backwards compatability.

This PR is a proposal, and includes no testing, but can be added once a decision is made.

### DIFF
--- a/packages/resources/src/NextjsSite.ts
+++ b/packages/resources/src/NextjsSite.ts
@@ -50,14 +50,13 @@ export interface NextjsSiteFunctionProps {
   permissions?: Permissions;
 }
 
-export type NextjsSiteDomainProps = BaseSiteDomainProps;
-
 export interface NextjsSiteCachePolicyProps {
-  staticsCachePolicy?: cloudfront.ICachePolicy;
+  staticCachePolicy?: cloudfront.ICachePolicy;
   imageCachePolicy?: cloudfront.ICachePolicy;
   lambdaCachePolicy?: cloudfront.ICachePolicy;
 }
 
+export type NextjsSiteDomainProps = BaseSiteDomainProps;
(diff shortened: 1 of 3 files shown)


dkershner6:
(Commits)

- Made judgement call and called the prop `cfCachePolicies` rather than `cloudfrontCachePolicies` since it is right next to `cfDistribution`.
- Added two tests, one for behavior regarding the LACK of a `cfCachePolicies`, and one with a fully filled out `cfCachePolicies`.
- I didn't add any more tests or functionality since other items are still being decided.

fwang:
Hey @dkershner6, sorry for the delay. I like your suggestions. I'm going to take a look at this soon.

PR #918, opened 2021-10-12
Function: Allow disabling minification

thdxr:


pabacham:
@thdxr I just moved all node modules to externalModules and nodeModules and my tested lambda function package size increased from 400kb to 1.4Mb. I guess it is because that minify option applies only to the handler.

PR #1002, opened 2021-11-06
Add type definitions for Node and ensure that the major version is the same with default runtime

omer-to:
Since the nodejs runtime for TypeScript template is now 14, it would be beneficial and even in some cases saving from bugs to add @types/node as dev dependency with the latest version for major being 14.

Because often the type definitions for node is installed without specifying version, and in that case the latest is installed, and usually it is recommended to match the major version with the node version and types.


fwang:
Hey @omer-to, thanks for bringing this up.

@thdxr, the TS master, can you take a quick look at this PR?

PR #1005, opened 2021-11-06
Doc: add Thundra integration

rcoundon:
Added docs for integration Thundra APM. 
I also noticed that the Lumigo docs had the integration being deployed only for local running instead of only on deploys


www/docs/monitoring-your-app-in-prod.md
```diff
@@ -142,3 +142,43 @@ if (scope.local) {
...
+        this,
+        'ThundraLayer',
+        `arn:aws:lambda:${scope.region}:${thundraAWSAccountNo}:layer:thundra-lambda-node-layer:${thundraNodeLayerVersion}`,
+      );
+      this.addDefaultFunctionLayers([thundraLayer]);
+      
+      const env: Record<string, string> = {
+        THUNDRA_APIKEY: process.env.THUNDRA_API_KEY,
+      };
+      env.NODE_OPTIONS += ' -r @thundra/core/dist/bootstrap/lambda';
```
fwang, line 178: Can we rewrite this as?
```
  const env: Record<string, string> = {
    THUNDRA_APIKEY: process.env.THUNDRA_API_KEY,
    NODE_OPTIONS: "-r @thundra/core/dist/bootstrap/lambda";
  };
```

rcoundon: The only reason I had it the way I did was to prevent the overwriting of any existing NODE_OPTIONS values at the point the default function environment is applied. If that wouldn't happen, it makes sense to rewrite as you have it

fwang: Got it! Thanks Ross.

PR #1024, opened 2021-11-10
Permissions: Strongly typed / enums

### LINKED ISSUE #913: Permissions: Strongly typed / enums
It would be better to have something like 

```ts
fun.attachPermissions([
 [topic, PermissionType.grantPublish],
 [table, PermissionType.grantReadData],
]);
```
or 
```ts
fun.attachPermissions([
 [topic, PermissionType.PUBLISH], // or PermissionType.GRANT_PUBLISH
 [table, PermissionType.READ], // or PermissionType.GRANT_READ
]);
```

instead of
```ts
fun.attachPermissions([
 [topic, "grantPublish"],
 [table, "grantReadData"],
]);
```

so as to avoid deploy-time issues and enhance developer experience in their IDE

junaway:
Fixes #913.

### DIFF
--- a/packages/resources/src/util/permission.ts
+++ b/packages/resources/src/util/permission.ts
@@ -19,13 +19,107 @@ import {
 } from "../index";
 import { isConstructOf } from "./construct";
 
+
+type ApiGatewayV1ApiGrant = // @aws-cdk/aws-apigateway/lib
+  | "grantRead"
+  | "grantWrite"
+  | "grantReadWrite";
+
+type AppSyncApiGrant = // @aws-cdk/aws-appsync/lib
+  | "grantMutation"
+  | "grantQuery"
+  | "grantSubscription";
+
+type BucketGrant = // @aws-cdk/aws-s3/lib
+  | "grantRead"
+  | "grantWrite"
+  | "grantPut"
+  | "grantPutAcl"
+  | "grantDelete"
+  | "grantReadWrite";
+
+type EventBusGrant = // @aws-cdk/aws-events/lib
+  | "grantPutEventsTo";
+
+type FunctionGrant = // @aws-cdk/aws-lambda/lib
+  | "grantInvoke";
+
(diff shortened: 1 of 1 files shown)


thdxr:
I'm not sure if we want to have to maintain a full list of these on our end. Even if we maintain a list it should be inferred from the underlying type so we can get errors if the underlying CDK changes.

Also I sort of wonder if we can just change the API to pass the function that's supposed to be called eg
```
permissions: [EventBus.grantPutEventsTo]
```

thdxr:
Let me take a crack and defining an inferred type

thdxr:
Some weird typescript for you
```
  type GrantFunc = `grant${string}`;
  type GrantExtract<T, K = keyof T> = K extends GrantFunc ? K : never;
  
  type EventBusGrant = GrantExtract<typeof events.EventBus>;
```

`EventBusGrant` evaluates to `"grantPutEvents" | "grantAllPutEvents"`

thdxr:
You can use the above method for now but hypothetically there is a way for them to pass in _any_ construct + function pair and have the typings + autocomplete work.  Don't have time to get that working at the moment.

Longer term I do wonder if this `[construct, string]` api is a good one. I feel like we might as well ask them to pass in the function right?

junaway:
The issue is that not all the functions that look like `grantSomething` behave the same way. Some Receive more arguments than others.

thdxr:
Ah ok.  You can also use typing to filter out function signatures that match what we support, but further hints at the idea that this maybe is the wrong api if we're not supporting all `grant` functions

junaway:
Yeah. It's good enough for most grants which is great since we can always use other permission formats.
But it's true that we can always imagine other ways to handle permissions.
In a trade-off between speed-and-simplicty over quality-and-complexity, you're the one deciding :D

thdxr:
Yeah let's hold on this until I can chat with Frank and Jay on the long term direction we want to go

53jk1 (approved):

thdxr:

PR #1030, opened 2021-11-13
Function: Enable source maps by default using NODE_OPTIONS

### LINKED ISSUE #984: Enable sourcemaps by default
Enable sourcemaps by adding `NODE_OPTIONS=--enable-source-maps`. We should try and detect if they're already passing in NODE_OPTIONS as an env variable and append to it rather than replace

junaway:
Fixes #984.


fwang:
Looks good!

Let's just add 1 more test case for users already passed in `--enable-source-maps`, and make sure `--enable-source-maps` doesn't show up twice.

For example:
```js
  new Function(stack, "Function", {
     handler: "test/lambda.handler",
     runtime: lambda.Runtime.NODEJS_10_X,
     environment: {
       NODE_OPTIONS: "--enable-source-maps --flag --key=value",
     }
   });
```

thdxr:

PR #1070, opened 2021-11-21
added example for angular

Manitej66:


umarmohammed:
Hi, I was looking for examples for Anglular and came across this PR.  An extra step I had to do was add a `seed.yml` file to the root directory with the following
```
before_build:
  - cd frontend && npm install
 ```

Is this required or am I missing something else? 

Thanks.

jayair:
@umarmohammed Yeah we do the same for React as well: https://serverless-stack.com/chapters/deploying-through-seed.html

Issue #1091, opened 2021-11-29
Debug stack creates S3 bucket that allows public access


krivachy:
The deployed debug stack's S3 bucket fails our Security Hub checks, specifically:

```
[S3.8] S3 Block Public Access setting should be enabled at the bucket level
Category: Protect > Secure access management > Access control
Severity: High
Resource type: AWS::S3::Bucket
AWS Config rule: s3-bucket-level-public-access-prohibited
Schedule type: Change triggered 
```

[Link](https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#fsbp-s3-8)

I don't see a need for the Debug Stack's S3 bucket to ever allow public access so it should be a simple fix to block it and make sure it never accidentally exposes resources.

PR #1092, opened 2021-11-29
Block public access for DebugStack S3 bucket

krivachy:
Closes: #1091


krivachy:
@fwang do you have any thoughts on this change?

thdxr:
Nope looks good to me

krivachy:
@thdxr it seems like this PR is only missing a label (I can't seem to add it) and then it can be merged?

PR #1138, opened 2021-12-10
added external custom domain support to Api construct

Manitej66:
Added support for adding custom domains to the `Api` construct that are outside of Route 53. 

Sample code example:

```ts
const api = new sst.Api(this, "Api", {
  customDomain: {
    domainName: "api.sst.sh",
    isExternalDomain: true,
    certificate: acm.Certificate.fromCertificateArn(
      this,
      "Certificate",
      "arn:aws:acm:us-east-1:000639264671:certificate/xxxx"
    ),
  },
  routes: {
    "GET /": "src/lambda.main",
  },
});
```


JonHolman:
Thank you @Manitej66 and @fwang

Issue #1113, opened 2021-12-04
Possible to support using NodejsFunction?


revmischa:
I have been using the esteemed NodejsFunction construct for many months now in production. It's really terrific, wraps ESbuild and adds nice shortcuts for features like sourcemaps. Now I'm trying to port my app to SST and it's a sore point.

👉🏻   https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-lambda-nodejs.NodejsFunction.html

I really miss some of the options this provides, along with the fact that it will use esbuild if installed instead of bundling with docker. This is much faster and nicer. There are a host of great reasons to use this construct.

I understand that SST adds the debugging magic and needs its own Function which is totally cool, but maybe it's possible to subclass NodejsFunction when doing this? I can try to investigate it myself if you point me in the right direction.

thdxr:
We actually use esbuild as well and do not build in Docker.

Are there specific features you're particularly interested in using? We tend to add them one by one as people need them

revmischa:
Right now I use the bundling options:
* https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-lambda-nodejs.BundlingOptions.html#metafile
* https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-lambda-nodejs.BundlingOptions.html#sourcemap
* https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-lambda-nodejs.BundlingOptions.html#sourcemapmode

revmischa:
I'm trying to just hack this together really quickly right now but I have managed to get this started:

```ts
/* eslint-disable prefer-const */
import { Runtime, Tracing } from "@aws-cdk/aws-lambda"
import {
  BundlingOptions,
  NodejsFunction as CdkNodejsFunction,
  NodejsFunctionProps as CdkNodejsFunctionProps,
} from "@aws-cdk/aws-lambda-nodejs"
import { Construct, Duration } from "@aws-cdk/core"
import { Function, FunctionBundleObject, FunctionProps, Stack } from "@serverless-stack/resources"
import { existsSync } from "fs"
import path from "path"

export interface NodejsFunctionProps extends CdkNodejsFunctionProps {
  grantDatabaseAccess?: boolean
  handler?: string
  srcPath?: string
}

/**
 * Compatibility layer between SST.Function and NodejsFunction
 */
[...]

PR #1156, opened 2021-12-15
RFC: NodejsFunction

revmischa:
Following up on #1113 -

I like NodejsFunction from AWS because it's from AWS and does almost everything I want, except for keeping track of default function options...

It has a lot of bundling options like fine-grained source map control, every esbuild option supported, custom hashing, unobtrusive builds that don't generate scripts to run inside my packages (which causes issues like #1154 for me), and it's semi-supported by the official AWS CDK team.

I created a NodejsFunction construct that I'm using to help migrate my existing application to SST. Maybe it's possible to refactor it with Function and give SST users their choice of construct while using the lovely function defaults that SST provides. I'm opening this PR to get some feedback on this idea.

### DIFF
--- a/packages/resources/package.json
+++ b/packages/resources/package.json
@@ -85,6 +85,7 @@
     "@aws-cdk/assert": "1.132.0",
     "@types/cross-spawn": "^6.0.2",
     "@types/fs-extra": "^9.0.6",
+    "@types/lodash.clonedeep": "^4.5.6",
     "@types/node": "^14.0.27",
     "esbuild-runner": "^2.2.1"
   },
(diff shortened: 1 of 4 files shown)


Dzhuneyt:
Any movement on this @thdxr @revmischa ? Looks like this effort was abandoned.

thdxr:
In SST 2.0 we expose esbuild directly so you can override everything you want

revmischa:
I am pumped for this, thank you for all the work you are doing @thdxr even if you disagree with my wisdom on leveraging NodejsFunction.

PR #1162, opened 2021-12-16
Clear artifacts per function

thdxr:


revmischa:
A little hard to tell for me, does this mean it won't delete my lambda layer artifacts between deployments?

thdxr:
The generated layer is done through CDK and that ends up being deleted by CDK on every run so I don't think that it'll stick around

revmischa:
> The generated layer is done through CDK and that ends up being deleted by CDK on every run so I don't think that it'll stick around

If I use CDK without SST it doesn't delete any of the assets. If my layer assetHash doesn't change then it doesn't rebuild it, if it changes or I manually delete my cdk.out directory then it rebuilds the layer.

PR #1200, opened 2021-12-30
Display last update time on docs

micalevisk:
Now we'll get something like this _(without author's name)_:

image

so we can easily see if the docs has been updated since the last time :)

I didn't tested this locally, but I've followed this guide: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs


jayair:
Good idea!

PR #1213, opened 2022-01-03
fix: upload loading sign shows when creating new folder

Manitej66:


thdxr (changes requested):
I'm not sure we need to introduce another piece of state here.  `isCreating` represents the open state of the input although we should rename it to be something more specific.

Manitej66:
> I'm not sure we need to introduce another piece of state here. `isCreating` represents the open state of the input although we should rename it to be something more specific.

We are using the same query to upload both files & folders. We need 2 states, 1 to differentiate folder from file & the other to hold the state of input state, else the loading sign will be shown at 2 places when we upload a file.

thdxr:
I'll take a look at this, whenever there's a lot of `useState` it's usually a sign of something wrong

PR #1215, opened 2022-01-03
fix: show sign if bucket is empty

Manitej66:


thdxr (changes requested):
Delete code instead of commenting - bit hard to tell what exactly changed here.  Also I'm pretty sure we did key count < 1 on purpose because it === 1 when a subfolder is empty

Manitej66:
> Delete code instead of commenting - bit hard to tell what exactly changed here.  Also I'm pretty sure we did key count < 1 on purpose because it === 1 when a subfolder is empty

Yeah, if there exists a subfolder then the bucket is not empty right? Previous logic returns `isEmpty` to true when key count is 1.

thdxr (approved):

PR #1216, opened 2022-01-03
fix: invocation log is cutoff on the right

Manitej66:


thdxr (changes requested):
I don't think horizontal scrolling is a good fix here, the behavior should be to wrap when it overflows.  Also do not use inline `css` or `style`

Manitej66:
I'll take a look again

thdxr (approved):

PR #1217, opened 2022-01-03
fix: Leave the function request body after an invocation

Manitej66:


thdxr (changes requested):
You can delete this line of code, generally prefer deleting to commenting since git preserves all history

thdxr (approved):

PR #1225, opened 2022-01-04
fix: Error fetching metadata when sst start boots up

Manitej66:


thdxr:
I actually fixed this already - the issue Frank was seeing was because the WS server was available immediately but the proxy was not.  I started them both together in the CLI - however this PR is a good change anyway

PR #1233, opened 2022-01-05
fix: do not use a remote svg image in preview

Manitej66:


thdxr (changes requested):
Do you think we can use one of the react icons here?

Manitej66:
> Do you think we can use one of the react icons here?

Yeah we can, but we're changing the src value dynamically so I kept the local image.

We can conditionally render icon vs image, but changing the image src dynamically makes more sense i believe

thdxr:
I think we should swap out the element because for SVGs rendering them in an `img` element isn't entirely right.  react-icons runs all the svgs through SVGR which places the svg directly on the page

packages/console/src/App/Stage/Buckets/Detail.tsx
```diff
@@ -519,16 +520,18 @@ export function Detail() {
...
+          {selectedFile.data.info.ContentType?.startsWith("image") &&
+          !isFileSizeTooLargeToPreview(
+            selectedFile.data.info.ContentLength!
+          ) ? (
+            <Image src={selectedFile.data.url} />
+          ) : (
+            <ImFileEmpty
+              size={180}
+              color="#e27152"
+              style={{ margin: "0 auto" }}
```
thdxr, line 532: Avoid inline styles, wrap this in a styled component

thdxr (changes requested):

packages/console/src/App/Stage/Buckets/Detail.tsx
```diff
@@ -519,16 +524,16 @@ export function Detail() {
...
-                : "https://img.icons8.com/ios/12/e27152/file.svg"
-            }
-          />
+          {selectedFile.data.info.ContentType?.startsWith("image") &&
+          !isFileSizeTooLargeToPreview(
+            selectedFile.data.info.ContentLength!
+          ) ? (
+            <Image src={selectedFile.data.url} />
+          ) : (
+            <Placeholder>
```
thdxr, line 533: Sorry I actually meant you can wrap the icon directly like this

```
const Placeholder = styled(ImFileEmpty, { ... })
```

thdxr (changes requested):

PR #1261, opened 2022-01-13
Update cdkHelpers to verify CDK mismatch versions

fwang:


packages/cli/scripts/util/cdkHelpers.js
```diff
@@ -285,6 +214,119 @@ async function reTranspile() {
...
+function formatDepsForInstall(depsList, version) {
+  return depsList
+    .map((dep) =>
+      isCdkV2CoreDep(dep) ? `${dep}@${version}` : `${dep}@${version}-alpha.0`
+    )
+    .join(" ");
+}
+
+function isCdkV2CoreDep(dep) {
+  return dep === "aws-cdk" || dep === "aws-cdk-lib";
```
Fyko, line 319:
```suggestion
  return ["aws-cdk", "aws-cdk-lib"].includes(dep);
```

packages/cli/scripts/util/cdkHelpers.js
```diff
@@ -285,6 +214,119 @@ async function reTranspile() {
...
+    const devDepString = formatDepsForInstall(mismatchedDevDeps, cdkVersion);
+    logger.info(
+      usingYarn
+        ? `  yarn add ${devDepString} --dev --exact`
+        : `  npm install ${devDepString} --save-dev --save-exact`
+    );
+  }
+
+  logger.info(
+    `\nLearn more about it here — https://github.com/serverless-stack/serverless-stack#cdk-version-mismatch\n`
```
aisflat439, line 287: Hey Frank. I encountered this today. This link doesn't go anywhere meaningful. I looked around but couldn't find a link that is more appropriate than `https://docs.serverless-stack.com/known-issues`. However, this doesn't seem like teh correct link either.

fwang: Hey @aisflat439, nice catch! `https://docs.serverless-stack.com/known-issues` is the correct link.

PR #1267, opened 2022-01-16
Update esbuild and remove sourcemaps

thdxr:


paambaati:
@thdxr Would you mind giving us context on why this was removed? We're still on an older version, and we're planning an upgrade path.

PR #1276, opened 2022-01-19
Dependencies moved to dev/peer Dependencies where appropriate

### LINKED ISSUE #1151: ESLint, Jest, TSJest, Typescript should be devDependencies and (maybe) peerDependencies in cli and resources
https://github.com/serverless-stack/serverless-stack/blob/3985a15a2f02b281fa4a9e8c41e3005444a3c62e/packages/resources/package.json#L70-L82

Example is for resources.

These dependencies cause downstream projects to install multiple versions of each of these (unless they exactly match) which causes conflicts.  There are some less concerning ones that could probably also be devDependencies like `chalk`, `graphql` and `@types`, but they are less likely to cause problems.

I don't know of any downside to this change.

dkershner6:
Resolves #1151 

This prevents folks from needing to `--force` update and install when using versions different from `serverless-stack` packages.

This is technically a breaking change for anyone who does not have ESLint, Jest, or Typescript installed explicitly, although they will receive `peerDepedency` warnings telling them to do so, a common occurrence in Node development.


dkershner6:
RE: `peerDependencies`, I checked what Vercel does with Next.js because I was curious, and they do Linting and Type-Checking as part of their build.

They actually list those items in no way that I can see. This can make some sense, as `peerDependencies` are generally only filled by `dependencies` in the installing project, and things like ESLint, TypeScript, Jest are usually `devDependencies`.

PR #1281, opened 2022-01-19
Update readme examples link

aisflat439:
Currently the path is incorrect for the link:

image

This changes it to this instead:

image


jayair:
Good catch!

PR #1327, opened 2022-02-03
feat(StaticSite): Optional atomic deploys

### LINKED ISSUE #1323: [StaticSite]: Can we make deployment prefixes "optional"?
## In-place Deploys: (3mins - avg build time)
In the past we have never really been concerned about true blue-green deploys and so out production deploys of our static sites would just use the aws cli to sync and remove orphans with each deploy.

## sst.StaticSite Deploys: (16mins - avg build time)
Contrasting this to us now using `sst.StaticSite` and its use of `deployid` as a bucket prefix. 
This now means that the CloudFront distro needs to be updated (origin path) so that adds 6-15mins of replication/deploy time.

## Optional prefixes and us
With optional prefixes, I would like to write a CDK construct that [handles Cloudfront distro cutovers](https://serverfault.com/a/752140) and **true** blue-green deploys. We would only do that for production, while dev deploys stay spicy fast. Even without this construct… one could at lease make the deploy prefixes optional for Prod only, again, keeping dev deploy fast.

Happy to write and present a PR for this if you're agreeable to it.

iDVB:
Fixes #1323

### DIFF
--- a/packages/resources/assets/BaseSite/custom-resource/s3-handler.py
+++ b/packages/resources/assets/BaseSite/custom-resource/s3-handler.py
@@ -88,9 +88,15 @@ def handler(event, context):
             aws_command("s3", "rm", s3_dest, "--recursive")
 
         if request_type == "Update" or request_type == "Create":
+            if dest_bucket_prefix == "":
+                aws_command("s3", "rm", s3_dest, "--recursive")
+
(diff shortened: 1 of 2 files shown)


fwang:
Deploy prefix is removed in v0.61.3

PR #1351, opened 2022-02-09
added precompiling in middy example

Manitej66:


examples/middy-validator/.gitignore
```diff
@@ -22,3 +22,4 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+yarn.lock
```
fwang, line 25: Do not add yarn.lock to .gitignore. Remove yarn.lock before committing.

PR #1361, opened 2022-02-14
Docs: making Aurora RDS Database example costless

mariovalney:


fwang:
@mariovalney Thanks for pointing this out! We added an [sst.RDS](https://docs.serverless-stack.com/constructs/RDS). And we took ur feedback and made the default VPC configuration to launch the cluster in ISOLATED subnet.

Also updated the example to use the RDS construct.

Thanks again!

PR #1396, opened 2022-02-21
Prevent loading jsdom, which mocks fetch/XHR

hermanbanken:
In a test I'm using a library that uses Axios, this _should_ be fine... but

somehow I get an error indicating that a `OPTIONS` request gets a 405 error. This is curious, as this library should not do a OPTIONS request. After viewing the stacktrace, it became apparent that jsdom had stubbed the nodejs-native request functionality, tricking Axios in using that. This is insane, that in a SST app jsdom is loaded (a React utility).

So this PR sets the testEnvironment to `node` to prevent loading jsdom.

### DIFF
--- a/packages/cli/scripts/util/createJestConfig.js
+++ b/packages/cli/scripts/util/createJestConfig.js
@@ -11,6 +11,7 @@ const paths = require("./paths");
 module.exports = (resolve, rootDir) => {
   const config = {
     collectCoverageFrom: ["./**/*.{js,jsx,ts,tsx}"],
+    testEnvironment: "node",
     testMatch: [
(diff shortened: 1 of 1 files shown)


hermanbanken:
Alternatively, `testEnvironment` could be made configurable. It is not currently.

fwang:
@hermanbanken thanks for bringing this up! @thdxr can you take a quick look at this?

hermanbanken:
If required I can try to make a minimal reproduction. Let me know if that is required. Regardless, the mode should be node right?

fwang:
That'd be great.

Yeah, the change makes sense. @thdxr knows the jest stuff better, I will let him take a second look.

PR #1423, opened 2022-02-24
Api: Allow lambda.function 1078

danieloi:
❗️ This solution requires the user import `lambda.Function` (named as `cdkLambda.Function`) we re-export from `aws-cdk-lib/aws-lambda` for `instanceof` to work internally.

Guards were placed in the console UI so it doesn't fail when lambda.Function is used in live development sessions.

`Fn.attachPermissions` was replaced with `attachPermissionsToRole` calls so things would continue working for both `sst.Function` and `lambda.Function`


fwang:
Hey @danieloi, thanks for the putting this together.

Design-wise, we want to treat `lambda.Function` as an escape hatch when ppl want to use a runtime we don't support (ie. Java, DockerFunction). So let's:
- not show `lambda.Function` not show up in the console
- not support `attachPermissions`
- `Api.getFunction` only return `sst.Function`

danieloi:
@fwang the `lambda.function` calls actually don't show up in the console.

I display this message in the console: `Use the sst.Function Construct for the best console experience` when a route with a cdk function is selected.

Also added some default values and optional chaining otherwise the whole UI crashes when the `Functions` page is loaded in this l`ambda.Function` case.


I've removed the `attachPermissions` calls and modified `getFunction` as requested.

Is this ok?

packages/resources/src/Api.ts
```diff
@@ -576,13 +579,22 @@ export class Api extends Construct implements SSTConstruct {
...
     // Create Function
     ///////////////////
-    const lambda = Fn.fromDefinition(
-      scope,
-      `Lambda_${postfixName}`,
-      routeProps.function,
-      this.defaultFunctionProps,
-      `The "defaultFunctionProps" cannot be applied if an instance of a Function construct is passed in. Make sure to define all the routes using FunctionProps, so the Api construct can apply the "defaultFunctionProps" to them.`
-    );
+    let lambda: cdkLambda.Function | Fn;
```
fwang, line 582: Let's duplicate the `createFunctionIntegration` function, so have `createSstFunctionIntegration` and `createLambdaFunctionIntegration`. This way, we don't have to keep checking `instanceof Fn`.

packages/resources/src/Construct.ts
```diff
@@ -21,7 +21,7 @@ export interface SSTConstruct {
 
 export function getFunctionRef(fn?: any) {
   if (!fn) return undefined;
-  if (!(fn instanceof Fn)) return undefined;
+  if (!(fn instanceof lambda.Function)) return undefined;
```
fwang, line 24: Let's check for `sst.Function` here, so the `getFunctionRef` returns undefined for `lambda.Function`. This way, we can probably remove all the handling on the frontend.

packages/resources/src/Api.ts
danieloi: done

packages/resources/src/Construct.ts
danieloi: done!

danieloi: I changed it but I still need the frontend handling.

danieloi:
Fixing up the tests rn

dhrrgn:
@danieloi @fwang First, I apologize if this isn't the best place for this, but there are a few issues around this. Let me know if you want to move this discussion elsewhere. 

We are evaluating migrating from Serverless Framework to CDK, and SST looks perfect, except for the lack of custom runtime support.

Most of our functions use the `provided.al2` runtime, including api handlers, SQS consumers, scheduled functions (crons), and others. So, unfortunately, this solution only solves part of our problem. 

To solve it globally, my idea is this:

- Add a `SSTFunction` interface, and implement it on `Function`, and add a new `CdkFunction` construct that extends `lambda.Function` and implements the interface.
- Update `FunctionDefinition` to use the interface.
- Profit?

I am obviously over-simplifying things, but this should allow sending a `CdkFunction` to anywhere a function is expected.

The new `CdkFunction` could always have live dev disabled to avoid complications (could add support later). I think people using custom functions/runtimes would understand this limitation.

Thoughts?

danieloi:
> @danieloi @fwang First, I apologize if this isn't the best place for this, but there are a few issues around this. Let me know if you want to move this discussion elsewhere.
> 
> We are evaluating migrating from Serverless Framework to CDK, and SST looks perfect, except for the lack of custom runtime support.
> 
> Most of our functions use the `provided.al2` runtime, including api handlers, SQS consumers, scheduled functions (crons), and others. So, unfortunately, this solution only solves part of our problem.
> 
> To solve it globally, my idea is this:
> 
> * Add a `SSTFunction` interface, and implement it on `Function`
> * Add a new `CdkFunction` construct that extends `lambda.Function` and implements the interface.
> * Update `FunctionDefinition` to use the interface.
> * Profit?
> 
> I am obviously over-simplifying things, but this should allow sending a `CdkFunction` to anywhere a function is expected.
> 
> The new `CdkFunction` could always have live dev disabled to avoid complications (could add support later). I think people using custom functions/runtimes would understand this limitation.
> 
> Thoughts?

Did your strategy work? 

The best place to hear back quickly is actually the discord https://discord.gg/sst

fwang:
Adopted the implementation for SST v1 in #1940

PR #1439, opened 2022-02-27
fix ctrl-c twice issue

danieloi:
it had to do with this

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0a2f0c574ce4fa573ef4e85f8c98f90c2fdf683a/types/node/readline.d.ts#L372


fwang:
Released this in v0.66.2. Thanks @danieloi!

PR #1444, opened 2022-02-28
Clean up dependencies

thdxr:


davideicardi:
With this PR it seems that `jest` is not more included automatically and I should add it manually in my dependency. Is this by design?

davideicardi:
With this PR it seems that `jest` is not more included automatically and I should add it manually in my dependency. Is this by design?

thdxr:
This was an error, I'm adding it back in

PR #1476, opened 2022-03-07
Added example for Angular

### LINKED ISSUE #1062: Example: StaticSite + Angular

Manitej66:
closes #1062


kuncevic:
Hey guys, I've been noticing several issues related to Angular support https://github.com/sst/sst/issues?q=angular. The latest one is being merged at https://github.com/sst/sst/pull/1476. However, concerning the homepage and the docs site, it appears there are no indications that Angular is supported. I wonder if there anything is missing and if so how can we make it to happen?

jayair:
We could definitely use some help here since we aren't Angular experts. But since this post our setup has matured quite a bit. We have an SSRSite construct that does most of the heavy lifting and all the others Remix, Astro, Svelte, etc. implement that. So a good starting point would be poking around those and putting together a prototype. We can also setup a channel in our Discord to get others involved once we have something working.

PR #1523, opened 2022-03-14
Add `cdk.context.json`  to .gitignore template

urugator:
The file was generated when specifying `customDomain`.  There are other CDK related files, dunno if they should be added as well:
https://github.com/aws/aws-cdk/blob/5bad3aaac6cc7cc7befb8bdd320181a7c650f15d/.gitignore#L32-L36

### DIFF
--- a/packages/create-serverless-stack/templates/javascript/.template.gitignore
+++ b/packages/create-serverless-stack/templates/javascript/.template.gitignore
@@ -16,6 +16,9 @@
 .build
 .sst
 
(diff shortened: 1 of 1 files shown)


thdxr:
CDK actually recommends that you check this into the repo to avoid changing configurations from breaking your deployments

urugator:
Yes, I realized that later, therefore closed the PR. The initial google results suggested to gitignore it.

PR #1541, opened 2022-03-17
StaticSite: support removing old files

fwang:


kernwig:
Since these tickets are all that describe the change in the release notes, some description of the impact would be greatly appreciated.

fwang:
Hey @kernwig, yeah that makes sense! Just updated the [release note](https://github.com/serverless-stack/serverless-stack/releases/tag/v0.69.1) with more details, and announced in Slack.

Will make sure to be more transparent about release changes.

PR #1553, opened 2022-03-18
Fix invalid typescript monorepo example name

ilumin:


fwang:
Nice catch. Thanks @ilumin!

PR #1572, opened 2022-03-23
Update consumer.js

adsc:
Just a small change to show new Users of SQS that the payload of the message is in the event passed to the consumer.


fwang:
Hey @adsc, if you are using the SST Console, the `event` object is being printed to the console by default. Is that what you are looking for?

For example:

thdxr:

PR #1621, opened 2022-04-13
Use ExpressionAttributeNames instead of the key directly

kujtimiihoxha:
Currently using special column names like `_type` does not work in the console, this PR should fix that.


thdxr:
Great fix! Thank you

PR #1622, opened 2022-04-13
Use ExpressionAttributeNames instead of the key directly

kujtimiihoxha:
V0 PR for https://github.com/serverless-stack/serverless-stack/pull/1621


thdxr:
The console is actually only deployed from v1 so can close this one

PR #1660, opened 2022-04-26
Handle JSON encoding errors in bootstrap.py

estyrke:
Try to do something informative when the handler returns something that is not JSON serializable. Previous behavior resulted in bootstrap.py crashing, leading to a timeout in the client.

Another option would be to just serialize the unserializable part to something that is easy to recognize such as `<unserializable bla bla bla...>`, but it felt more explicit to throw an error.


fwang:
Thanks @estyrke!

PR #1672, opened 2022-05-05
NextJSSite: Added CloudFront Origin Req Policy override

### LINKED ISSUE #1531: NextjsSite: allow using existing CloudFront Request Origin policies
Much like in https://github.com/serverless-stack/serverless-stack/pull/900 there should be a way to share Request Origin policies across stacks

There's currently a hard limit of 20 off these across a single AWS account causing rate limits when several branches are being deployed at once.

remotealex:
As mentioned in #1531, there is a hard limit of 20 Origin Request Policies on a single AWS account. This causes rate limits when several branches are being deployed at once as when using `next/image`, the NextJSSite constructor creates a new one for each PR.

This fixes #1531 by adding a custom override property (`imageOriginRequestPolicy`) to the NextJSSite constructor. Now users can share AWS Request Origin Policies across stacks.

### DIFF
--- a/packages/resources/src/NextjsSite.ts
+++ b/packages/resources/src/NextjsSite.ts
@@ -58,7 +58,7 @@ export interface NextjsSiteProps {
     /**
      * Override the default CloudFront cache policies created internally.
      * @example
-     * ### Reusng CloudFront cache policies
+     * ### Reusing CloudFront cache policies
      *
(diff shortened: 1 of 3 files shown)


fwang:
Thanks @remotealex, I had to make a minor change and wasn't able to push to your branch (permission error). I created a new branch/PR here #1744 and merged.

PR #1678, opened 2022-05-07
Update migration.md

shechter:
I had to remove 'run' from npx command


fwang:
Thanks @shechter, nice catch!

PR #1690, opened 2022-05-11
ISSUE-831 This addition makes the WebSocketRoute accessible through the API 

masahji:
so that one can easily work around features not yet added to SST or even do custom things.


fwang:
Thanks @masahji, I had to make a minor tweak and wasn't able to push to your branch (permission error). I created a new branch/PR here #1742 and merged.

PR #1692, opened 2022-05-12
Fixed import typo in expo-app example

kevin86wright:


fwang:
Nice catch @kevin86wright! Fixed in the example template.

PR #1700, opened 2022-05-16
Preset update

Manitej66:


packages/create-sst/bin/presets/examples/rest-api-python/preset.mjs
```diff
@@ -1,3 +1,3 @@
 import { extend, extract, install } from "create-sst";
 
-export default [extend("presets/base/example"), extract()];
+export default [extend("presets/starters/typescript-starter"), extract()];
```
fwang, line 3: Just case across this, should we use the python-starter/go-starter/etc for non-typescript projects?

Manitej66: fixed

PR #1890, opened 2022-07-18
Allow specifying ESM/CJS compat shim contents

revmischa:
This is related to my issue here: https://discord.com/channels/983865673656705025/996271841897349132

It's not really a solution because I don't want to have to essentially hardcore `__dirname` to `/var/task` but this is blocking me now and others may have a similar need for more robust/custom shims.

---

My issue is that SST doesn't put the bundle at the top level and prisma tries to look in `__dirname` for the `schema.prisma` file.

> ideally i could either choose to use a flat bundle structure or have a way to copy a file into the same dir as the output bundle

### DIFF
--- a/packages/core/src/runtime/handler/node.ts
+++ b/packages/core/src/runtime/handler/node.ts
@@ -12,6 +12,11 @@ const require = createRequire(import.meta.url);
 
 const BUILD_CACHE: Record<string, esbuild.BuildResult> = {};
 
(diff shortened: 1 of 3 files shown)


revmischa:
The new `banner` option solves my immidiate need but I still would love the ability to copy a file to the bundle output dir @thdxr instead of hardcoding `__dirname` in my shim

PR #1891, opened 2022-07-18
Added custom 'execute' method to the 'createGQLHandler' method's 'HandlerConfig'

abmohan:
This will allow us to incorporate plugins like [this](https://pothos-graphql.dev/docs/plugins/authz), which need to [wrap the execute function similar when calling graphql-helix](https://github.com/hayes/pothos/blob/35f123da076fda2575c17da181680695ac9dddf9/packages/plugin-authz/tests/example/server.ts).


jonnedeprez:
I would be very interested in this feature. :+1:

thdxr:

PR #1892, opened 2022-07-19
Update Cron construct documentation

abirtley:
As far as I can tell, it should be `enabled: !app.local` if you want to "[d]isable the cron job from automatically running while developing" ? Otherwise it's only enabled if running locally.


fwang:
Of course! Thanks @abirtley.

PR #1894, opened 2022-07-20
Examples: Remix app workspace corrections

simonireilly:
## Changes

```sh
yarn create sst --template examples/remix-app
```

Would fail to fetch node modules from `web` workspace because it was missing a name in the workspace scope, and a version.

Some readme updates to remove svelte references.


fwang:
Thanks @simonireilly!!

PR #1935, opened 2022-08-09
Function: allow using @ edge

mikejpeters:
[If the stack is in `us-east-1`, a "normal" `lambda.Function` can be used instead of an `EdgeFunction`](https://docs.aws.amazon.com/cdk/api/v1/docs/aws-cloudfront-readme.html#lambdaedge).

However, the SST Function construct sets some environment variables, which [are not supported @ edge](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/edge-functions-restrictions.html#lambda-at-edge-runtime-restrictions).

Enabling the `removeInEdge` option for these environment variables solves the problem

### DIFF
--- a/packages/resources/src/Function.ts
+++ b/packages/resources/src/Function.ts
@@ -960,8 +960,8 @@ export class Function extends lambda.Function implements SSTConstruct {
     const { config } = this.props;
 
     // Add environment variables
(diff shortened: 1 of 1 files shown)


fwang:
Release in v1.6.9

Thanks @mikejpeters!

PR #1947, opened 2022-08-12
docs: updated queue docs to specify the cdk queue vs sst queue

justindra:
There are two imports of `Queue` that happens and the docs made it seem like it should both be coming from the main AWS CDK instead of separating the imports.


fwang:
Thanks @justindra!

PR #1960, opened 2022-08-18
Interface elements should end with ; not ,

revmischa:
Doc: https://www.typescriptlang.org/docs/handbook/2/objects.html


fwang:
Ah good catch!

PR #1969, opened 2022-08-19
Resources/config: Set ConfigType interface properties readonly

kevin86wright:
The Config object performs the SDK calls to fetch Parameters and Secrets as a Lambda starts.

Setting the interface properties will discourage the changing of the Config.MY_PARAM values during runtime of the lambda.

If someone were to inadvertently perform:
`Config.MY_PARAM = "someNewValue"`
The Config values would be incorrect on subsequent invocations of the Lambda's lifecycle.

This change will prevent changing these values at compile time of the project.


thdxr:

PR #1975, opened 2022-08-21
fix: Remove single quotes to avoid JSON parse error

vladimirblaz:
When a new minimal SST project is created, and run `npm start`, it throws the following error:

```
'{}'
^

SyntaxError: Unexpected token ' in JSON at position 0
```

Removed single quotes from caller to avoid JSON parse error.

### DIFF
--- a/packages/core/src/bootstrap/index.ts
+++ b/packages/core/src/bootstrap/index.ts
@@ -96,7 +96,7 @@ async function deployStack(config: any, cliInfo: any, tags: Record<string, strin
       "node",
       "bin/index.mjs",
       region,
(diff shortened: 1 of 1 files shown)


fwang:
Hey @vladimirblaz, thanks for the contribution!

Without single quote, parameters with spaces won't get pass correctly. I ended up adding a fix in v1.8.3 to encode the parameter in base64. You can see the change in #1977.

Let me know if v1.8.3 fixes it for you.

vladimirblaz:
@fwang, yes, a new SST project in v1.8.3 works as expected, and my existing project as well, great job!!

PR #1979, opened 2022-08-21
Function: merge config with default function props

fwang:


revmischa:
Thank you! 🙏🏻 
Trying this now

PR #1995, opened 2022-08-24
Docs: Parameter and Secret SSM name

ilhamadun:
The docs specified that the generated SSM names are prefixed with `/aws/`, while in fact the generated names are prefixed with `/sst/`.

https://github.com/serverless-stack/sst/blob/0bc3ef4cd10977c13ec3856742a75133d0e6722f/packages/core/src/function-config/index.ts#L142-L144
https://github.com/serverless-stack/sst/blob/0bc3ef4cd10977c13ec3856742a75133d0e6722f/packages/core/src/function-config/index.ts#L150-L152

Also fixed some typos.


fwang:
Ah nice catch! Thanks @ilhamadun, that's a lot of typos.

PR #2008, opened 2022-08-28
Clarify that additional pkg needs to be installed

Sheldonfrith:
As a beginner working through this page of the docs I spent a bit of time struggling to make the `@serverless-stack/node/config` import work, not realizing that I needed to manually install an additional npm package. 

Other users may assume, like me, that create-sst would install all needed packages automatically. I think it would be useful to add a bit of clarification about that point on this page to remove friction from the learning process. 

I've made changes I think would be best, but the exact wording and positioning are of course up for debate.

### DIFF
--- a/www/docs/environment-variables.md
+++ b/www/docs/environment-variables.md
@@ -68,7 +68,7 @@ In this chapter we'll look at how they work.
    };
    ```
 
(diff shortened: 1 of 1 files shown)


53jk1 (approved):

thdxr:

PR #2021, opened 2022-09-02
RemixSite: Fix malformed cookie header

### LINKED ISSUE #2020: Remix-site malformed cookie in request header when using regional server
When building request headers for the remix handler, cookies are appended twice. Once when mapping through `event.headers`, where Cookie is a prop, and once using `event.cookies`. This results in a comma as a separator in the middle of an otherwise semi-colon separated string.

Example result: `key1=value1; key2=value2, key1=value1; key2=value2`.

This causes parsing issues, eg: `key=value, key=value`, which gets parsed as `{ key: "value, key=value"}`.

Relevant lines:
https://github.com/serverless-stack/sst/blob/9da24d49cebf2a166e3b947188de10767e8f03d1/packages/resources/assets/RemixSite/server-lambda/regional-server.js#L106
https://github.com/serverless-stack/sst/blob/9da24d49cebf2a166e3b947188de10767e8f03d1/packages/resources/assets/RemixSite/server-lambda/regional-server.js#L111
https://github.com/serverless-stack/sst/blob/9da24d49cebf2a166e3b947188de10767e8f03d1/packages/resources/assets/RemixSite/server-lambda/regional-server.js#L116

lwsn:
Fixes #2020

### DIFF
--- a/packages/resources/assets/RemixSite/server-lambda/regional-server.js
+++ b/packages/resources/assets/RemixSite/server-lambda/regional-server.js
@@ -93,7 +93,7 @@ function createRemixRequest(event) {
 
   return new NodeRequest(url.href, {
     method: event.requestContext.http.method,
-    headers: createRemixHeaders(event.headers, event.cookies),
+    headers: createRemixHeaders(event.headers),
     body:
       event.body && event.isBase64Encoded
         ? isFormData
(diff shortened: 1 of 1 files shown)


fwang:
Thanks @lwsn!

PR #2047, opened 2022-09-08
Rewrite pothos extractor in babel

Mokshit06:
This PR ports the pothos schema extractor over to babel. This also fixes the issue with the acorn extractor where the references used wouldn't get extracted, but since babel tracks bindings this now works. I've also added a vitest test to ensure that the output is correct and it matches the one produced by the acorn extractor.


RobbeCl:
Im curious if this will get merged soon :-)

JakeHL:
Looking forward to this being merged. @Mokshit06 Could you suggest a workaround for the meantime?

thdxr:
I'm including this in sst 2.0

michelts:
I believe this ticket can be closed now, can't it?

PR #2049, opened 2022-09-09
Next NextjsSite - using standalone output

### LINKED ISSUE #1988: NextjsSite build fails when using Next 12.2 middleware
My issue: Nextjs removed the `target: serverless` option recently. If you try to use middleware (the latest 12.2 version not 12.1 or 12.0) then it throws a build-time error:
```
> Build error occurred
MiddlewareInServerlessTargetError: Next.js Middleware is not supported in the deprecated serverless target.
Please remove `target: "serverless" from your next.config.js to use Middleware.
```

The problem as I understand it is that NextjsSite is using @sls-next/lambda-at-edge which does not appear to be super actively maintained. It tries to force `target: serverless` by rewriting your nextjs config. Also it doesn't work with ESM configs. See my draft PR: https://github.com/serverless-nextjs/serverless-next.js/pull/2478

## Proposed solution

This guy has a solution with code for building nextjs with [target: standalone](https://nextjs.org/docs/advanced-features/output-file-tracing#automatically-copying-traced-files) and deploying it with a single lambda function and CDK: https://github.com/vercel/next.js/discussions/33275#discussioncomment-3343424

I suppose that would be better than using @slsnext/lambda-at-edge? (it does use `import lambdaAtEdgeCompat from '@sls-next/next-aws-cloudfront`). It looks like pretty a pretty minimal setup for a nextjs server request handler, though needs some of the S3/CF resources which I think are already set up in NextjsSite.The handler and static file deployments looks relatively straightforward.

CDK example: https://github.com/iiroj/iiro.fi/commit/bd43222032d0dbb765e1111825f64dbb5db851d9

revmischa:
Fixes #1988

**Update:** now trying to build this as a standalone construct over at https://github.com/jetbridge/cdk-nextjs

## Problem
`NextjsSite` relies on a not-very-maintained package for Serverless to build a NextJS app and deploy it using CDK. Instead of trying to [patch](https://github.com/serverless-nextjs/serverless-next.js/pull/2478) this library we can simplify the setup and not rely on this legacy dependency. Using the standalone output option for nextJS should enable new features such as middleware and .mjs config files (should help fix #1988).

## Solution
There is a new (in Next 12) output mode `standalone` which uses [output tracing](https://nextjs.org/docs/advanced-features/output-file-tracing) to generate a minimal server and static files. This output can be converted into a CloudFront distribution and a Lambda handler that translates between a APIGWProxyV2 req/res and Next req/res. The Lambda can be a HTTP origin using Lambda function URLs.

* We have a small NextJS server lambda function that initializes the standalone server and invokes it with the lambda function URL event. It is bundled by esbuild and uses external `next` and `sharp` modules that are included with the function. `sharp`
* CloudFront default behavior is an OriginGroup that looks for an asset in S3 first, otherwise falls back to Lambda. We need this to support `public` files (e.g. `robots.txt`). This could be an issue if you have a static file and replace it with a dynamic route - currently we don't prune old files when deploying so the old file would take precedence.
* Environment vars are replaced with placeholder values because the resolved values (e.g. `api.url`) are not resolved at NextJS build time. The placeholders are swapped out post-deployment in S3 for static files and using `process.env` for backend code.

## Status
I have managed to get the server bundling working even under the most finicky of circumstances (pnpm monorepo). I did my best to merge the RemixSite and original NextjsSite code but there is definitely some cleanup to be done and props and features missing. I could use some help getting this polished and with docs.

## Caveats
Nextjs requires the `sharp` native library, for simplicity i grabbed the layer zip (I noticed SST even links to this specific zip in [the layer docs](https://docs.sst.dev/advanced/lambda-layers#1-packaging-node_modules-into-a-layer)...) and use that. we can build the layer from scratch using npm or docker. there is some commented out code to that effect in the PR right now.

Edge functions should be possible to set up soon, the main blocker is resolving the CDK tokens in env vars on the server side. One of these issues needs to be fixed for that to work I think: https://github.com/vercel/next.js/issues/40827 https://github.com/aws/aws-cdk/issues/19257

## Performance

**On sst-prisma:**

`Duration: 616.43 ms	Billed Duration: 617 ms	Memory Size: 2048 MB	Max Memory Used: 131 MB	Init Duration: 481.08 ms`

**On my nextjs app using Material-UI**

`Duration: 957.56 ms	Billed Duration: 958 ms	Memory Size: 1024 MB	Max Memory Used: 127 MB	Init Duration: 530.86 ms`

next-server-mui

### Based on:
* https://github.com/iiroj/iiro.fi/commit/bd43222032d0dbb765e1111825f64dbb5db851d9
* https://github.com/sladg/nextjs-lambda
* https://github.com/serverless-nextjs/serverless-next.js/tree/master/packages/compat-layers/apigw-lambda-compat
* RemixSite construct
* NextjsSite construct

## Questions
* Do we need to manually handle CloudFront invalidation? It looks like `BucketDeployment` takes care of that for us
* How is the `public` dir supposed to be handled? (Right now using an OriginGroup to look in the S3 origin first and if 403/404 then try lambda origin)
* Is there anything we should be doing with the various manifests nextjs spits out? (e.g., not sure what the purpose of [this](https://github.com/serverless-stack/sst/blob/master/packages/resources/src/NextjsSite.ts#L1357) is)
  * Do we need to create static routes? Or anything else?
* Can we release this as an "experimental" construct?
* Do we need to handle ISR? 
* Do we need to create edge functions? - Need to fix the env var substitution in server code for this to work - the NextjsSite's lambda-code-updater.py does not support symlinks in zipfiles so it ruins node_modules when it runs
* How should images be handled?


revmischa:
Rebuilt nextjs site construct lives at https://github.com/jetbridge/cdk-nextjs now
If you have any questions come talk to us in #nextjs on discord

PR #2066, opened 2022-09-14
Expose edge lambdas for NextjsSite

revmischa:
I want to customize the caching behavior for certain paths in NextjsSite. I can do:

```ts
const origin = new S3Origin(frontendSite.cdk.bucket);
  frontendSite.cdk.distribution.addBehavior('/some/path/*', origin, {
    cachePolicy: profilesCachePolicy,
    edgeLambdas:  // need value here
  });
```

And I need to get the built edge lambdas to make this work I think.

Can be continued in #2049 as well for the new construct


thdxr:

PR #2078, opened 2022-09-16
Core: Fix lockfile not found in bundle.nodeModules

kuba-gaj:
Based on discussion: https://discord.com/channels/983865673656705025/983866416832864350/1019587704536715295

In monorepo setups, when using `bundle.nodeModules` the lockfile was searched for in the wrong place (`srcPath`).

Based on the discord discussion linked above:
1. Search upwards from `srcPath` until `lockfile` is found
2. Throw an error if no l`ockfile` found

Manually tested on examples provided in the discord thread

### DIFF
--- a/.changeset/ten-walls-work.md
+++ b/.changeset/ten-walls-work.md
@@ -1,0 +1,5 @@
+---
+"@serverless-stack/core": patch
+---
+
+Fix lockfile not found in some monorepos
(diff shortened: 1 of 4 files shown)


kuba-gaj:
Thank you, I removed the dependency with simple implementation of `find-up`

PR #2082, opened 2022-09-19
AppSync: Fix None DataSource code example

parodoTS:
Fix the AppSyncApiNoneDataSourceProp code example.

Not sure if the change applied in the correct file/way because of the:

_!! This file has been automatically generated, do not edit  !!_

comment at the beginning of the file.


fwang:
Thanks @parodoTS, nice catch!

PR #2084, opened 2022-09-21
Await runAfterBundling

revmischa:


fwang:
Thanks @revmischa, nice catch! Will cut a release tmr w/ #2066

PR #2101, opened 2022-09-28
Core: Enhanced java build support including natively compiled lambdas

mgranell:
* Uses gradlew to launch gradle if it exists in folder
* Outputs build log from gradle if either "DEBUG" or "SST_BUILD_OUTPUT" environment variables set
* Enable customising build command through new bundle option "buildCommand"
* Enable customising which folder .zip file is picked up through bundle option "buildOutputFolder"
* Enables native lambdas (i.e. compiled to executable) by setting bundle option "customRuntime" to true

Added two new example projects "minimal-java" and "minimal-java-native" to demonstrate the above.


fwang:
Released in v1.15.10

PR #2105, opened 2022-09-28
cli: Support http proxy via global-agent

mgranell:
Required to enable sst to be used behind a proxy. 

Tested with build and deploy. 

Works for live debugging (may depend on the proxy).


fwang:
Thanks @mgranell. Released in 1.15.9.

PR #2117, opened 2022-10-03
fix: RemixSite polyfilling with treeshaking enabled

ctrlplusb:
This PR enables treeshaking of ESBuild whilst also ensuring that the required Remix polyfills do not get removed during the treeshake. Results in smaller bundle sizes that still work when deployed.

Remix by default opts into treeshaking via the `"sideEffects": false` config in `package.json`. Users can opt out per their needs still.

Tested and verified against a deployed application.

### DIFF
--- a/packages/resources/assets/RemixSite/server-lambda/polyfill.js
+++ b/packages/resources/assets/RemixSite/server-lambda/polyfill.js
@@ -1,0 +1,24 @@
+import { installGlobals } from "@remix-run/node";
+
+// A Remix app by default has `"sideEffects": false` set in the package.json
+// which will trigger ESBuild's treeshaking capabilities. This results in the
+// polyfill being removed from the bundle.
+//
+// This behaviour can either be prevented by setting "sideEffects": true` within
+// the Remix application's package.json, or by providing the `--ignore-annotations`
+// argument to the ESBuild CLI.
+//
+// Whilst the above strategy works it will have a negative impact
+// on the overall bundle size. It would be better to allow users to opt into
+// the disabling of tree shaking themselves per their own needs.
+//
+// We therefore do not disable treeshaking in ESBuild and instead us a hack of
+// performing a "+" operation against the polyfill installation code below. By
+// performing an operation ESBuild will mark the code as impure and will not
+// consider this line for treeshaking.
+//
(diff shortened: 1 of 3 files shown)


ctrlplusb:
Woops, test broke as was unable to reference the `@remix-run/node` package. I copy it into the Remix app now so that package resolution works as expected.

packages/resources/src/RemixSite.ts
```diff
@@ -789,9 +798,13 @@ export class RemixSite extends Construct implements SSTConstruct {
...
-      ignoreAnnotations: true,
-    })
+      // We need to ensure that the polyfills are injected above other code that
+      // will depend on them. Importing them within the top of the lambda code
+      // doesn't appear to guarantee this, we therefore leverage ESBUild's
+      // `inject` option to ensure that the polyfills are injected at the top of
+      // the bundle.
+      inject: [polyfillDest],
```
fwang, line 806: Does passing in `polyfillSource` not work?

fwang (commented):
Thanks @ctrlplusb! I just had 2 minor things that I wanted to clarify.

packages/resources/src/RemixSite.ts
ctrlplusb: Unfortunately not. That was my original implementation, but then ESBuild tried to do package resolution to the referenced remix package which failed your unit tests. If we added the remix package as a dependency to your resources package, or higher then it could work. As a dev dependency maybe?

packages/resources/assets/RemixSite/server-lambda/polyfill.js
```diff
@@ -0,0 +1,24 @@
...
+// We therefore do not disable treeshaking in ESBuild and instead us a hack of
+// performing a "+" operation against the polyfill installation code below. By
+// performing an operation ESBuild will mark the code as impure and will not
+// consider this line for treeshaking.
+//
+// Yes, it's a bit dirty for now, but its only a light amount of dirt that gives
+// the user great benefit in having the required polyfills ready for their
+// application, whilst also maintaining optimal bundle sizes.
+
+installGlobals() + "⛔️🌲";
```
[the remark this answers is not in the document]
ctrlplusb: @fwang happy to push this either way. Your call. :)

PR #2118, opened 2022-10-03
changed https to http in localhost

CRRogo:
https in localhost causes redirect errors on logout.


fwang:
Thanks @CRRogo! Nice catch.

Issue #2158, opened 2022-10-23
 Improve environment variables doc quickstart example


tlcheah2:
## Problem

![Screenshot 2022-10-23 at 2 12 32 PM](https://user-images.githubusercontent.com/7456033/197377054-85084023-8e9c-49b1-bd52-c435f01725f3.png)

The quickstart guide step 2 will mislead beginners. Adding the `api` code to finish the whole example would provide more clarify and better understand. For e.g,

```
  const myFunction = new Function(stack, "MyFunction", {
    handler: "lambda.handler",
    config: [STRIPE_KEY, GITHUB_TOKEN],
  });

  const api = new Api(stack, "api", {
   routes: {
     "GET /notes": myFunction,
   },
  });
```

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #2159, opened 2022-10-23
doc: improve environment variables doc quickstart example

tlcheah2:
## Description

- Improve environment variable quickstart example

@jayair @fwang will the changes make it better to understand?

#2158

### DIFF
--- a/www/docs/environment-variables.md
+++ b/www/docs/environment-variables.md
@@ -78,10 +78,16 @@ Secrets and Parameters are stored in AWS SSM with the _Standard Parameter type_ 
    export default function MyStack({ stack }: StackContext) {
      const { USER_UPDATED_TOPIC } = use(TopicsStack);
 
-     new Function(stack, "MyFunction", {
+     const myFunction = new Function(stack, "MyFunction", {
        handler: "lambda.handler",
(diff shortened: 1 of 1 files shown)


fwang:
Hey @tlcheah2 can you share what was confusing?

tlcheah2:
Hi @fwang 

Sorry for the delay. It's confusing from the beginner's point of view.

```
new Function(stack, "MyFunction", {
       handler: "lambda.handler",
       config: [USER_UPDATED_TOPIC],
     });
```

The original doc only show that I need to pass the parameters via `config`. This part is great.

However, in order to have a complete working example, the function has to be assigned to `routes` like the example below.

```
const myFunction = new Function(stack, "MyFunction", {
       handler: "lambda.handler",
       config: [STRIPE_KEY, GITHUB_TOKEN],
     });

     const api = new Api(stack, "api", {
      routes: {
        "GET /notes": myFunction,
      },
   }):
```

So I find it adding the API part seems better to me. I also realize the environment variables docs had been removed. So probably this PR is not needed anymore?

53jk1 (approved):

thdxr:

PR #2160, opened 2022-10-25
docs: added default value information for events on buckets

ghost:
- Added the missing default value information on the optional 'events' field in the v1 constructs documentation.

@jayair @fwang


jayair:
This is editing the generated doc @fwang can we move it to the source file?

Sagar-Chudamani:
@jayair @fwang I have reverted the edits on the generated file and updated the source file.

53jk1 (approved):

PR #2217, opened 2022-11-23
examples/kinesisstream: update incorrect lambda paths

magnuswahlstrand:
I tried this example (SST version: 1.18.3), and got the following error:
```bash
==========================
 Starting Live Lambda Dev
==========================


Cannot find a handler file for "consumer1.handler"
```
I think my change fixes this. The lambda in the API GW has the "functions/" path prefix, the KinesisStream does not.

https://github.com/serverless-stack/sst/blob/master/examples/kinesisstream/stacks/MyStack.ts#L20


magnuswahlstrand:
Not sure how to add, tags. Sorry! 
I think it should be "bug".

53jk1 (approved):

PR #2219, opened 2022-11-24
DevX improvement for SST + prisma

ackinc:
While doing live-lambda development on ubuntu, I found that running `yarn start` after previously running `yarn deploy` resulted in a "query engine not found" error from Prisma. This change fixes that.

errorScreenshot

### DIFF
--- a/examples/prisma/stacks/MyStack.ts
+++ b/examples/prisma/stacks/MyStack.ts
@@ -4,13 +4,13 @@ import * as lambda from "aws-cdk-lib/aws-lambda";
 import { Api, StackContext } from "@serverless-stack/resources";
 
 export function MyStack({ stack, app }: StackContext) {
+  const layerPath = ".sst/layers/prisma";
+  // Clear out the layer path
+  fs.removeSync(layerPath, { force: true, recursive: true });
+
   if (!app.local) {
     // Create a layer for production
     // This saves shipping Prisma binaries once per function
-    const layerPath = ".sst/layers/prisma";
-
-    // Clear out the layer path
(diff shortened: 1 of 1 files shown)


ackinc:
The error was occurring because:

1. Running `yarn deploy` creates a `.sst/layers` folder and copies over the `.prisma/client` folder into it

2. On subsequently starting a live-lambda development session with `yarn start`, the @prisma/client package fails to find a schema.prisma in `__dirname` (which will be `.sst/artifacts/.../functions/services`)

3. @prisma/client then commences a search for `.prisma/client` in the entire current working directory, finding it inside `.sst/layers` instead of inside `node_modules`

prismaClientCode


4. It then looks for the platform-specific query engine binary (debian for live-lambda development) inside the found directory, failing to find it because only the rhel-binary was copied over due to

```
fs.copySync(file, path.join(layerPath, "nodejs", file), {
  // Do not include binary files that aren't for AWS to save space
  filter: (src) => !src.endsWith("so.node") || src.includes("rhel"),
});
```

5. The function fails with a "query engine not found" error

---

If (in step 3 above) @prisma/client had found the `.prisma/client` folder inside `node_modules` instead of inside `.sst/layers`, this error would be avoided, since the `node_modules/.prisma/client` folder contains both the platform-appropriate-query-engine-binary and the rhel-query-engine-binary.

By making the removal of `.sst/layers/prisma` unconditional, we ensure that @prisma/client finds `node_modules/.prisma/client` during live-lambda development rather than `.sst/layers/prisma/...`, and the "query engine not found error" is avoided.

53jk1 (approved):

thdxr:

PR #2279, opened 2023-01-01
Api: Adding `usePathParams`/`usePathParam` helpers

joshstrange:
There was an existing `usePath` but it returned an array of path segments (split on `/`). If you define your route like this:

```typescript
{
  'POST /user/{id}': 'function/user/update.handler',
}

```

Then `id` will be present in `event.pathParameters`. In this case you could also get the id with:

```
const path = usePath();
const id = path[1];
```

But it's much nicer to use the 'named' path param by getting it from `event.pathParameters`. Especially if you are trying to re-use code where the `id` might not always be in the same position in the path.

### DIFF
--- a/packages/node/src/api/index.ts
+++ b/packages/node/src/api/index.ts
@@ -104,3 +104,13 @@ export function useQueryParams() {
 export function useQueryParam(name: string) {
   return useQueryParams()[name];
 }
+
+export function usePathParams() {
+  const evt = useEvent("api");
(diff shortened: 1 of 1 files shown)


thdxr:
Good addition! Right now we're focused on our sst2 branch so if you make the PR there we can add it in

joshstrange:
Closing this because #2299 replaces it (for sst2)

PR #2298, opened 2023-01-05
console: Enable CMD/Ctrl + Enter to submit form

andynaguyen:
This enables CMD + Enter to submit the form on the API or Function pages in the sst console (Ctrl + Enter for Windows).

I've tested this locally and confirmed it passes in the correct query params, headers, and body for API.

Issue: https://github.com/serverless-stack/sst/issues/2276


thdxr:
awesome thank you, can't release this until we merge sst2 but should be soon

PR #2340, opened 2023-01-17
feat: rewrite nextjs static assets to replace env vars

0xFlicker:
Sorry.... I tried a couple of times to submit these files w/o prettifier but my VSCode was insistent. At some point it's better to just go with what you have...

Adds options to the custom resource deployer to replace env vars in built nextjs files with resolved values.

When nextJS builds, we don't have everything so env vars end up looking like:
```
  // ie. environment => { API_URL: api.url }
  //     environment => API_URL="{{ API_URL }}"
```
ref: https://github.com/serverless-stack/sst/blob/1ed8fed87427774f67e8d1984f5e9caf0ad8bd1d/packages/sst/src/constructs/BaseSite.ts#L132-L147

When doing the deployment we can then resolve these values using the `s3-upload.py` and it's custom resource.  For example, an attempt to replace an env var with an API URL the custom resource can use a `ReplaceValue` of:
```
[{"files":"**/*.html","search":"{{ NEXT_PUBLIC_GRAPHQL_URL }}","replace":{"Fn::Join":["",[{"Fn::ImportValue":"my-app-Api:ExportsOutputFnGetAttapiApiC3C473D0ApiEndpointEFEAABD2"},"/graphql"]]}},{"files":"**/*.js","search":"{{ NEXT_PUBLIC_GRAPHQL_URL }}","replace":{"Fn::Join":["",[{"Fn::ImportValue":"my-app-Api:ExportsOutputFnGetAttapiApiC3C473D0ApiEndpointEFEAABD2"},"/graphql"]]}}]
```

This is achieved by adding the `ReplaceValue` option to the deployment custom resource in `SsrSite.ts` with a default value of `[]`.  `NextjsSite` then extends that default with a sane option.

### DIFF
--- a/packages/sst/src/constructs/SsrSite.ts
+++ b/packages/sst/src/constructs/SsrSite.ts
@@ -13,6 +13,7 @@ import {
   CfnOutput,
   RemovalPolicy,
   CustomResource,
+  Token,
 } from "aws-cdk-lib";
 import * as s3 from "aws-cdk-lib/aws-s3";
 import * as iam from "aws-cdk-lib/aws-iam";
@@ -36,6 +37,7 @@ import {
   getBuildCmdEnvironment,
   BaseSiteCdkDistributionProps,
   buildErrorResponsesForRedirectToIndex,
+  BaseSiteReplaceProps,
 } from "./BaseSite.js";
 import { Permissions, attachPermissionsToRole } from "./util/permission.js";
 import {
@@ -55,8 +57,9 @@ export type SsrBuildConfig = {
   siteStub: string;
 };
 
-export interface SsrDomainProps extends BaseSiteDomainProps { }
-export interface SsrCdkDistributionProps extends BaseSiteCdkDistributionProps { }
+export interface SsrDomainProps extends BaseSiteDomainProps {}
+export interface SsrSiteReplaceProps extends BaseSiteReplaceProps {}
+export interface SsrCdkDistributionProps extends BaseSiteCdkDistributionProps {}
 export interface SsrSiteProps {
   /**
    * The SSR function is deployed to Lambda in a single region. Alternatively, you can enable this option to deploy to Lambda@Edge.
@@ -140,6 +143,8 @@ export interface SsrSiteProps {
    */
   waitForInvalidation?: boolean;
 
+  replaceValues?: SsrSiteReplaceProps[];
+
   cdk?: {
     /**
      * Allows you to override default id for this construct.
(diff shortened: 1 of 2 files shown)


packages/sst/src/constructs/SsrSite.ts
```diff
@@ -140,6 +143,8 @@ export interface SsrSiteProps {
    */
   waitForInvalidation?: boolean;
 
+  replaceValues?: SsrSiteReplaceProps[];
```
0xFlicker, line 146: added new prop, to mirror `StaticSite` behavior

packages/sst/src/constructs/SsrSite.ts
```diff
@@ -655,6 +664,7 @@ export class SsrSite extends Construct implements SSTConstruct {
             ];
           }
         ),
+        ReplaceValues: this.getS3ContentReplaceValues(),
```
0xFlicker, line 667: Added custom resource option for `ReplaceValues`

packages/sst/src/constructs/SsrSite.ts
```diff
@@ -1067,6 +1077,11 @@ export class SsrSite extends Construct implements SSTConstruct {
     }
   }
 
+  protected getS3ContentReplaceValues(): SsrSiteReplaceProps[] {
+    const replaceValues: SsrSiteReplaceProps[] = this.props.replaceValues || [];
+    return replaceValues;
+  }
```
0xFlicker, line 1083: Default for base class is to do nothing. Recognizing now that I could probably just remove this and do any transforms I need in `NextjsSite` constructor

0xFlicker:
LMK if this is the general right direction, is useful, if you want me to clean anything up and/or add changeset.  Apologies for jumping the contributor guidelines and going right to step 3, but I was just trying to make it work.

FilipPyrek:
Also encountered this issue. Got referenced here from Discord. Quite a blocker for me.

Just to validate that this is the same issue:

When I developed it locally with `sst env "next dev"`, it worked.

When I deployed it via `sst deploy`, it doesn't work.

When I took a look into `.next` folder, I saw in the bundle
```js
userPoolId:"{{ NEXT_PUBLIC_USER_POOL_ID }}"
```

In the source code I have 
```typescript
userPoolId: process.env.NEXT_PUBLIC_USER_POOL_ID,
```

And I see the same thing as in the `.next` folder in the final files in the browser. There should be the value from the env variable instead - the actual User Pool ID.

The setup worked for me on SSTv1. This issue came with upgrade to SSTv2.

jayair:
@fwang were you working on fixing this?

PR #2341, opened 2023-01-17
chore: add yarn release

0xFlicker:
I needed to add this file from mainline in order to use the `sst2` branch


0xFlicker:
team uses pnpm for sst2 which is A-OK!

PR #2342, opened 2023-01-17
typo: incorrect bracketing

jagnani73:
the construct for bucket in exmaples has some incorrect bracketing


fwang:
Thanks @jagnani73, good catch!

PR #2348, opened 2023-01-17
feat: allow StaticSite to use existing CloudFront Distribution

### LINKED ISSUE #2311: Reuse existing Cloudfront distribution for StaticSite
We currently have a micro frontend setup that I'm looking to convert to use SST StaticSite construct.

In our current setup, the frontend sites are deployed to a single S3 bucket, a single Cloudfront distribution is used to serve the sites (Lambda@Edge is used to route traffic to different paths/folder of the S3 bucket).

If we were to use StaticSite for the sites, we would hit Cloudfront limit for our dev AWS account, as each site would have its own Cloudfront distribution.

I just wonder if there is any issues would prevent us from reusing an existing Cloudfront distribution across our StaticSite constructs?
I saw there is an option to reuse S3 bucket, but it seems to always create new Cloudfront distribution at the moment.

hoangnd25:
fixes #2311 

Allow StaticSite to use existing CloudFront distribution instead of create new distribution.
The construct should still create custom resource for CF invalidation.

### DIFF
--- a/packages/resources/src/StaticSite.ts
+++ b/packages/resources/src/StaticSite.ts
@@ -273,7 +273,7 @@ export interface StaticSiteProps {
      */
     bucket?: s3.BucketProps | s3.IBucket;
     /**
-     * Configure the internally created CDK `Distribution` instance.
+     * Configure the internally created CDK `Distribution` instance or provide an existing distribution
      *
      * @example
      * ```js
@@ -287,7 +287,7 @@ export interface StaticSiteProps {
      * });
      * ```
      */
-    distribution?: StaticSiteCdkDistributionProps;
+    distribution?: cloudfront.IDistribution | StaticSiteCdkDistributionProps;
   };
 }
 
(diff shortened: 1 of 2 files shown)


53jk1 (approved):

JDanielGalvan-Justo:
Hi there, is there a date to release this?

jayair:
Hmm I don't think we've reviewed this. @fwang can we take a look?

ktree-itsupport:
@fwang @hoangnd25 

Requesting you to please check and approve...this important feature for us...as we are deploying nextjs app to the existing website as part...thanks

fwang:
Just merged w/ master, looks good.

Quick question before merging in, if multiple sites are sharing the same distribution, are all sites deployed to the same S3 bucket? And are they sharing the same domain?

Can someone share a snippet of their setup? I'd like to document it in the docs.

Thanks /cc @hoangnd25 @JDanielGalvan-Justo @ktree-itsupport

fwang:
Released in v2.8.8

PR #2361, opened 2023-01-22
fix: typo in jsdoc examples

estyrke:
`bundle` has been renamed to `nodejs`. Not sure if there are more things renamed.


estyrke:
On second glance, it looks like this applies to JavaProps and CopyFiles as well...

fwang:
Thanks @estyrke. Merged the fixes.

^ top part of the Function doc is still outdated. We will be fixing it in the next day or two.

Issue #2362, opened 2023-01-22
docs: Code example link directing to 404 Not Found


jasonrowsell:
The following SST.dev pages have links to view code in GitHub for the "Rest API" example:
- https://sst.dev/examples/index.html
- https://sst.dev/examples/how-to-create-a-rest-api-with-serverless.html

However, in the repo, the example has been created under the name `rest-services` instead of `rest-api`. This leads to the user being redirected to GitHub's 404 Page not found.

Proposed solution:
- Recreate example as `rest-api`, replacing `rest-services`

jayair:
Thank you!

PR #2363, opened 2023-01-22
docs: Rename `rest-services` example to `rest-api`

jasonrowsell:
Addresses #2362


jasonrowsell:
@jayair @thdxr Would you guys be ok to give this a review when possible? Thanks

jayair:
Oh good catch. Appreciate it!

PR #2379, opened 2023-01-26
Fix broken AWS CDK links in AppSync docs

jeansibelius:
### Issue

AWS CDK alpha version links seem to have broken.

### Fix
Replace broken links with working AWS CDK links.


fwang:
Thanks @jeansibelius! Fixed in v2 doc.

PR #2387, opened 2023-01-28
Update resource-binding.md "Cost" documentation

SpencerDuball:
I believe an amendment to the "Cost" section of the resource-binding would be helpful. While AWS SSM is free to use, decrypting the secrets in AWS KMS is not free forever, the free tier has 20,000 API interactions per month and costs $0.03 per 10,000 subsequent API calls. I think this might be hit a lot sooner than many might think, I have been working on a Remix app and hit 85% of this limit with 3 secrets when developing this (lots of reloads with file saves etc). The Remix construct also does not have the api for `Config.bind` in order to cache subsequent requests with a warm lambda. This limit can probably be reduced but I believe other will incur costs in production that might otherwise not be anticipated.

### DIFF
--- a/www/docs/resource-binding.md
+++ b/www/docs/resource-binding.md
@@ -319,7 +319,7 @@ When binding resources that contain sensitive values, placeholders are stored in
 
 ## Cost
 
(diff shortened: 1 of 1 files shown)


jayair:
Thanks for this! Edited and added to the new docs.

Issue #2417, opened 2023-02-03
start broken when using multiple userPoolIds for authorizer


MoeQuadrat:
when using a cognito authorizer like this 

```ts
export const cognitoAuthorizer = (): ApiGatewayV1ApiAuthorizer => {
	const { cognitoUserPools } = useCognitoStack();
	return {
		type: 'user_pools',
		userPoolIds: [...cognitoUserPools],
	};
}
``` 
and adding multiple userPoolIds SST start/deploy breaks with following 
> Error: There is already a Construct with name 'cognitoAuthorizer-ImportedUserPool' in ApiGatewayV1Api [api]
message

fwang:
Released in v2.0.2

PR #2418, opened 2023-02-03
fix(#2417): changing construct name to be dynamic depending on userpool

MoeQuadrat:
should be the fix for #2417 as far as im aware


thdxr:
Hey we're currently not accepting PRs to master as we're gearing up to release SST2

we can merge this if you resubmit to the sst2 branch

PR #2426, opened 2023-02-06
fix: handle alternate domain names in SsrSite

mathisobadia:
When using the SsrSite construct, the customDomain.alternateNames has no effect on the cloudfront distribution created.
This PR fixes this so that it is in line with what the StaticSite construct does (the code was copy pasted from the staticSite construct)


mathisobadia:
I just noticed that we also need to update the generated hosted zone records to include alternate names! I'm going to update this PR to fix it

fwang:
Thanks @mathisobadia. Released in v2.0.2.

mathisobadia:
@fwang I think you might have forgot to commit something because I just tested and it does not work in 2.0.2 and looking at the code there is still no reference at all to alternateNames in the SsrSite.ts file

PR #2434, opened 2023-02-07
fix: handle alternate domain names for SsrSite construct

mathisobadia:
Currently the alternateNames prop has no effect when set for SsrSite construct (while it is present on many examples)
This PR is pretty basic and is just a copy paste of the code in StaticSite

While working on this I have also found that for both StaticSite and SsrSite the alternate names are only used in the cloudfront distribution settings and the additional route53 records are not created, I can either update this PR to include the code to generate the required route53 records or open a separate PR as this is kind of a different issue that affects both Static and SSR sites.

### DIFF
--- a/.changeset/grumpy-bobcats-boil.md
+++ b/.changeset/grumpy-bobcats-boil.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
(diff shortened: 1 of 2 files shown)


Runela31:
Thank you

PR #2475, opened 2023-02-13
Update configuring-sst.md

ashishpandey001:
fix incorrect config key from `region` to `profile`


fwang:
Thanks @ashishpandey001!

PR #2484, opened 2023-02-13
adding bind to SsrFunction construct so bind will work with nextjssit…

yorbincorbin:


53jk1 (approved):

fwang:
Thanks @yorbincorbin. Added bind to `NextjsSite` based on this PR and also applied to other SSR sites.

Released in v2.3.0.

PR #2495, opened 2023-02-14
fix: NextjsSite allow rsc headers for in-place routing

khuezy:
Issue: NextjsSite application router navigations were hard refreshing the page instead of routing in place.

Fix: router navigations now happen in-place without a full page reload.

NOTE: very small change but took a long time to figure out what was missing.

### DIFF
--- a/packages/sst/src/constructs/NextjsSite.ts
+++ b/packages/sst/src/constructs/NextjsSite.ts
@@ -282,6 +282,8 @@ export class NextjsSite extends SsrSite {
       headerBehavior: cloudfront.CacheHeaderBehavior.allowList(
         // required by image optimization request
         "accept",
+        // required by middleware for <Link /> routing in place
+        "rsc",
         // required by server request
         "x-op-middleware-request-headers",
         "x-op-middleware-response-headers",
--- a/packages/sst/src/constructs/NextjsSite.ts
+++ b/packages/sst/src/constructs/NextjsSite.ts
@@ -282,8 +282,8 @@ export class NextjsSite extends SsrSite {
       headerBehavior: cloudfront.CacheHeaderBehavior.allowList(
         // required by image optimization request
         "accept",
-        // required by middleware for <Link /> routing in place
-        "rsc",
+        // required by middleware for routing in place
+        "rsc", "next-router-prefetch", "next-router-state-tree",
         // required by server request
         "x-op-middleware-request-headers",
         "x-op-middleware-response-headers",
--- a/packages/sst/src/constructs/NextjsSite.ts
+++ b/packages/sst/src/constructs/NextjsSite.ts
@@ -282,7 +282,9 @@ export class NextjsSite extends SsrSite {
       headerBehavior: cloudfront.CacheHeaderBehavior.allowList(
         // required by image optimization request
         "accept",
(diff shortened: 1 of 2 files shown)


packages/sst/src/constructs/NextjsSite.ts
```diff
@@ -282,6 +282,8 @@ export class NextjsSite extends SsrSite {
       headerBehavior: cloudfront.CacheHeaderBehavior.allowList(
         // required by image optimization request
         "accept",
+        // required by middleware for <Link /> routing in place
+        "rsc",
```
edvinerikson, line 286: Isn't there more headers, like router state that might be needed as well? https://github.com/vercel/next.js/blob/canary/packages/next/src/client/components/app-router-headers.ts

khuezy: Yep there's a few more headers but I didn't see any difference with or without them so I only included what was necessary.

I may have missed something though and it probably wouldn't hurt to include the others. But one thing to keep on mind is that there's a max number of headers you can set.

khuezy: It might have something to do with prefetching links on hover. I'll test it out in a few hours.

edvinerikson: Looking at the code, _I think_ the router state is used to know how much of the RSC layout components that need to be sent to the client.

edvinerikson: So by not allowing it will lead to entire layout tree must be sent on every navigation.

khuezy: Thanks for looking into it! Updated.

khuezy: Hmm.. behavior wise, I don't think anything changes though. My `<Link />`s are getting prefetched immediately, while local development only loads them when I hover over the links.

edvinerikson: Yeah it's all pretty subtle. Would be nice to have docs for this from nextjs side.

edvinerikson (approved):
Looks good to me 👍🏼

packages/sst/src/constructs/NextjsSite.ts
khuezy: Good idea, I'll add links to the nextjs source in a bit

khuezy:
@thdxr @fwang Ready for merge.

sho-ai-main:
I think it would be nice to include a prop in the nextjs site construct that allows adding headers to the allow list. E.g if I want to include the authorization header I need to go throught the cdk prop and check the source code for the nextjs construct, copy paste the whole distribution cache policy from there and add my own headers. If there was an additionalAllowedHeader prop or something it could be merged automatically with whatever is already set up by the construct and I woudn't have to check if the source code for the nextjs construct was modified to add headers.

fwang:
@khuezy I want to document these headers in the OpenNext readme. So just to clarify, are they required by Next.js routing — https://nextjs.org/docs/routing/introduction?

@sho-ai-main yeah that makes sense. Are you looking to add the `Authorization` header? Are you using NextAuth?

fwang:
Released in v2.0.27

PR #2499, opened 2023-02-15
small fix for  create-sst --template=graphql/dynamo

aisflat439:
When I created sst today I got this:
```js
import { Api } from "@serverless-stack/node/api";
```

and it should be
```js
import { Api } from "sst/constructs";
```

I think this catches all the places the CLI will update it.


thdxr:
The examples folder is actually autogenerated, the place they're generated from are `packages/create-sst/bin/presets`

aisflat439:
I see thanks @thdxr. I updated the PR so it only makes teh change to presets

Issue #2124, opened 2022-10-05
Function: support .NET 6.0 minimal APIs


fwang:
Reference:
- https://aws.amazon.com/blogs/compute/introducing-the-net-6-runtime-for-aws-lambda/

---
Requests
- vegardb https://discordapp.com/channels/983865673656705025/1026826873805746236

dex2dot0:
@fwang I have been looking at adding support and moving towards submitting a PR. There is currently a bootstrap project that is used https://github.com/serverless-stack/sst/tree/master/packages/sst/support/dotnet6-bootstrap. I can bypass that locally and get everything to work as expected but didn't know if the bootstrap was important to something SST was doing?

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #2514, opened 2023-02-18
Docs: Add Dotnet Core Minimal API Example

dex2dot0:
ASP.NET Core Minimal APIs work a little differently than the standard C# API implementation so this adds an example and an update to the documentation on how to make this work with SST. Addresses #2124.


thdxr:
Appreciate this example! Our examples are generated from the `packages/create-sst/bin/presets` folder so we don't edit them directly

PR #2538, opened 2023-02-22
Fix python lambda invocations

### LINKED ISSUE #2530: Python example is broken
Hey,

Trying to deploy a python lambda to my project. Seems to be dead on arrival. Decided to try running the python example and have run into similar issues.

The example is kinda broken, but the default handler still deploys. I'm open to fixing the example if I can test it properly.

```
npx create-sst@latest --template=examples/rest-api-python

cd my-app --whatever you named it
pnpm install
pnpm dev
```

Then try to run the lambda in any way locally.
- from the sst console
- from going to the api gateway url (should just be a GET)

The error I get is:

```
|  Invoked functions/lambda.handler
|  +84ms   Traceback (most recent call last):
|  +84ms     File "/Users/rjerue/dev/sst-python-ex/node_modules/sst/support/python-runtime/runtime.py", line 73, in <module>
|  +85ms   context = Context(
|  +85ms                 ^^^^^^^^
|  +85ms     File "/Users/rjerue/dev/sst-python-ex/node_modules/sst/support/python-runtime/runtime.py", line 30, in __init__
|  +85ms       self.identity = Identity(**json.loads(identity))
|  +85ms                                  ^^^^^^^^^^^^^^^^^^^^
|  +85ms     File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/json/__init__.py", line 339, in loads
|  +85ms       raise TypeError(f'the JSON object must be str, bytes or bytearray, '
|  +85ms   TypeError: the JSON object must be str, bytes or bytearray, not NoneType
```

Expected result: should say "Hello world"

rjerue:
Out of the box, python API invocations do not work. There's a crash because "Identity" isn't defined. The example also has things that do not work.

I could not find a clean way to get dependencies to work besides using lambda layers. NumPy itself is also sorta a pain since it out of the box doesn't work great with al2, but I think the dependency issues should be solved in a different PR.

Fixes #2530


thdxr:
Can you resubmit this PR without the changes to the example folder? those are auto generated so you shouldn't edit them directly

thdxr:
Think I fixed this in 2c332911ffa2ee6d6d459ab2464dc21cd04c9405

rjerue:
Thanks for fixing the issue and taking a look!

PR #2541, opened 2023-02-22
fix node import

iloewensen:


thdxr:
Our examples are autogenerated from packages/create-sst/bin/presets so they should not be edited directly

PR #2554, opened 2023-02-24
Removed `t.field.resolve` functions and unused functions, variables, etc when generating Pothos schema

### LINKED ISSUE #2553: Extracting Pothos schema results in not defined errors
I am creating a GraphQL API using the recommended setup by SST. When `sst dev` tries to write the schema to `schema.graphql` it gets the following error:

```
Pothos: Failed to extract schema:
Tracer is not defined
```

I am using `@aws-lambda-powertools/tracer` which has the export `Tracer`. The import for it is being removed from the transpiled file.

What seems to be happening is `resolve` functions inside of `t.field` are not being removed. Even after those resolve functions are removed, it will still fail if you have global variables that the resolve functions use.

lukeshay:
Resolves #2553

I am not well versed in babel so this is what I came up with to fix my problem. I had a variable defined in the root of a file, it was then being bundled but the import was being removed so it would error when importing it. Running esbuild + babel a second time, removes all unused variables from out.mjs. If there is a better way to do this, LMK!

### DIFF
--- a/packages/sst/src/pothos.ts
+++ b/packages/sst/src/pothos.ts
@@ -99,7 +99,7 @@ export async function extractSchema(opts: { schema: string }) {
                   !schemaBuilder ||
                   (callPath.node.callee.object.name !==
                     (schemaBuilder.node.id as any).name &&
-                    callPath.node.callee.property.name !== "implement")
+                    !["field", "implement"].includes(callPath.node.callee.property.name))
                 ) {
                   return;
                 }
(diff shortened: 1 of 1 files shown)


53jk1 (approved):

lukeshay:
Is there anything I need to do to get this merged?

thdxr:
I started taking a look at this and I need to think about this more - I'm not sure if double building is the right solution. Need to understand the scenario a bit better - a minimal reproduction would be helpful

lukeshay:
> I started taking a look at this and I need to think about this more - I'm not sure if double building is the right solution. Need to understand the scenario a bit better - a minimal reproduction would be helpful

I will make an example.

estyrke:
@lukeshay Does https://github.com/serverless-stack/sst/pull/2831 solve this for you? (add `"internalPackages": ["your-removed-import"]` to the Pothos route)

PR #2579, opened 2023-03-02
StaticSite: Allow supplying functionAssocations

### LINKED ISSUE #2578: StaticSite: Allow overwriting viewer request function
PR #2564 introduced a viewer request function and there doesn't seem to be a way to remove it.  We have setups that use an Edge function on viewer request and CloudFront does not allow both a CloudFront function and an Edge function at the same time on viewer request or viewer response.

brad-hader-FHR:
Fixes #2578


brad-hader-FHR:
Not needed as setting to `undefined` works.

PR #2580, opened 2023-03-03
Update what-is-sst.md

wmbond:
Old:  "that can be configured it through its props." 
New: "that can be configured through its props."


jayair:
Thanks!

PR #2586, opened 2023-03-04
Remove unusable cognitoIdentityId from log format

### LINKED ISSUE #2587: Deployment fails with AWS::ApiGatewayV2::Stage CREATE_FAILED The following context variables are not supported: [$context.identity.cognitoIdentityId]
Deploying any SST (create, not update) in us-west-2 results in:

AWS::ApiGatewayV2::Stage CREATE_FAILED The following context variables are not supported: [$context.identity.cognitoIdentityId]

Presumably, the change @ AWS that results in this failure will come to other regions shortly.

Root cause:

It is not valid in HTTP APIs to use $context.identity.cognitoIdentityId in a log statement, as HTTP APIs do not support direct integration with Cognito. The AWS create-stage API was recently updated such that attempting to use $context.identity.cognitoIdentityId in a log format results in an error when creating a stage. It was never possible for $context.identity.cognitoIdentityId to work, and so removing it should be harmless.

Note the AWS docs on this are wrong. [This page](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-logging-variables.html) claims it's valid to use $context.identity.cognitoIdentityId, but [this page](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html) makes it clear that HTTP APIs do not support direct Cognito integration.

Furthermore, the AWS Console makes it clear you cannot add a Cognito authorizer to an HTTP API (whereas you can to a REST API).

ttarhan:
It is not valid in HTTP APIs to use $context.identity.cognitoIdentityId in a log statement, as HTTP APIs do not support direct integration with Cognito. The AWS create-stage API was recently updated such that attempting to use $context.identity.cognitoIdentityId in a log format results in an error when creating a stage (currently only in us-west-2). It was never possible for $context.identity.cognitoIdentityId to work, and so removing it should be harmless.

Note the AWS docs on this are wrong. [This page](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-logging-variables.html) claims it's valid to use $context.identity.cognitoIdentityId, but [this page](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html) makes it clear that HTTP APIs do not support direct Cognito integration.

Furthermore, the AWS Console makes it clear you cannot add a Cognito authorizer to an HTTP API (whereas you can to a REST API).

Resolves #2587

### DIFF
--- a/.changeset/modern-cycles-pay.md
+++ b/.changeset/modern-cycles-pay.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Remove unusable cognitoIdentityId from log format
--- a/packages/sst/src/constructs/util/apiGatewayV2AccessLog.ts
+++ b/packages/sst/src/constructs/util/apiGatewayV2AccessLog.ts
@@ -26,8 +26,7 @@ const defaultHttpFields = [
   `"integrationServiceStatus":"$context.integration.integrationStatus"`,
   // caller info
   `"ip":"$context.identity.sourceIp"`,
-  `"userAgent":"$context.identity.userAgent"`,
-  `"cognitoIdentityId":"$context.identity.cognitoIdentityId"`,
+  `"userAgent":"$context.identity.userAgent"`
 ];
 
 const defaultWebSocketFields = [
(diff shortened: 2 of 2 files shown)


ttarhan:
@AbdullhRtima yes, it is

fwang:
Thanks @ttarhan! The field works for `WebSocketApi`, I only removed it for `Api` — #2602.

Released in v2.1.9.

PR #2604, opened 2023-03-06
Fix `sst dev` Node runtime not having all lambda `context` values

DCzajkowski:
As discussed in [Discord](https://discord.com/channels/983865673656705025/983866416832864350/1081224601419382814), when Live Lambda Development is active not all props of the Lambda context are available.

I've made the Node runtime contain all props. When working on it, I saw Python is missing two props so I've added them as well. 

**Note:**
`clientContext` and `identity` are `undefined` in the real AWS lambda implementation, whereas they are `{}` in Live Lambda Development. I didn't want to change the current behavior as that could be a breaking change (?), but we should address that as well.


DCzajkowski:
BTW Node lambda not having some of those props is a regression in SST v2.

thdxr:
thanks for this! think it's correct that `clientContext` and `identity` should be undefined - we did some some issues in the python runtime that tried to access it but feel free to make that change

DCzajkowski:
@thdxr I've made `clientContext` and `identity` behave the same in Node and Python runtimes as on AWS, ie:

* In Node in both cases, when they are empty they are set to `undefined`.
* In Python, `identity` is set to an object with both keys set to `None`, but `client_context` itself becomes `None`.

DCzajkowski:
@thdxr Resolved the conflicts :)

PR #2633, opened 2023-03-09
Fix authenticator handler path.

nsaunders:


thdxr:
Hey nick good to see you here!

ty for the fix

PR #2641, opened 2023-03-09
Remove mention of Astro in StaticSite construct docs

pawelblaszczyk5:
Hello 👋🏼 
While reading docs I spotted that Astro is mentioned instead of static site in the `StaticSite` construct description. This PR fixes it


jayair:
Good catch!

PR #2646, opened 2023-03-10
Fixed local lambda runtime having access to local env variables

DCzajkowski:
Removed `process.env` being merged to local lambda.

### Rationale

When running lambda in `dev` mode, we should try to mimick the real AWS Lambda environment. By merging in local env variables, we give access to local variables that won't be available when deploying final code to Lambda.

### The problem

The problem I encountered appeared after upgrading to v2.  I noticed, that I didn't have to give my lambda any permissions and it would still execute successfully. What I have found out, was that yes, the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` were correctly set.

But, that is when I noticed, that my lambda also had access to `AWS_PROFILE`, which was linking to my `~/.aws/credentials` file. 

### Other possible solutions

- Overwrite `AWS_PROFILE` setting it to the one coming from lambda,
- State in the docs that local env variables leak to the code when using `dev`.

### DIFF
--- a/packages/sst/src/runtime/handlers/node.ts
+++ b/packages/sst/src/runtime/handlers/node.ts
@@ -38,7 +38,6 @@ export const useNodeHandler = Context.memo(async () => {
           ),
           {
             env: {
-              ...process.env,
               ...input.environment,
               IS_LOCAL: "true",
             },
--- a/.changeset/real-pets-mix.md
+++ b/.changeset/real-pets-mix.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Fixed local lambda runtime having access to local env variables


kculmback-eig:
We are experiencing a similar issue to this. We use [granted cli](https://docs.commonfate.io/granted/introduction) to manage AWS sessions. The CLI sets the `AWS_PROFILE` env variable in addition to `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`. Since the profile variable is set, SST ignores the other env variables and looks for that profile in the credentials file.

As a workaround we can do `AWS_PROFILE="" npm run sst:dev`.

packages/sst/src/runtime/handlers/node.ts
```diff
@@ -38,7 +38,6 @@ export const useNodeHandler = Context.memo(async () => {
           ),
           {
             env: {
-              ...process.env,
```
fwang, line 41: @thdxr will this affect the spawned process?

thdxr: I _think_ this is the correct thing to. We shouldn't let your local env variable pollute right?

PR #2651, opened 2023-03-11
Fix example

Rohithgilla12:


jayair:
@Rohithgilla12 Oh this file is generated. Can you change this and submit a PR instead: https://github.com/serverless-stack/sst/blob/master/packages/create-sst/bin/presets/examples/layer-chrome-aws-lambda/templates/stacks/ExampleStack.ts#L25

PR #2673, opened 2023-03-16
Lambda runtime has no nodejs.18x in China partition

elonniu:
Fixed Lambda runtime in China


fwang:
Thanks @elonniu

fwang:
Released in v2.2.0

PR #2679, opened 2023-03-17
fix: localstorage.removeItem rather than clear

aisflat439:
While working through auth I noticed that this was clearing with an argument. What it should actually do is remove the item that's set previously. 

You can see the docs here:
https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#examples

There's [another PR here](https://github.com/AnomalyInnovations/serverless-stack-com/pull/703) for `serverless-stack-com`


jayair:
Thanks!

PR #2688, opened 2023-03-19
Updated source map docs with details on how to upload to Sentry

ffxsam:
This still needs editing from someone on the team.

But also.. I just bothered to look at Sentry docs again, and check this out:

https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/esbuild/

The Sentry team really stays on top of things! That didn't exist last time I checked. I wonder if this would work with SST, though there's still the question of removing `.map` files from the SST dist folder so that they don't get bundled into the Lambda funcs.

### DIFF
--- a/www/docs/advanced/source-maps.md
+++ b/www/docs/advanced/source-maps.md
@@ -35,3 +35,77 @@ export default function main(app) {
   // Add stacks
 }
 ```
+
+## Uploading source maps to Sentry
+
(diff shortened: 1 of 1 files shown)


ffxsam:
Whoa. Gotta put some sort of check in that hook! That runs in dev mode too, and will cause some big problems.

jayair:
Thanks for putting this together @ffxsam!

PR #2696, opened 2023-03-21
Update advanced confiugraiton

dev-cyprium:
Update the configuration to make it work as an example. Solves https://github.com/serverless-stack/sst/issues/2695


jayair:
Thanks!

PR #2700, opened 2023-03-22
sst/future-auth: added support for domain specific cookie for the token

justindra:
So the issue I'm having is that auth will be hosted on a different domain to the api and the frontend. E.g.
* Auth - `auth.example.com`
* API -  `api.example.com`
* Astro app - `example.com`

Now when we sign-out, whether that's through the api or the astro app, generally we are just removing the cookie so that any requests that comes after is no longer authenticated.

But you can't remove a cookie that was set for a specific domain, which it is currently there. So adding a `cookieDomain` when setting the cookie will make it available to all subdomains of that specified domain.

I don't really like that you have to set it for each Adapter though, so just wondering if there might be a better way of doing this instead of duplicating the preferences. Maybe adding an extra props that gets passed into the adapter with the event? Thoughts?

Other issues I see with this:
* Dev env might not be deployed at a specific domain, so won't work well with this

### DIFF
--- a/.changeset/silly-foxes-swim.md
+++ b/.changeset/silly-foxes-swim.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Setup support for root-domain specific cookie for sst_auth_token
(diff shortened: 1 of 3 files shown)


thdxr:

PR #2710, opened 2023-03-27
sst bind: fix command not exiting

### LINKED ISSUE #2709: Running tests hanging forever
Testing with latest version, from `create-sst` command, tests are running forever

Steps to reproduce.
```shell
npx create-sst@latest sst-test
cd sst-test
pnpm install
cd packages/core
pnpm test
```

image

Running vitest without sst bind exit as expected.
image

fwang:
Closes #2709

### DIFF
--- a/.changeset/smooth-humans-rule.md
+++ b/.changeset/smooth-humans-rule.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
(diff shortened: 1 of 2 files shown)


RobbeCl:
Don't you have to return the code in the process.exit() call?

Issue #2719, opened 2023-03-30
GO build handler not compatible with Windows


tmianhill:
When trying to build a stack containing a GO lambda (including the "other/go" template unmodified), on Windows, the following output is generated:

```
$ npm run build

> samplego@0.0.0 build
> sst build

Error: Failed to build
```

This is due to a combination of:

1. the path separator not being escaped correctly;
2. use of single quotes to surround the _-ldflags_ and _-o_ values.

In addition, the source of the error was obfuscated by the fact that the exception details are not output.

Our modification, which worked on Windows (but may not be compatible with other platforms?) was:

```diff
              if (input.mode === "deploy") {
                try {
                    const target = path.join(input.out, "bootstrap");
-                   await execAsync(`go build -ldflags '-s -w' -o '${target}' ./${src}`, {
+                   const srcPath = src.replaceAll("\\", "\\\\");
+                   await execAsync(`go build -ldflags "-s -w" -o "${target}" ./${srcPath}`, {
                        cwd: project,
                        env: {
                            ...process.env,
                            CGO_ENABLED: "0",
                            GOARCH: input.props.architecture === "arm_64" ? "arm64" : "amd64",
                            GOOS: "linux",
                        },
                    });
                }
-               catch {
-                    throw new VisibleError("Failed to build");
+               catch(ex) { // include exception detail in error message
+                   throw new VisibleError(`Failed to build: ${ex}`);
                }
            }
```

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #2721, opened 2023-03-30
fix go build for Win32 #2719

tmianhill:
fix "sst build", "sst deploy" and "sst dev" for GO lambdas in Win32:
- replace single-quotes with double-quotes in "go build" arguments
- escape file path separator \ in Win32


jayair:
Thanks for this. @thdxr

PR #2733, opened 2023-04-04
fix: remove ReactStaticSite leftover

pawelblaszczyk5:
Hello 👋🏼 ,
The url part of the StaticSite client docs is duplicated, it's a leftover from an old ReactStaticSite construct. It seems like it wasn't completely removed during the V2 overhaul. This PR removes the duplicated mention 😃 

Screenshot of how this looks currently:

image

https://docs.sst.dev/clients/site

### DIFF
--- a/www/docs/clients/site.md
+++ b/www/docs/clients/site.md
@@ -38,18 +38,6 @@ console.log(StaticSite.myWeb.url);
 
 ---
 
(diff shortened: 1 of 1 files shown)


fwang:
Thanks @pawelblaszczyk5!

PR #2740, opened 2023-04-05
Pin the open-next build command to a specific version

Brian-Azizi:
By default, the build command in the NextjsSite construct uses the latest version of open-next.

This is unsafe as the latest version of open-next may not be compatible with the current version of SST and can cause production failures even when pinning down the version of SST.

Users already have the option to manually override the build command via the `buildCommand` prop. It's still not great that the default behaviour of the construct is unsafe.

### DIFF
--- a/.changeset/stale-snakes-kiss.md
+++ b/.changeset/stale-snakes-kiss.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+NextjsSite: pin minor OpenNext version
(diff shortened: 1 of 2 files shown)


Brian-Azizi:
Why is open-next run via `npx` and not as a direct dev dependency

fwang:
Thanks @Brian-Azizi, I just made a minor tweak to pin to the minor version instead of the exact version.

> Why is open-next run via npx and not as a direct dev dependency
The thinking was that users shouldn't need to add `open-next` as a dev dependency in their Next.js app. `open-next` is meant to be used by framework authors. Also we want to keep SST's dependencies lean, and including `open-next` doesn't make sense for non `NextjsSite` users.

arpadgabor:
Glad this was fixed so quickly! Great job :)

PR #2758, opened 2023-04-07
Reduce IOT message chunk size to fix Live Lambda bug

### LINKED ISSUE #2723: `sst dev` fails to return payload over a certain size
We have a function in our SST app that crashes lambda local `sst dev` mode. It functions properly when it has been deployed to aws with `sst deploy`.
Minimal test reproduction: https://github.com/jensonb/sst-local-payload-bug
Discord thread: https://discord.com/channels/983865673656705025/1091177378513948713

The bug seems related to payload size. In my minimal reproduction, the failure seems to start once the response payload reaches ~80kB.

When the function blows up, `sst dev` becomes unresponsive. The request receives a `{"message":"Internal Server Error"}` error response after a long hang, and further API requests (including to other routes) receive the error about sst dev not running. Killing the process and restarting is the only solution I have found.

## Expected Behavior

The function should behave the same whether running locally after `sst dev` or on aws after `sst deploy`. `sst dev` should be able to respond with payloads that are not larger than the 6MB lambda payload limit.

## Current Behavior

When the function has been deployed with `sst dev`, the larger responses are handled properly. When the function is running in `sst dev` mode, the function will hang then return a `{"message":"Internal Server Error"}` response, after which the `sst dev` session becomes completely unresponsive to further requests. Small payloads behave consistently between the two modes, the issue is only once the payload surpasses around 80kB.

## Steps to Reproduce

1. Clone the minimal reproduction from https://github.com/jensonb/sst-local-payload-bug
2. (optional) Select nodejs version: `nvm use`
3. Install dependencies: `yarn`
4. Start local lambda mode: `yarn sst dev`
5. Visit `{API_GATEWAY_URL}/test/1` in your browser. Repeat for tests 2-5. Tests 4 and 5 should crash the `sst dev` session.
6. Deploy to aws: `yarn sst deploy`
7. Repeat step 5, all test cases work this time!
8. Cleanup: `yarn destroy`

## Context (Environment)

```
node: v18.12.1
SST: v2.3.5
CDK: v2.62.2
```
This bug makes `sst dev` keep crashing when using a certain part of our app, forcing us to resort to constantly re-deploying for lambda code changes.

edit: markdown formatting
edit: add link to thread in Discord

jensonb:
fixes #2723

### DIFF
--- a/.changeset/quiet-tools-buy.md
+++ b/.changeset/quiet-tools-buy.md
@@ -1,0 +1,6 @@
+---
+"sst": patch
+"@serverless-stack/docs": patch
+---
+
+Reduced IOT message chunk size to fix Live Lambda bug (#2723)
--- a/packages/sst/src/iot.ts
+++ b/packages/sst/src/iot.ts
@@ -72,7 +72,7 @@ export const useIOT = Context.memo(async () => {
         },
       ];
     }
-    const parts = json.match(/.{1,100000}/g);
+    const parts = json.match(/.{1,50000}/g);
     if (!parts) return [];
     Logger.debug("Encoded iot message into", parts?.length, "parts");
     return parts.map((part, index) => ({
--- a/www/docs/live-lambda-development.md
+++ b/www/docs/live-lambda-development.md
@@ -115,7 +115,7 @@ The AWS IoT events are published using the format `/sst/<app>/<stage>/<event>`. 
 
 Each Lambda function invocation is made up of three events; the original request fired by the remote Lambda function, the local CLI acknowledging the request, and the response fired by the local CLI.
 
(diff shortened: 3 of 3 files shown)


jensonb:
rebased on latest master.

jensonb:
@thdxr this PR is ready (including changeset and docs update), is there anything I can do to help facilitate landing this?

jayair:
lgtm!

jensonb:
Just for future reference, if we were ever interested in trying to increase the message size again to reduce the number of messages needed for large payloads, I think 80kb would do the trick as well. My payloads were around 90kb when they started failing prior to this fix.

PR #2763, opened 2023-04-08
fix: graphql examples front-end error

tyrauber:
The `urql` library api has changed causing the graphql examples to error when running the front-end.

Instead of calling `createClient`, the `urql` client must now be invoked with `new Client`.  Additionally, new client params are required. The new syntax is:

```
import { Provider as UrqlProvider, Client, cacheExchange, fetchExchange } from "urql";
…
const urql = new Client({
  url: import.meta.env.VITE_GRAPHQL_URL,
  exchanges: [cacheExchange, fetchExchange],
});
```

This PR updates the examples with these changes.

### DIFF
--- a/packages/create-sst/bin/presets/graphql/basic/templates/packages/web/src/main.tsx
+++ b/packages/create-sst/bin/presets/graphql/basic/templates/packages/web/src/main.tsx
@@ -1,13 +1,14 @@
 import React from "react";
 import ReactDOM from "react-dom/client";
-import { Provider as UrqlProvider, createClient } from "urql";
+import { Provider as UrqlProvider, Client, cacheExchange, fetchExchange } from "urql";
 import { BrowserRouter, Routes, Route, Navigate } from "react-router-dom";
 import Home from "./pages/Home";
 import Article from "./pages/Article";
 import "./globals.css";
 
-const urql = createClient({
+const urql = new Client({
   url: import.meta.env.VITE_GRAPHQL_URL,
+  exchanges: [cacheExchange, fetchExchange],
 });
 
 ReactDOM.createRoot(document.getElementById("root")!).render(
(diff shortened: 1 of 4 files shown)


jayair:
@tyrauber Thanks for this! The `examples/` files are generated from this one https://github.com/serverless-stack/sst/pull/2763/files#diff-a8049fd7e222e99eca959f0b0fc45fb59cc089c8b9176580c3d8007541fa861d. So we can remove those. Can you remove them from the PR and I'll merge them and generate the examples.

packages/create-sst/bin/presets/graphql/basic/templates/packages/web/src/main.tsx
```diff
@@ -1,13 +1,14 @@
...
+import { Provider as UrqlProvider, Client, cacheExchange, fetchExchange } from "urql";
 import { BrowserRouter, Routes, Route, Navigate } from "react-router-dom";
 import Home from "./pages/Home";
 import Article from "./pages/Article";
 import "./globals.css";
 
-const urql = createClient({
+const urql = new Client({
   url: import.meta.env.VITE_GRAPHQL_URL,
+  exchanges: [cacheExchange, fetchExchange],
```
jayair, line 11: Oh and also leave out this line, we add this as a part of the tutorial.

jayair:
Also which version of urql is causing this? I need to upgrade another repo to this.

tyrauber:
Hey @jayair, No problem. Happy to help wherever I can.  I think the version is correct, just the examples and tutorials contain the old syntax.  I am seeing 4.0.0 everywhere in the repo, and that is the latest release. I think just updating the examples and the tutorials to the new syntax should be enough.

If the examples are autogenerated, I think you can actually close this PR, no?

jayair:
Yeah I think what happened is that the examples got generated with the newer version but the tutorial was still using the older one. I'll update it.

Thanks for bringing it up!

PR #2768, opened 2023-04-09
Fixing the New Relic docs with code that will work in SST2

joshstrange:
Also adding some comments and more explanation.


jayair:
Thank you!

PR #2780, opened 2023-04-13
Disable Introspection in GraphQL Helix

Bhavikpatel576:
# Description
This pull request adds the ability to disable introspection in GraphQL Helix by using the disableIntrospection flag in the GraphQLHandlerConfig object. This feature allows for increased security in production environments by preventing unauthorized clients from querying the server's schema.

# Changes
The following changes have been made:

1. Import the necessary dependencies: specifiedRules and NoSchemaIntrospectionCustomRule from graphql
2. Add a new optional disableIntrospection boolean property to the GraphQLHandlerConfig interface
3. In the GraphQLHandler function, set the validationRules variable based on the value of disableIntrospection
4. Pass the validationRules variable to the processRequest function

# Reference
[Here](https://github.com/contra/graphql-helix/issues/50#issuecomment-922414985) is a conversation posted on graphql-helix

### DIFF
--- a/packages/sst/src/node/graphql/index.ts
+++ b/packages/sst/src/node/graphql/index.ts
@@ -55,7 +55,7 @@ export function GraphQLHandler<C>(config: GraphQLHandlerConfig<C>) {
     };
 
     const { operationName, query, variables } = getGraphQLParameters(request);
-    const validationRules = config.disableIntrospection ? [...specifiedRules, NoSchemaIntrospectionCustomRule] : [specifiedRules];
+    const validationRules = config.disableIntrospection ? [...specifiedRules, NoSchemaIntrospectionCustomRule] : [...specifiedRules];
 
     const result = await processRequest({
       operationName,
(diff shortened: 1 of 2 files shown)


packages/sst/src/node/graphql/index.ts
```diff
@@ -51,12 +55,14 @@ export function GraphQLHandler<C>(config: GraphQLHandlerConfig<C>) {
     };
 
     const { operationName, query, variables } = getGraphQLParameters(request);
+    const validationRules = config.disableIntrospection ? [...specifiedRules, NoSchemaIntrospectionCustomRule] : [specifiedRules];
```
jensonb, line 58: Should `specifiedRules` either be spread or just returned falsy expression?

jensonb:
Hey nice, I would be interested in this feature too! I left a comment on the code changes, I think there is a slight typo.

michaelgmcd:
Hey @Bhavikpatel576 @jensonb, I opened up https://github.com/serverless-stack/sst/pull/2784 which would change this entire handler. The team seemed to support the move to yoga in Discord, but I'd love some thoughts/eyes!

packages/sst/src/node/graphql/index.ts
Bhavikpatel576: Thanks for the catch @jensonb. Updated the PR.

PR #2784, opened 2023-04-15
GraphQLHandler: replace graphql-helix with graphql-yoga

michaelgmcd:
GraphQL Helix doesn't seem to be maintained anymore. graphql-yoga is a very powerful server fully capable of running in lambdas. I've been using an implementation almost identical to this for a couple weeks now and it's been working. GraphiQL works as well.

The handler implementation is heavily based on graphql-yoga's documentation:
https://the-guild.dev/graphql/yoga-server/docs/integrations/integration-with-aws-lambda

### DIFF
--- a/.changeset/dry-candles-run.md
+++ b/.changeset/dry-candles-run.md
@@ -1,0 +1,6 @@
+---
+"sst": patch
+"@serverless-stack/docs": minor
+---
+
+GraphQLHandler: replace graphql-helix with graphql-yoga
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 cdk.out
 .DS_Store
 dist
+.vscode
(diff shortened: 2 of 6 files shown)


michaelgmcd:
Note that this is a breaking change. I left it as WIP since I haven't tested it. What's the best way to test this? I can use this in my own project and confirm the functionality works, but open to other suggestions.

thdxr:
@michaelgmcd using it in your own project makes sense! let me know if you get to a place of confidence and I'll work out a release plan

Bhavikpatel576 (commented):
Might be helpful to update the docs as well [here ](https://github.com/serverless-stack/sst/blob/master/www/docs/learn/graphql-api.md) on what input shape is

michaelgmcd:
I've been testing this locally on my project without issues. I think it's good to review!

michaelgmcd:
Working on doc updates now

michaelgmcd:
Updated the docs and checked all the examples. May be worth adding some more complex examples, but each one only passes schema, so they don't need to be updated:
```javascript
export const handler = GraphQLHandler({
  schema,
});
```

PR #2785, opened 2023-04-15
fix: main.tsx to include exchange parameters

alvinlee123:
Error: Argument of type '{ url: string; }' is not assignable to parameter of type 'ClientOptions'.
Property 'exchanges' is missing in type '{ url: string; }' but required in type 'ClientOptions'.ts(2345)
urql-core-chunk.d.ts(88, 5): 'exchanges' is declared here.

This error is being caused because when creating the urql client, it expects an object of type ClientOptions. However, you are passing in an object with only a url key, which doesn't match that expected type.


alvinlee123:
Duplicate of https://github.com/serverless-stack/sst/pull/2763

PR #2812, opened 2023-04-26
Cognito: Expose cfnIdentityPoolRoleAttachment

jlanzarotti:
Exposing this allows users to work with the role configuration using override: 

```js
cognito.cdk.CfnIdentityPoolRoleAttachment.addOverride("Properties.RoleMappings", {...})
```
 
One basic use case is users from the pool to assume a role based on the group they belong. E.g `Type: "Token"`

### DIFF
--- a/.changeset/twenty-monkeys-drive.md
+++ b/.changeset/twenty-monkeys-drive.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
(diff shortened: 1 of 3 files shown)


ankit-gautam23 (approved):
LGTM

fwang:
Released in v2.8.8

PR #2792, opened 2023-04-19
fix: a few small fixes to future/auth

estyrke:
* Pass caught error on to LinkAdapter.config.onError
* Use correct (public) key for verifying token in LinkAdapter callback step
* Allow redirect_uri to contain search parameters
* Log errors caught in Session.verify()

I am happy to rewrite if any of these changes are unwanted. The critical fix is the second one, using the public key for verification.

PR #2814, opened 2023-04-27
Fix repository link in packages.json

Talljoe:
SST package no longer points to the official repo. Introduced in #2792


jayair:
Oh haha great catch.

PR #2817, opened 2023-04-27
feat: WebSocketApiHandler

justindra:
Similar to how we have the `ApiHandler`, adding a `WebSocketApiHandler` allows access to the `useSession` and all other contexts set from the event and requestContext.

Make sure to import the hooks from the right construct as required.
```ts
import {
  WebSocketApiHandler,
  useHeaders,
  useQueryParam,
} from 'sst/node/websocket-api'; // <-- instead of `sst/node/api`
import { useSession } from 'sst/node/future/auth';

export const handler = WebSocketApiHandler(async () => {
  const session = useSession();
  const headers = useHeaders();
  const params = useQueryParam('hello');
  console.log(session, headers, params);
  return {
    statusCode: 200,
  };
});

```

### DIFF
--- a/.changeset/cold-dingos-drive.md
+++ b/.changeset/cold-dingos-drive.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Updated the function binding to include httpsUrl as that is required to send messages using the ApiGateway client
--- a/.changeset/fifty-ants-whisper.md
+++ b/.changeset/fifty-ants-whisper.md
@@ -2,4 +2,4 @@
 "sst": patch
 ---
 
-Added WebsocketApiHandler so that we can use the auth sessions inside of a Websocket's connect and disconnect requests
+Added WebSocketApiHandler so that we can use the auth sessions inside of a WebSocket's connect and disconnect requests
--- a/.changeset/fifty-ants-whisper.md
+++ b/.changeset/fifty-ants-whisper.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Added WebsocketApiHandler so that we can use the auth sessions inside of a Websocket's connect and disconnect requests
--- a/.changeset/gentle-mails-compare.md
+++ b/.changeset/gentle-mails-compare.md
@@ -1,0 +1,6 @@
+---
+"sst": patch
+"@serverless-stack/docs": patch
+---
+
+Added a new WebSocketApiHandler to allow for websockets to use session hooks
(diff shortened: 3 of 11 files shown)


xinha-sh:
> Similar to how we have the `ApiHandler`, adding a `WebSocketApiHandler` allows access to the `useSession` and all other contexts set from the event and requestContext.
> 
> Make sure to import the hooks from the right construct as required.
> 
> ```ts
> import {
>   WebSocketApiHandler,
>   useHeaders,
>   useQueryParam,
> } from 'sst/node/websocket-api'; // <-- instead of `sst/node/api`
> import { useSession } from 'sst/node/future/auth';
> 
> export const handler = WebSocketApiHandler(async () => {
>   const session = useSession();
>   const headers = useHeaders();
>   const params = useQueryParam('hello');
>   console.log(session, headers, params);
>   return {
>     statusCode: 200,
>   };
> });
> ```

The only issue that having same name is in auto imports. When auto importing, there's high chances of importing it from wring file. 

I feel a better import and hook strategy should be: 

```ts
import {useHeaders, useQueryParam} from "sst/node/request"
import { WebSocketApiHandler } from 'sst/node/websocket-api';
import { useSession } from 'sst/node/future/auth';

export const handler = WebSocketApiHandler(async () => {
  const session = useSession();
  const headers = useHeaders({type: "ws"}) // useHeaders("ws")
  const params = useQueryParam("hello", {type: "ws"}) // useQueryParam("hello", "ws")
  console.log(session, headers, params);
  return {
    statusCode: 200,
  };
});

```

thdxr:
^ I think I agree with trying to make the hooks work for both types of requests.  We can leave them in the `sst/node/api` package for now until we figure out a good structure (and backwards compatibility)

@justindra did you consider that approach?

justindra:
Just pushed up the changes with it as a parameter and defaults to `api` for backwards compatibility. I kept some of the hooks that are WebSocket specific to still be kept in the `websocket-api/index.ts` file. 

Let me know what you think!

thdxr:
I think what I was suggestion was
`useHeaders()` would check for both event types and depending on what it found it would return the right value

justindra:
Gotcha! Yea, I could probably just use the `useContextType` hook and then just use that. Because realistically that should always return the right type it needs. We shouldn't be looking for a `api` header inside of a `WebSocketHandler`

justindra:
@thdxr All ready to go there!

PR #2831, opened 2023-05-03
Allow packages to be marked as "internal" for Pothos extractor

estyrke:
As mentioned by various people, for example at https://discord.com/channels/983865673656705025/983866416832864350/1080219063252435115, the Pothos extractor fails when importing "internal" packages, for example if I have an Enum type in my `core` package and import it by reference into my `functions` package (`import { MyEnum } from `@myproject/core/file`), that import will be stripped when the Pothos extractor runs.

This is my stab at a solution to this problem, by allowing the user to specify a list of package names that should be considered internal, and subsequently inlined into the schema generated by esbuild. Maybe the naming (and certainly the docstring) could be improved to better clarify the intent of the feature.

### DIFF
--- a/.changeset/chilled-jobs-march.md
+++ b/.changeset/chilled-jobs-march.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Allow packages to be marked as "internal" for Pothos extractor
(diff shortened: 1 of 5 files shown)


estyrke:
More references:
https://discord.com/channels/983865673656705025/983866416832864350/1083131147388719224
https://discord.com/channels/983865673656705025/1042862909396635728/1067149696411369583

jayair:
I'll tell @thdxr about it.

PR #2833, opened 2023-05-03
chore: add sveltekit link to docs index

TGlide:


jayair:
Haha so quick. We still need to add other docs for it but thanks!

TGlide:
@jayair sorry, I got excited 😆

PR #2835, opened 2023-05-04
fix(remix): delete deprecated options in favor of new supported ones

danestves:
The previous options `serverBuildTarget` were deprecated in favor of a combination of options from the remix config, this PR deletes the deprecated ones and adds the new ones to maintain the same behavior

More info: https://remix.run/docs/en/1.16.0/pages/v2#node-cjs

### DIFF
--- a/.changeset/young-hairs-shout.md
+++ b/.changeset/young-hairs-shout.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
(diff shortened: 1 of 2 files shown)


sockthedev (approved):

sockthedev:
Can't wait till we can point at the ESM output.

PR #2838, opened 2023-05-04
fix: set header keys to be lowercase for consistency

justindra:
Depending on your clients, sometimes headers are sent as either "Authorization" or "authorization", either capitalized or all lower case. Because of this, would be good to not have to make that mistake. So by default we always lowercase the header keys.

### DIFF
--- a/.changeset/six-waves-mate.md
+++ b/.changeset/six-waves-mate.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+API Headers can have different cases based on the client, this lowercases everything to keep it consistent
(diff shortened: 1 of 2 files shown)


jlarmstrongiv:
I would love to see this PR merged!  I‘ve seen several codebase comments referencing this issue like:

```
    // TODO: Once https://github.com/sst/sst/pull/2838 is merged,
    // then we should no longer need to check both casing for the header.

  // WebSocket may also set the token in the protocol header
  // TODO: Once https://github.com/sst/sst/pull/2838 is merged,
  // then we should no longer need to check both casing for the header.

```

Plus, I’m checking for both header casings in my personal sites too.

thdxr:

PR #2839, opened 2023-05-04
Job: Add support for runtime setup commands

gslandtreter:
This commit adds support for running commands inside the Job runtime environment before the actual node entrypoint. This allows users to run setup commands such as installing packages or missing libraries without having to modify the lambda base image.

We have this requirement on a project where we don't want to modify the lambda images but still need to run simple setup tasks.

Thanks.

### DIFF
--- a/packages/sst/src/constructs/Job.ts
+++ b/packages/sst/src/constructs/Job.ts
@@ -146,6 +146,21 @@ export interface JobProps {
    * ```
    */
   logRetention?: Lowercase<keyof typeof RetentionDays>;
+  /**
+   * Commands to be executed during runtime before the job handler entrypoint.
+   * @default []
+   * @example
+   * ```js
+   * new Job(stack, "MyJob", {
+   *   handler: "src/job.handler",
+   *   setupCommands: [
+   *     "yum install google-chrome",
+   *     "npm install",
+   *   ],
+   * })
+   *```
+   */
+  setupCommands?: string[];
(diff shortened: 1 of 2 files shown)


jayair:
Hmm I wonder if the better option here just might be to customize the docker image?

gslandtreter:
@jayair Thanks for the feedback!

Customizing the image is probably the standard path. However, there are scenarios where it's easier to just run setup commands during runtime, and having this functionality doesn't add a lot to the code base.

jayair:
Yeah let me think about it. We might want to support Docker properly here.

thiagozf:
It would be nice to have both options: setting up a custom Docker image **and** run pre/post commands. I have the same use case of @gslandtreter: installing browsers (Playwright) for screenshots/scrapping.

jayair:
Docker we are working on soon!

fwang:
Hey @gslandtreter, I just added `container` runtime support to `Job` in v2.21.1.

You can override `CMD` of the Docker image. Here's an example — https://docs.sst.dev/constructs/Job#creating-a-container-job

Does that work for you?

gslandtreter:
Hi @fwang, we heard the news! 

Although we have not tested it yet, I'm confident it will work perfectly for our use case. Appreciate your support and the fact that you guys are hearing the community when it comes to adding new features.

Feel free to close this PR!
Thanks!

PR #2846, opened 2023-05-07
clean: existsAsync function's arguments

piroz:
I thought fs accessed a file path, so I changed the argument to that.


thdxr:

PR #2849, opened 2023-05-08
bug: Dev role defined for the function cannot be assumed by Lambda

### LINKED ISSUE #2848: NextJsSite -- Tried to start a few versions locally and get the following error:
" CREATE_FAILED Resource handler returned message: "The role defined for the function cannot be assumed by Lambda. (Service: Lambda, Status Code: 400, Request ID: 599fd384-26c2-42ec-a937-b890f6f22c8c)" (RequestToken: f533030b-2cdb-3807-388b-efb80fc469af, HandlerErrorCode: InvalidRequest)
|  Site LogRetentionaae0aa3c5b4d4f87b02d85b201efdd8a AWS::Lambda::Function CREATE_FAILED Resource creation cancelled"

version 2.84, 2.85, 2.87
node v18.16.0
ubuntu 22.04

***After a week of successful testing and deploys as late as yesterday.

danecwalker:
This change resolves #2848 by specifying the Lambda Service Policy when creating the lambda function for the SsrSite server.

### DIFF
--- a/.changeset/weak-books-design.md
+++ b/.changeset/weak-books-design.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+SsrSite: Fix The role defined for the function cannot be assumed by Lambda.
--- a/.changeset/weak-books-design.md
+++ b/.changeset/weak-books-design.md
@@ -2,4 +2,4 @@
 "sst": patch
 ---
 
-SsrSite: Fix The role defined for the function cannot be assumed by Lambda.
+SsrSite: Fix dev server role cannot be assumed by Lambda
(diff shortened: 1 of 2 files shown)


fwang:
Hey @danecwalker, thanks for digging into the code!

I made a minor change to allow both the `AccountPrincipal` and `ServicePrincipal` to assume the role. The reason we need the former is b/c the local user needs to be able to assume when working locally, ie. `sst bind next dev`.

fwang:
Released in v2.8.8

danecwalker:
> Hey @danecwalker, thanks for digging into the code!
> 
> I made a minor change to allow both the `AccountPrincipal` and `ServicePrincipal` to assume the role. The reason we need the former is b/c the local user needs to be able to assume when working locally, ie. `sst bind next dev`.

No Problem. Happy to help!

PR #2859, opened 2023-05-10
docs(SsrSite): generalise attachPermissions

nilskj:
The other SSR frameworks that extend the attachPermissions method has the mention of Astro, even though they are not using Astro. See SvelteKit as example:
https://docs.sst.dev/constructs/SvelteKitSite#attachpermissions
This simply generalises to "framework" instead in the docstring


jayair:
Thanks!

Issue #2855, opened 2023-05-09
Support Java 17 runtime


julbrs:
Hi!

CDK is supporting Java 17 since [2.77](https://github.com/aws/aws-cdk/releases/tag/v2.77.0). I would love to see this in SST now 🤟 

I can work on a PR if needed ;) 

Thanks !

jayair:
Ah would appreciate it!

julbrs:
It's here @jayair 🤓 #2870 
(require CDK version bump 🙈)

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #2870, opened 2023-05-11
sst: support java 17

julbrs:
This PR is providing Java 17 support.
include CDK update to 2.79

refs: #2855


fwang:
Thanks @julbrs! Java 17 runtime is now supported in 2.18.0

PR #2877, opened 2023-05-12
Allow setting architecture on `SsrFunction`

### LINKED ISSUE #2876: Bug: sst does not respsect nextjs server architecture changes
The following setup still deploys an arm64 server lambda:
```typescript
import { Architecture } from "aws-cdk-lib/aws-lambda";
import { NextjsSite, StackContext } from "sst/constructs";

function MyStack({ stack }: StackContext) {
  const site = new NextjsSite(stack, "NextjsApp", {
      path: appRoot,
      environment,
      cdk: {
        server: {
          architecture: Architecture.X86_64,
        }
      }
  });
}
```
This is most likely because the [`SsrFunction` constructor](https://github.com/serverless-stack/sst/blob/192066d2939dc98da123aa3a83bfbf156158f483/packages/sst/src/constructs/SsrFunction.ts#L133) doesn't destructure `architecture` from props and just uses ARM_64.

relsunkaev:
fix: allow setting SsrFunction architecture

closes #2876


topaxi:
I arrived here as we are currently using a package that is only distribution for x86. It would be nice if we could get this merged :blush:

jayair:
Sorry let me check on this.

fwang:

Thanks @relsunkaev.

topaxi:
Thank you for the quick response time :bow:

PR #2878, opened 2023-05-12
[Fix/Parallel Project Init]: Skip cleanup of very recent config files

jamesgibbons92:
Aims to solve 
https://discord.com/channels/983865673656705025/1106299750636068946

Tries to avoid the situation where another sst process is spawned before another one finishes then the first one deletes the config files of the other parallel process

### DIFF
--- a/.changeset/chilled-books-learn.md
+++ b/.changeset/chilled-books-learn.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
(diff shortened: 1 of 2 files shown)


jayair:
I wonder if for this we should show a message that another process is running?

jayair:
@thdxr?

fwang:
Hasn't come up too often. I think that can be another PR. Let's merge this one first.

PR #2879, opened 2023-05-13
chore: replace zip-local with adm-zip

estyrke:
The zip-local package has dependencies with security vulnerabilities:

* https://nvd.nist.gov/vuln/detail/CVE-2022-48285
* https://nvd.nist.gov/vuln/detail/CVE-2021-43138

There is an open pull request on zip-local to fix this (https://github.com/Mostafa-Samir/zip-local/pull/18), but it's two weeks old and the maintainer has not reacted to it. Maybe it's easier to just use adm-zip instead, since it's already in the deps anyway.

I have tried to verify that it's still possible to build java functions, but my experience with Java in Lambda is nil, so I might have broken something.

### DIFF
--- a/.changeset/shaggy-panthers-vanish.md
+++ b/.changeset/shaggy-panthers-vanish.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Replace zip-local package with adm-zip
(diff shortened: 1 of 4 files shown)


thdxr:
thank you!

PR #2880, opened 2023-05-14
feat: SsrSite functionAssociations

tyrauber:
This change modifies SsrSite’s `buildDefaultBehaviorForRegional` in the same manner as `buildDefaultBehaviorForEdge`, allowing for additional `functionAssociations` to be passed to the SsrSite (NextJsSite) construct through `cfDistributionProps.defaultBehavior.functionAssocations`.

This change (already implemented for edge distributions) will allow sites deployed through regional distribution to add additional distribution functions.

Currently, the default function can only be overwritten.

With this change, a VIEWER_RESPONSE function could be added - say to geolocate the user with ‘cloudfront-viewer-country’ - without overwriting the required default VIEWER_REQUEST function.


fwang:
Thanks @tyrauber. Released in v2.8.23

PR #2888, opened 2023-05-17
doc: Update removal-policy.md

carlos-dasintel:
`setDefaultRemovalPolicy` expects a policy argument of type ` AppRemovalPolicy` not a string


jayair:
Oh no it's a string. That's not a CDK method, it's SST's.

PR #2894, opened 2023-05-18
Add IDE autocomplete instructions for Config construct

rcoundon:
Following conversation on Discord, this information was only present in the upgrade guide. This PR adds it to the `Config` construct docs too.


jayair:
Thanks!

PR #2903, opened 2023-05-22
Fix AWS profiles broken link

BSteffaniak:
The link to the AWS Profiles "cli configure files format" page has an invalid redirect on the Amazon side. It improperly encodes the hash character in the URL when attempting to link the user to the "#cli-configure-files-format" section. I've updated the link to have the correctly (un)encoded hash instead of having Amazon break the link. This is technically a bug on the Amazon side.


jayair:
Good catch1

PR #2909, opened 2023-05-24
future/auth: add prompt support for oidc

archieedwards:
This PR adds support to oidc for the `prompt` parameter in the authorisation URL.

This allows users to authenticate with an app on different accounts using the same provider.

## Links
- Discord [chat](https://discord.com/channels/983865673656705025/1109401107215687752/1109401107215687752)
- Google oidc [doc](https://developers.google.com/identity/openid-connect/openid-connect#authenticationuriparameters) showing prompt param

### DIFF
--- a/.changeset/four-llamas-talk.md
+++ b/.changeset/four-llamas-talk.md
@@ -1,0 +1,5 @@
+---
+"sst": minor
+---
+
+future/auth: add prompt support for oidc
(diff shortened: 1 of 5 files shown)


justindra:
I think it would be nice to add the types for the GoogleAdapter specifically to show the options available for the `prompt` when using the GoogleAdapter.

File is here: `packages/sst/src/node/future/auth/adapter/google.ts`

archieedwards:
@justindra updated for [google](https://developers.google.com/identity/openid-connect/openid-connect#authenticationuriparameters) and [microsoft](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc#send-the-sign-in-request) 👍 

I checked some the other auth providers too:
- Facebook has a separate [reauthenticate](https://developers.facebook.com/docs/facebook-login/guides/advanced/re-authentication/#nonce) instead of `prompt` so thought it wasn't worth adding
- Github doesn't seem to have a notion of reauthenticating users

PR #2915, opened 2023-05-26
sst: Allow filtering secrets by fallback values

digitaltoad:
In some cases its been useful to know if any fallback values are being used in certain stages. For instance, making sure a prod stage isn't using any fallback values.

### DIFF
--- a/.changeset/light-zebras-double.md
+++ b/.changeset/light-zebras-double.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Allow filtering secrets by fallback values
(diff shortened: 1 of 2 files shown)


jayair:
Can you share how this works?

digitaltoad:
Sure. This just adds a boolean flag for listing secrets in cli that will filter the list down to just fallback values. A use case that I have is making sure certain stages aren't using fallback values. It can be used like `sst secrets list --fallback`. Second to that, it adds a comment in the `env` layout to denote fallback values, again for easier parsing.

PR #2924, opened 2023-05-30
Suggest where to place sst create command.

ivanroosvelt:
According to used web framework, suggest where to place sst create command.
The cd command suggest to navigate to default project names, when the framework creates a new folder (astro, next, remix).
Svelte and Solid frameworks uses current directory and isn't necesary to move.


jayair:
Thanks!

PR #2928, opened 2023-05-31
Change authorizer function types to IFunction

Takadimi:
### Updates

- Change token authorizer function type from SST Function to CDK IFunction
- Change request authorizer function type from SST Function to CDK IFunction
- Add tests for referencing `currentVersion` when setting authorizer function on token and request authorizers

### Context

In a [recent help thread](https://discord.com/channels/983865673656705025/1113133552305975407/1113133552305975407) a user was attempting to leverage provisioned concurrency for their authorizer lambda function on an API Gateway V1 instance. This was technically doable by just referencing the authorizers `currentVersion` to ensure that a version got created that could be targeted by the provisioned concurrency configuration in Lambda and directly referenced by API Gateway.

That said, the type definitions for the authorizer functions didn't allow for using a `fn.currentVersion` which is of type `Version` and required directly passing SST's Function construct type.

Underneath the hood it looks like the CDK can accept an `IFunction` interface, which the `Version` type implements. Just switching from the concrete type to this interface seems to make Typescript happy.

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.Version.html
image
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.RequestAuthorizer.html
image
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.TokenAuthorizer.html
image

### DIFF
--- a/packages/sst/src/constructs/ApiGatewayV1Api.ts
+++ b/packages/sst/src/constructs/ApiGatewayV1Api.ts
@@ -359,7 +359,7 @@ export interface ApiGatewayV1ApiLambdaTokenAuthorizer
   /**
    * Used to create the authorizer function
    */
-  function?: Fn;
+  function?: lambda.IFunction;
   /**
    * The identity source for which authorization is requested.
    */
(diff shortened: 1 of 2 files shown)


thdxr:

PR #2934, opened 2023-06-01
Fix: Prevent overriding environment variables with values from .env

slyapustin:
Fixed https://github.com/serverless-stack/sst/issues/2523


thdxr:
thanks for the PR! i think the default is false so you could just remove the config object entirely

slyapustin (commented):
-

jayair:
@slyapustin any idea why this got closed?

ackinc:
@jayair Sergey seems to have deleted his fork of sst, so I don't think this PR's coming back. I created a new one doing the same thing here: https://github.com/sst/sst/pull/3305

PR #2936, opened 2023-06-01
adding a prop to allow configuring the server lambda log retention

webdevcody:


fwang:
Merged and released in v2.11.13

PR #2958, opened 2023-06-06
Fix/removal policy docs

ridzchandra:
[Here]

(https://docs.aws.amazon.com/cdk/api/v1/python/aws_cdk.core/RemovalPolicy.html#aws_cdk.core.RemovalPolicy.DESTROY:~:text=%5D%22%22%22)%0A)-,Attributes,-DESTROY) is the source docs page from "aws-cdk" 🙂

I also tested to check if it's correct, on my local lambda dev before making the corrections. ✌🏻

### DIFF
--- a/www/docs/advanced/removal-policy.md
+++ b/www/docs/advanced/removal-policy.md
@@ -3,42 +3,44 @@ title: Removal Policy
 description: "Learn how to set the removal policy for resources in your SST app."
 ---
 
+<!-- @format -->
+
 A Removal Policy controls what happens to the resource when it's being removed. This can happen in one of three situations:
 
-1. The resource is removed from the Stack.
-2. A change to the resource is made that requires it to be replaced.
-3. The stack is deleted, so all resources in it are removed.
+1. The resource is removed from the template, so CloudFormation stops managing it.
+2. A change to the resource is made that requires it to be replaced, so CloudFormation stops managing it.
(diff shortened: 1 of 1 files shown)


jayair:
Hmm I might be a little confused. Do you mind sharing why you wanted to make this change?

ridzchandra:
> https://docs.aws.amazon.com/cdk/api/v1/python/aws_cdk.core/RemovalPolicy.html#aws_cdk.core.RemovalPolicy.DESTROY:~:text=%5D%22%22%22)%0A)-,Attributes,-DESTROY

@jayair , SST docs are showing that the default policy is RETAIN and we need to manually set it to DESTROY when needed. However, As it shows in the aws-cdk-lib docs, DESTROY is the default, and we need to manually set it to RETAIN when needed.

fwang:
Just chiming in, there actually isn't a default removal policy. It's RETAIN for stateful resources, and DESTROY for non-stateful resources. It's briefly mentioned in the doc @ridzchandra shared above:
> Many stateful resources in the AWS Construct Library will accept a removalPolicy as a property, typically defaulting it to RETAIN.

This can be think of as SST/CDK will retain your data upon removal (ie. S3 bucket, logs, etc). And if you want a clean remove, you'd have to set the removal policy to `DESTROY`.

thdxr:

PR #2962, opened 2023-06-08
remove console.log from sqsQueueProps.visibilityTimeout if condition in Queue.js

bobbypiper:
remove console.log from sqsQueueProps.visibilityTimeout if condition


jayair:
Ooops

Issue #2988, opened 2023-06-19
[SsrSite]: Add function warming to all SsrSite constructs


SpencerDuball:
# Feature Request
I want to have the option to warm functions with all `SsrSite` constructs similar to how we can warm function with `NextjsSite` construct via open-next. The specific ask is to have the `SsrSite` updated to accept `warm: number` as an option keeping the same API as `NextjsSite` - this would allow all other constructs such as `RemixSite`, `SvelteKitSite`, `SolidStartSite`, `AstroSite` to keep instances warm.

# Overview
I was having a look at the implementation for the open-next to mimic this for `RemixSite` and try to get a PR pushed but there were two larger design decisions that I think warrant some discussion before pushing something.
1. `warm` feature should be part of the `SsrSite` API
2. The warmer function handler should be defined in `sst`, not as an implementation detail of the specific site you are deploying.

## 1. `warm` part of `SsrSite`
I was having a look at the `NextjsSite` to mimic this behavior and see that this warming feature is really two parts. There is a [`createWarmer`](https://github.com/serverless-stack/sst/blob/master/packages/sst/src/constructs/NextjsSite.ts#L179) function on the `NextjsSite` construct that will deploy a warmer lambda and a cron job to invoke this lambda. Then there is the actual warmer lambda (which is part of the open-next package) [here](https://github.com/serverless-stack/open-next/blob/main/packages/open-next/src/adapters/warmer-function.ts).

The reason this could be generalized for all `SsrSite` constructs is because there is only ever one lambda for these regional deploys (maybe many instances, but only one lambda arn) which exists as `this.serverLambdaForRegional`. There is nothing unique about the `NextjsSite` that would preclude the others from using this exact same `this.createWarmer` function.

## 2. Warmer Handler in SST, not open-next
The next thought would be to have the actual warmer lambda be defined as part of SST not open-next (of any other type of site). Currently, the `NextjsSite` statically defines this location for the warmer lambda as `.open-next/warmer-function.ts` but there is also nothing unique to `NextjsSite` that would preclude us from making this function generic to all constructs. It should be possible to have this pre-bundled lambda be included as part of SST itself.

I would be happy to create a PR for this, the only issue I could see if that open-next is not specific to SST. Would moving this handler lambda to SST be a deal breaker for compatibility with other deploy options that open-next supports? For example, do we _need_ to have the warmer lambda be `.open-next/warmer-function.ts` for any reason?

SpencerDuball:
[...]

PR #2996, opened 2023-06-20
enhancement: Added `warm` feature to all `SsrSite` constructs

SpencerDuball:
# Bug Fix
There is a TypeScript issue when running `pnpm build` where the value returned on `packages/sst/src/stacks/build.ts:66` from `babel.parse` can be either `const ast: ParseResult<babel.types.File> | null`. This is an issue as the next command `babel.traverse` can only accept the type `Node` - and `null` does not fit this interface.

After having a look into `@babel/core` it would appear that this value would only be null in the instance that there is not a config defined for babel:
- Issue is in the [parse](https://github.com/babel/babel/blob/main/packages/babel-core/src/parse.ts#L40) function.
- This returns [beginHiddenCallStack](https://github.com/babel/babel/blob/main/packages/babel-core/src/errors/rewrite-stack-trace.ts#L91) which returns `fn`. If we go back to `parse` we can see this `fn` is equal to either `parseRunner.sync` or `parseRunner.errback`.
- Inside [`parseRunner`](https://github.com/babel/babel/blob/main/packages/babel-core/src/parse.ts#L34) if config is null we return null.

In practice this should never happen as `loadConfig` should throw an error if there is a file with no configuration, but for our purposes this will basically just fix the type error by checking for `null`.

# Enhancement: SsrSite `warm` Feature
- Added a new function `warmer-function` which is compiled by SST and warms regionally deployed lambdas for `SsrSite`
- Updated the `build.mjs` to build the `warmer-function` when SST is built, that way there is no dependence upon the consumers of the `SsrSite` via `RemixSite`, `NextjsSite`, (and all others) to implement the function themselves.
- Removed the `warm` and `createWarmer` pieces of the `NextjsSite` as this now happens more generally for all `SsrSite`
- Updated `SsrSite` to implement the `warm` and `createWarmer`


SpencerDuball:
closes #2988

fwang:
Thanks @SpencerDuball!

There were some conflicts.. I copied over the changes in #3301

PR #2998, opened 2023-06-20
feat: add --to argument to diff else it can only use the default buildDir

berenddeboer:
This makes it compatible with the build command. Else build builds to its own, but diff still creates the .sst/dist folder, and that does not work well within nx.

As we know have to provide arguments to build and deploy to make this work, I suggest buildDir becomes a single setting at a single spot somewhere.

### DIFF
--- a/.changeset/many-moons-change.md
+++ b/.changeset/many-moons-change.md
@@ -1,0 +1,5 @@
+---
+"sst": minor
+---
+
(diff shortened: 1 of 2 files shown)


jayair:
Oh cool didn't notice this in your replies on Discord. @thdxr let's take a look?

PR #3000, opened 2023-06-20
Fix:  add missing policy to describe stacks

iloewensen:


fwang:
Hey @iloewensen, why does the migrator function needs to describe stacks?

mikejpeters:
Hi @fwang I'm not the author of this PR, but I believe this is related to the following issue:
https://github.com/serverless-stack/sst/issues/2983

and the discord thread for the issue: 
https://discord.com/channels/983865673656705025/1119328118952435712

fwang:
Ah.. got it! Thanks @mikejpeters

PR #3009, opened 2023-06-21
feat: add --to option to dev so builddir can be outside project root

berenddeboer:
This makes it comatible with the build and diff command. Not writing to the project directory helps with nx caching.

Given this buildDir now surfaces about everywhere, a global setting somewhere would be nice at some point.


berenddeboer:
This is similar to #2998.

fwang:
@thdxr take a quick look?

PR #3026, opened 2023-06-26
Fix broken link in example

JPVan:
Fix broken link in examples/api-sst-auth-google/README.md


fwang:
Thanks @JPVan!

The examples are generated from templates in `sst/create-sst`. I just copied over the fix.

PR #3015, opened 2023-06-21
feat: enforce ssl connections to S3 to stop triggering S3.5 of the AWS Foundational Security Best Practices

berenddeboer:
The bootstrap stack triggers various alarms for folks using Security Hub. This address S3.5. In case you're not using Security Hub (really an enterprise thing), see screenshot here. This wakes up the security department and they start to ask questions about SST.

image

PR #3032, opened 2023-06-26
feat: enforce S3 https connections

berenddeboer:
Similar to #3015 but now enforced for every bucket. This to start stopping SST buckets showing up in AWS Security Hub.

### DIFF
--- a/.changeset/angry-waves-marry.md
+++ b/.changeset/angry-waves-marry.md
@@ -1,5 +1,5 @@
 ---
-"sst": minor
+"sst": patch
 ---
 
-feat: enforce S3 https connections
+Bucket: enforce S3 https connections
--- a/.changeset/angry-waves-marry.md
+++ b/.changeset/angry-waves-marry.md
@@ -1,0 +1,5 @@
+---
+"sst": minor
+---
+
+feat: enforce S3 https connections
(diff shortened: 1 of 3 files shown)


DCzajkowski:
This change has broken our production builds when upgrading a minor version of SST @fwang. We didn't notice the problem in staging, because there was a cache hit on our test object. 

The break occurred because [CloudFront communicates with S3 over HTTP only](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/distribution-web-values-specify.html#DownloadDistValuesOriginProtocolPolicy)! This breaks every CloudFront that is in front of S3.

---

This is also very confusing, as the doc blocks on AWS CDK say the default is `false`.

Screen Shot 2024-03-08 at 3 51 14 PM

I don't think it was a good call for SST to change the default in this case.

---

In any case, I think this should be marked as a breaking change.

jayair:
Wait wasn't this change from last year?

DCzajkowski:
We were on 2.8.26 and upgraded to 2.40.3 a few days ago.

berenddeboer:
No way! Wow, didn't know S3 only did HTTP. But this change got merged months ago.

berenddeboer:
But I suggest you set it explicitly to true. That S3 origins only supports HTTP seems like an insane default. What's the internet path between your CloudFront and S3? How many intermediaries get to see and change that???

PR #3035, opened 2023-06-27
feat: allow a secret to be encrypted with a custom KMS key

berenddeboer:
It's often a requirement in enterprise environments that secrets must be encrypted with a custom key, and not use the AWS default key in your environment. The reason is that destroying of the stack means all data encrypted with that key is also no longer accessible (if AWS really destroys the key).

This also means I had to fix which credentials were accepted: you can now also change the default root user of your cluster, another enterprise requirement.

### DIFF
--- a/packages/sst/src/constructs/RDS.ts
+++ b/packages/sst/src/constructs/RDS.ts
@@ -398,12 +398,8 @@ export class RDS extends Construct implements SSTConstruct {
       );
     }
 
-    // Validate Secrets Manager is used for "credentials"
-    if (
-      props.credentials &&
-      !props.credentials.secret &&
-      props.credentials.password
-    ) {
+    // Validate Secrets Manager is used for "credentials" not password
+    if (props.credentials?.password) {
       throw new Error(
         `Only credentials managed by SecretManager are supported for the "cdk.cluster.credentials".`
       );
--- a/packages/sst/src/constructs/RDS.ts
+++ b/packages/sst/src/constructs/RDS.ts
@@ -22,6 +22,7 @@ import {
 import { Code, Function, Runtime } from "aws-cdk-lib/aws-lambda";
 import { PolicyStatement } from "aws-cdk-lib/aws-iam";
 import { ISecret } from "aws-cdk-lib/aws-secretsmanager";
+import { IKey } from "aws-cdk-lib/aws-kms";
 import { App } from "./App.js";
 import { Stack } from "./Stack.js";
 import { getFunctionRef, SSTConstruct, isCDKConstruct } from "./Construct.js";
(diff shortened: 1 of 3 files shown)


packages/sst/src/constructs/RDS.ts
```diff
@@ -393,7 +403,7 @@ export class RDS extends Construct implements SSTConstruct {
     }
 
     // Validate Secrets Manager is used for "credentials"
-    if (props.credentials && !props.credentials.secret) {
+    if (props.credentials && !props.credentials.secret && props.credentials.password) {
```
fwang: Why is this required? Can a credential have both `secret` and `password`?

berenddeboer: The original check forbad using `credentials` in `cdk`. The example in the test was providing credentials with your own password. This would not create a secret. If you specify credentials differently, a secret is created, but later, so at this point it is not set. But password is also not set. So I added this check to still pass the initial test, but allow the latter case. Hopefully this is clear? The way I provide credentials creates that secret.

This is what it looks like in CF:

```yaml
  RDSE0E96D00:
    Type: AWS::RDS::DBCluster
    Properties:
      CopyTagsToSnapshot: true
      DBClusterParameterGroupName: default.aurora-postgresql11
      Engine: aurora-postgresql
      EngineMode: serverless
      EngineVersion: "11.16"
      MasterUsername:
        Fn::Join:
          - ""
          - - "{{resolve:secretsmanager:"
            - Ref: RDSSecret3683CA93
            - :SecretString:username::}}
      MasterUserPassword:
        Fn::Join:
          - ""
          - - "{{resolve:secretsmanager:"
            - Ref: RDSSecret3683CA93
            - :SecretString:password::}}
      StorageEncrypted: true
      VpcSecurityGroupIds: []
    UpdateReplacePolicy: Snapshot
    DeletionPolicy: Snapshot
```

The password in the example had a token with `{{resolve:ssm}}`.

fwang: Got it! Thanks for the explanation.

Changed to this as the `props.credentials.secret` check seems to be not useful.
```diff
- if (props.credentials && !props.credentials.secret && props.credentials.password) {
+ if (props.credentials?.password) {
```

kaisarkuanysh:
with those changes my code doesnt work

```
import { StackContext, RDS, Bucket, Service } from "sst/constructs";
import * as ecs from "aws-cdk-lib/aws-ecs";
import * as ec2 from "aws-cdk-lib/aws-ec2";
import * as rds from "aws-cdk-lib/aws-rds";
import * as cdk from "aws-cdk-lib/core";

const MyStack = ({ stack }: StackContext) => {
  const rdsSecret = new rds.DatabaseSecret(stack, "DirectusRdsSecret", {
    username: "clusteradmin",
  });

  const database = new RDS(stack, "DirectusDatabase", {
    engine: "postgresql13.9",
    defaultDatabaseName: "directus",
    cdk: {
      cluster: {
        removalPolicy: cdk.RemovalPolicy.DESTROY,
        credentials: rds.Credentials.fromSecret(rdsSecret),
      },
    },
  });
}
```

Error:
```
Error: Only credentials managed by SecretManager are supported for the "cdk.cluster.credentials".

Trace: Error: Only credentials managed by SecretManager are supported for the "cdk.cluster.credentials".
    at RDS.validateCDKPropWhenIsClusterProps
```

Which is doesnt make sense cuz secrets are managed by secrets manager

PR #3044, opened 2023-06-27
Fix examples/api-sst-auth-google

JPVan:
- Correct use of ViteStaticSite
- Correct name of auth.handler
- Import SSTConfig as type 
- Update aws-cdk-lib and constructs


jayair:
Thank you!

PR #3072, opened 2023-07-03
refactor: drop deprecated runtimes

berenddeboer:
Many runtimes were obsolete, simply remove them.


fwang:
Thanks @berenddeboer!

I was thinking if we should annotate them as @deprecated or remove them. According to AWS doc, after a runtime is deprecated:

> you can no longer create or update functions that use the runtime ... ... Lambda does not block invocations of functions that use deprecated runtime. Function invocations continue indefinitely after the runtime reaches end of support.

If we removed the runtime, ppl using deprecated runtime would get a type error (ie. `nodejs12.x`), even though they are not "required" to update.

What do u think? /cc @thdxr

berenddeboer:
Hi @fwang no-one using SST would be using any of these old run-times. They are very very old. Secondly, you don't want to support them even, so if you keep them deprecated, it still means they technically should work.

fwang:
Thanks @berenddeboer, good point! Agreed.

I put back `nodejs12.x` and `dotnetcore3.1` for now as they were recently deprecated by Lambda. I know a few folks from the early days of SST that were using these two. They should have updated. But let's just give them a few more months in case they haven't 🥲

PR #3082, opened 2023-07-04
documentation: Correction to auth documentation

smithy127:
Code examples previously showed the `Auth` object wrapping another `Auth` object which is confusing to read. This corrects the docs to show the correct `API` object type.

### DIFF
--- a/www/docs/constructs/v1/Auth.tsdoc.md
+++ b/www/docs/constructs/v1/Auth.tsdoc.md
@@ -73,7 +73,7 @@ The API to attach auth routes to
 
 
 ```js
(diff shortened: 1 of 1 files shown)


jayair:
Hmm so this file is generated from the source files. But for some reason it's not updating. Let me check on this.

smithy127:
@jayair I suspected it might be the case but thought it was easier to illustrate the issue over a PR. Feel free to bin it once you get to the root issue.

fwang:
Thanks @smithy127!

I was able to finally track down a bug in the tsdoc generation script. Fixed!

PR #3085, opened 2023-07-04
Function: support container runtime

fwang:


lefnire:
Oh crap, this is huge guys! I think this release will land you a lot more on-the-fence users

lefnire:
If I'm understanding, we will point the handler to the folder containing `./Dockerfile`. Might I propose an optional `dockerfile` prop, which specifies the actual dockerfile (and `Dockerfile` is used otherwise). I've worked on multiple projects where there are shared files (eg a `./common` directory) which gets copied into each of multiple Dockerfiles. Here's my [current project's folder ](https://github.com/ocdevel/gnothi/tree/main/services/ml/python)for example

alex-wenner:
@lefnire I think [this](https://github.com/serverless-stack/sst/pull/3100) PR would fix your issue - `ContainerProps` would expose `file` seen [here](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.AssetImageCodeProps.html#file)

lefnire:
@alex-wenner perfect!

PR #3091, opened 2023-07-06
[SsrSite] feat: enable AWS_IAM auth on function URLs

### LINKED ISSUE #3090: [SsrSite] Lambda function URLs' auth is set to None
Related discussion on Discord: https://discord.com/channels/983865673656705025/1124011751973859438/1124011751973859438

# Issue

CloudFront can be bypassed by targeting the function URL directly.

This can be a problem if, for example:
- you want to enforce IP whitelisting via a CloudFront WAF,
- you want to protect (too) many functions from being maliciously provisionned (aka Denial of Wallet).

# Proposal

[] Enable the AWS_IAM authentication mechanism on function URLs
[] Configure a function@edge that signs CloudFront requests targeting a function URL

ealain:
Fixes #3090


khuezy:
Thanks for putting this together! I skimmed over the PR and it looks good, but I'll do a thorough one later.

Question: How much over head would this introduce? Can this be an opt-in configuration? Or is the overhead insignificant and the AWS_IAM auth type standard/best practice that we should use it by default?

ealain:
Hi! I think adding the opt-in configuration to this PR is a good idea.

In the future, when we have long-term real-world data, we can do a thorough benchmark on the performance and cost impact of the AWS_IAM auth type. Only then, we could create another PR to configure AWS_IAM by default.

ealain:
Hi @khuezy!

@StanHannebelle implemented the opt-in configuration. By default, function URLs remain public.

I also fixed an issue where the signing function would not produce the correct signature on image optimization request because the `url` query param was already encoded by CloudFront.

Now the PR is ready for review.

packages/sst/src/constructs/SsrSite.ts
```diff
@@ -983,10 +1009,12 @@ function handler(event) {
     const cfDistributionProps = cdk?.distribution || {};
 
     const fnUrl = this.serverLambdaForRegional!.addFunctionUrl({
-      authType: FunctionUrlAuthType.NONE,
+      authType: !this.props.enableIAMAuth
```
khuezy, line 1012: Super nit: The other patterns above does a truth check, while this one uses `!` as the first ternary.

packages/sst/support/signing-function/helpers/queryParameterBagToQueryString.ts
```diff
@@ -0,0 +1,32 @@
...
+ * This function does not encodeURIComponent the values,
+ * because the result is meant to be used in a JS object.
+ */
+export const queryParameterBagToQueryString = (
+  query: QueryParameterBag
+): string => {
+  const keys: Array<string> = [];
+  const serialized: Record<string, string> = {};
+
+  for (const key of Object.keys(query)) {
```
khuezy, line 13: is `request.queryString` guaranteed to be an object, even if the url doesn't include it? Object.keys on null/undefined will throw an error.

khuezy (approved):
LGTM, some minor comments. @fwang PTAL

packages/sst/src/constructs/SsrSite.ts
ealain: Fixed

packages/sst/support/signing-function/helpers/queryParameterBagToQueryString.ts
ealain: Yes, this is guaranteed by the  `queryStringToQueryParameterBag` function where [`query` is initialized with `const query = {}`](https://github.com/ealain/sst/blob/67128b3c02c1d5e65a3ea167551330186be7016b/packages/sst/support/signing-function/helpers/queryStringToQueryParameterBag.ts#L6).

luc-ferron:
I think a few unit tests to validate the feature is working, and that it doesn't get broken by others in the future, would be a good idea. Thanks for putting all this together, I'm going to remove my home made support of IAM in favor of this when it gets in.

ealain:
@luc-ferron thank you for your support!

I agree that tests will definitely help the project in the future 🚀

However, I think adding tests in this PR is not appropriate because there is no testing framework already set up. This topic should be addressed by the core team.

luc-ferron:
> @luc-ferron thank you for your support!
> 
> I agree that tests will definitely help the project in the future 🚀
> 
> However, I think adding tests in this PR is not appropriate because there is no testing framework already set up. This topic should be addressed by the core team.

I do see a lot of tests in this folder for each high level construct: 

https://github.com/serverless-stack/sst/tree/master/packages/sst/test/constructs

Adding some validation of the additional feature in NextJSSite that inherits SsrSite could be a quick option?

ealain:
I just changed the dependency @aws-sdk/signature-v4 which was moved to @smithy/signature-v4 a few days ago 😄 

@jayair @fwang What are your thoughts on what should be tested here?

jayair:
Yup let me check on it with Frank.

fwang:
Made some cosmetic changes and added unit tests.

Thanks @ealain @StanHannebelle!

khuezy:
🙏 How much latency does it actually add?

fwang:
From my brief testing.. it seems the signing function takes ~200 to boot up (cold start) and <20ms to run.

khuezy:
Not bad on the warm run, can we warm up the signing function?

khuezy:
Actually, would it be possible to implement the signing function in rust or something to minimize the coldstart?

PR #3095, opened 2023-07-06
fix: ensure build directory exists when building RemixSite

michaelgmcd:
I've found that occasionally, the "copyFileSync" calls fail in CI because the build directory does not exist. I've made a slight change to create it if it doesn't.


jayair:
Thanks! I'll ask Frank to take a look.

fwang:
Thanks @michaelgmcd! Good catch.

PR #3099, opened 2023-07-07
Add arm64 support for the Rust runtime

Nirlah:
Modify the build command based on the user’s function configuration.


fwang:
Thanks @Nirlah!

I made a cosmetic change hoping to make it a bit more readable.

PR #3100, opened 2023-07-07
feat: Adds container props to Function construct

alex-wenner:
## Problem Statement 
When using the new Function construct support for containers, it seemed like there were a few gaps. The underlying props weren't previously exposed. For example, if you wanted to use one container for many different functions, when using CDK you could pass in CMD or Entrypoint overrides. In the latest release, that functionality did not exist.

## Proposal
Further extend this functionality, to support `AssetImageCodeProps`.

## Implementation
I took the approach that is used for `PythonProps`, `JavaProps`, etc. I have added `ContainerProps` which is just an extension of the `AssetImageCodeProps` Interface. So when using this, it looks like the following: 

```javascript
new sst.Function(this, "sample", { 
  runtime: "container",
  handler: "path/to/dockerfile",
  container: {
    cmd: ["cmdOverride"],
    entrypoint: ["entrypointOverride"]
    }
});
```

## Changes
packages/sst/src/constructs/Function.ts
packages/sst/tests/Function.test.ts

### DIFF
--- a/packages/sst/src/constructs/Function.ts
+++ b/packages/sst/src/constructs/Function.ts
@@ -31,6 +31,7 @@ import { useWarning } from "./util/warning.js";
 import {
   Architecture,
   AssetCode,
+  AssetImageCodeProps,
   CfnFunction,
   Code,
   Function as CDKFunction,
@@ -142,9 +143,9 @@ export interface FunctionProps
   python?: PythonProps;
 
   /**
-   * Used to configure image function properties
+   * Used to configure container function properties
    */
-  //image?: ImageProps;
+  container?: ContainerProps;
 
   /**
    * Hooks to run before and after function builds
@@ -534,6 +535,23 @@ export interface PythonProps {
   installCommands?: string[];
 }
 
+export interface ContainerProps extends AssetImageCodeProps {
+  /** 
+   * Extends AssetImageCodeProps to allow container implementations to specify additional properties.
+   * ```js
+  * new Function(stack, "Function", {
+  *   container: {
+  *     cmd: ['python3'],
+  *     buildArgs: {
+  *         ARG1 : "ARG"
+  *     }
+  * 
+  *   }
+  * })
+  * ```
+*/
+}
+
 /**
  * Used to configure Java package build options
  */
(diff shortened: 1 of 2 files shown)


jayair:
Thanks, I'll ask Frank to take a look.

alex-wenner:
@fwang @jayair 

Any chance you all could check this out soon? I am blocked with this unfortunately until _some_ functionality like this gets implemented 😄 

Let me know your thoughts!

jayair:
Yeah let me ping Frank again, he's rolling our containers for the Job construct.

fwang:
@alex-wenner in v2.21.1, u can override `CMD`. It seems that's what u need for now to have multiple functions share the same image. Here's an example — https://docs.sst.dev/constructs/Function#creating-a-container-function

I didn't want to allow all `AssetImageCodeProps` options at the moment. In most cases, it doesn't make sense to override things like `entrypoint` or `platform`(inferred from function's architecture). I want to apply some restricts, while learning what ppl want to customize. And see if we can provide a more convenient way to do what ppl try to do.

Let me know if that makes sense. And if there's anything from `AssetImageCodeProps` u'd like to customize, I'm 100% willing to add it in.

alex-wenner:
> @alex-wenner in v2.21.1, u can override `CMD`. It seems that's what u need for now to have multiple functions share the same image. Here's an example — https://docs.sst.dev/constructs/Function#creating-a-container-function
> 
> 
> 
> I didn't want to allow all `AssetImageCodeProps` options at the moment. In most cases, it doesn't make sense to override things like `entrypoint` or `platform`(inferred from function's architecture). I want to apply some restricts, while learning what ppl want to customize. And see if we can provide a more convenient way to do what ppl try to do.
> 
> 
> 
> Let me know if that makes sense. And if there's anything else from `AssetImageCodeProps` u'd like to customize, I'm 100% willing to add it in.

Got it! That makes sense. I see what you are referring to. I think also entrypoint and buildArgs would be all I personally need to be exposed... I did see someone mentioning a file path override which I would also be in support of, I see value there. But other than that - that works great! Thank you 😀

fwang:
> I think also entrypoint
Why do u need to override `entrypoint`, is ur Dockerfile not extending `FROM` AWS' images?

> I did see someone mentioning a file path
Is it on Discord, can u share a link? I wanted to learn more about their use case.

alex-wenner:
@fwang 

That is a fair challenge, if one CMD or Entrypoint is exposed you don't need the other. Especially if it's important to the underlying work you have done to make this possible. I think best practice is to override CMD as is so, I'm with you there. 

As far as file prop request goes, I referenced this PR in a merged PR... I think if you scroll up in conversation you should be able to see it, if not lmk and I will link it. 

Thanks!

fwang:
Both command and file can now be configured — https://docs.sst.dev/constructs/Function#containerprops

PR #3101, opened 2023-07-08
Output build errors in Rust runtime

gvidon:
If a build in Rust fails there is no other way to debug. This makes debugging of the issues very hard!
image

### DIFF
--- a/packages/sst/src/runtime/handlers/rust.ts
+++ b/packages/sst/src/runtime/handlers/rust.ts
@@ -78,6 +78,7 @@ export const useRustHandler = Context.memo(async () => {
             path.join(input.out, "handler")
           );
         } catch (ex) {
+          consile.log(ex);
           throw new VisibleError("Failed to build");
         }
       }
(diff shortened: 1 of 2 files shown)


Nirlah (commented):
...

gvidon:
Fixed a typo, thanks for pointing out!

packages/sst/src/runtime/handlers/rust.ts
```diff
@@ -78,6 +78,7 @@ export const useRustHandler = Context.memo(async () => {
             path.join(input.out, "handler")
           );
         } catch (ex) {
+          console.log(ex);
           throw new VisibleError("Failed to build");
```
fwang, line 82: @gvidon instead of `console.log`, can you try this in both places? And see if errors details show up.

```diff
- throw new VisibleError("Failed to build");
+ return {
+   type: "error",
+   errors: [String(ex)],
+ };
```

^ for context. This is what we do for other runtimes, not sure why we r throwing here. @thdxr might know better.

fwang:
@gvidon I added a comment. Let me know if that works for you.

packages/sst/src/runtime/handlers/rust.ts
gvidon: @fwang thanks, replaced in PR.

fwang:
Thanks @gvidon!

gvidon:
Thank for quick turnaround, and for the really cool tool!

PR #3103, opened 2023-07-10
feat: run only one gradle build per java directory

berenddeboer:
Currently a gradle build job is started for every Function even if it points to the same directory, causing massive machine load and build errors.

This could still start two builds, but only in succession. Ideally there's some mop up logic in the parent so it starts builds only once. But this is probably good enough for most scenarios.

Instead of `gradlew` ideally I would have the ability to run a `gradle:build` or so if one is defined in `package.json`. The idea here is to make sure gradle does not write output to the project directory, as that is annoying in monorepo projects. So would need to specify `--project-output-path` I believe.

But that might be another PR, this already greatly improves our builds.

### DIFF
--- a/.changeset/early-cameras-leave.md
+++ b/.changeset/early-cameras-leave.md
@@ -1,5 +1,5 @@
 ---
-"sst": minor
+"sst": patch
 ---
 
-feat: run only one gradle build per java directory
+Function/java: run only one gradle build per directory
--- a/.changeset/early-cameras-leave.md
+++ b/.changeset/early-cameras-leave.md
@@ -1,0 +1,5 @@
+---
+"sst": minor
+---
+
+feat: run only one gradle build per java directory
--- a/packages/sst/src/runtime/handlers/java.ts
+++ b/packages/sst/src/runtime/handlers/java.ts
@@ -18,8 +18,7 @@ export const useJavaHandler = Context.memo(async () => {
   const handlers = useRuntimeHandlers();
   const processes = new Map<string, ChildProcessWithoutNullStreams>();
   const sources = new Map<string, string>();
-  const runningBuilds = new Map<string, ReturnType<typeof execAsync>>();
-  const zippedBuilds = new Map<string, boolean>();
+  const runningBuilds = new Map<string, Promise<void>>();
 
   handlers.register({
     shouldBuild: (input) => {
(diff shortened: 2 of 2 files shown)


fwang:
Thanks @berenddeboer! Made some cosmetic changes.

I got an error when invoking the function after rebuild:
```
|  Invoked example.Handler::handleRequest
|  +13ms   Class not found: example.Handler: java.lang.ClassNotFoundException
|  +13ms   java.lang.ClassNotFoundException: example.Handler
|  +13ms        at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:476)
|  +13ms        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:589)
|  +14ms        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
|  +14ms        at java.base/java.lang.Class.forName0(Native Method)
|  +14ms        at java.base/java.lang.Class.forName(Class.java:398)
|  Error: Class not found: example.Handler
```

I think we need to clear `runningBuilds` and `zippedBuilds`, otherwise subsequent builds in the same directory won't trigger `gradle build` again?

berenddeboer:
Ah yes, totally right. I've only run it in just once mode, you're running this in dev mode.

berenddeboer:
And I appreciate your work on this, it's currently a big issue for our java code, so having this fix would go a long way.

berenddeboer:
And great changes, just pulled them in to have a look. Need to look into this context thing a bit more.

fwang:
Good point! I reworked the code:
1. Combining the `build` and `unzip` steps into 1 step
2. `runningBuilds` is set right after the build starts (before build/unzip actually runs)
3. `runningBuilds` is cleared after build/unzip is done

I briefly tested on my hello world setup. Can you give it a try?

berenddeboer:
Executing perfectly. The only thing I can see happening is that it is not impossible (but perhaps unlikely) the build could be executed twice, but only in succession (and gradle will detect it has been build, so no big deal at this stage).

Tested with: `while true; ps ax | grep appname=gradlew | wc -l; end`; saw just one java popping up. This is on a project where unpatched we see 9!

Btw, how do you execute new versions of SST in projects? I just copy the .js files after build to some target project, possibly could even symlink dist to `node_modules/sst` I suppose. Is there an easier way even? I'm not sure I can trust running the sst "binary" from a different location and really be sure it picks up all its files from the other location instead of the local node_modules folder.

fwang:
> Btw, how do you execute new versions of SST in projects?
1. Run `pnpm link --global` in SST's `packages/sst/dist` folder
  (^ if the command fails, copy `packages/sst/package.json` to `packages/sst/dist/packages.json` and remove the `publishConfig` block from line 2-5.. and then try again)
2. Run `pnpm link --global sst` in your repo
That's it.

Note that if you run `pnpm install` in your repo again, it will go back to use the one from npm.

berenddeboer:
Hi @fwang, any chance this can make it into a release soon? That would solve our biggest build problem.

fwang:
Will be in the next release.

^ sorry.. thought u were still testing.

Issue #3105, opened 2023-07-11
No ability to set custom boundary policy permissions for SST Bootstrap stack


MattJenko:
We need to be able to specify the boundary policy used by the SST bootstrap stack, in the same manner as the CDK stack options have a customPermissionsBoundary.

The addition of the MetadataHandler Lambda and the CDK CustomResource handler has run afoul of our organisations standard Boundary Policy, unless we specify a specific one that is provided.

Given that the option already exists for the purpose of provisioning the CDK bootstrap stack through the CDK options, we should be able to reuse the existing option to add the same policy to the SST Bootstrap stack.

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #3106, opened 2023-07-11
Feat/boundary policy sst bootstrap

MattJenko:
This is intended to fix issue #3105.

Extend the custom permissions boundary to apply to both the cdk and sst bootstrap stacks.


jayair:
Thanks for this!

PR #3109, opened 2023-07-11
fix: SsrFunction with id ServerFunction already exists

mathisobadia:
Starting with sst version 2.19.2 using NextjsSite construct failed to deploy with the error:

Trace: Error: SsrFunction with id "ServerFunction" already exists.

changing the ids in the construct fixes the issue, the weird thing is that theoretically only regional or edge should be used at once and not both but I'm not sure why the function that checks for duplicate ids sees both.

Note the error started appearing with 2.19.2 because it adds the id and metadata to SsrFunction construct

### DIFF
--- a/packages/sst/src/constructs/NextjsSite.ts
+++ b/packages/sst/src/constructs/NextjsSite.ts
@@ -145,7 +145,7 @@ export class NextjsSite extends SsrSite {
       environment,
       cdk,
     } = this.props;
-    return new SsrFunction(this, `ServerFunction`, {
+    return new SsrFunction(this, `ServerFunctionRegional`, {
       description: "Next.js server",
       bundle: path.join(this.props.path, ".open-next", "server-function"),
       handler: "index.handler",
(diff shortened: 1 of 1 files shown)


khuezy:
Should this also change:
```
return new CdkFunction(this, `ServerFunction`, {
```
to:
```
return new CdkFunction(this, this.id, {
```
in https://github.com/serverless-stack/sst/blob/master/packages/sst/src/constructs/SsrFunction.ts#L150

mathisobadia:
> Should this also change:
> 
> ```
> return new CdkFunction(this, `ServerFunction`, {
> ```
> 
> to:
> 
> ```
> return new CdkFunction(this, this.id, {
> ```
> 
> in https://github.com/serverless-stack/sst/blob/master/packages/sst/src/constructs/SsrFunction.ts#L150

I don't think so because the function that checks for unique ids only checks sst constructs and that's a raw cdk construct
https://github.com/serverless-stack/sst/blob/87f842b04cb2afd4f254e2846939d04b343455d3/packages/sst/src/constructs/App.ts#L485-L489

khuezy:
Cool. It is odd that the region/edge check is ran on both when they are mutually exclusive via the if/else condition.

jayair:
Will take a look!

khuezy:
After some investigation w/ Frank. This is due to multiple SsrSites clashing on the same id. So the fix here won't work w/ other setups.

fwang:
Fixed in 2.21.7

@mathisobadia Thanks for the PR. After digging into the details w/ @khuezy on Discord, it seems the issue happens when there are more than 1 site defined in the app. Fixed in 2.21.7.

PR #3116, opened 2023-07-12
Function: Add support for cgo and ldflags for go runtime

pzeinlinger:
Fixes #3097.

This commit adds a configurable ldflags & cgo to go handler functions.

PR #3133, opened 2023-07-18
Function/go: fix incorrect go -tags flag

pzeinlinger:
I introduced a -t flag in PR #3116. This works on my machine, because I have an alias set-up. In reality, the build flag is not "-t" but "-tags".

@fwang


packages/sst/src/runtime/handlers/go.ts
```diff
@@ -72,7 +72,7 @@ export const useGoHandler = Context.memo(async () => {
               "go",
               "build",
               ...(ldFlags ? [`-ldflags "${ldFlags.join(" ")}"`] : []),
-              ...(buildTags ? [`-t ${buildTags.join(" ")}`] : []),
+              ...(buildTags ? [`-tags ${buildTags.join(" ")}`] : []),
```
michaelgmcd, line 75: You're also defaulting buildTags to `[]`, so this is always going to be true. Maybe remove that default or check the length here?

packages/sst/src/runtime/handlers/go.ts
```diff
@@ -101,7 +101,7 @@ export const useGoHandler = Context.memo(async () => {
...
```
michaelgmcd, line 104: Same here

packages/sst/src/runtime/handlers/go.ts
pzeinlinger: Thanks! That slipped through in the last PR.

PR #3134, opened 2023-07-18
Update auth.md

joerter:
Add an explanation on how to tell if the request was unauthenticated. Based on this example: https://sst.dev/examples/how-to-add-google-login-to-your-sst-apps.html


jayair:
Thanks!

PR #3138, opened 2023-07-18
console: fix rendering sets in DynamoDB table view.

vincemtnz:
DyanmoDB sets are currently rendered as an empty object `{}` in the console. I've added a replacer function to convert them to an array before stringifying. 

Note: this only addresses rendering sets in the table view. The DynamoDB editor is a whole different story since converting it to an array there would change the data type to a DynamoDB List after saving.


jayair:
Oh interesting. Asking @thdxr to take a look.

Issue #3140, opened 2023-07-18
Cannot find module @aws-sdk/types with yarn pnpm linker


jonmast:
Both the `sst` and `sst-aws-cdk` package have missing dependencies that cause errors with yarn's pnpm linker. Unlike pnpm yarn doesn't allow dependencies to access other dependencies they haven't explicitly declared. This is roughly equivalent to setting `hoist=false` on pnpm.

For example, `sst` [imports @aws-sdk/types in credentials.ts](https://github.com/serverless-stack/sst/blob/master/packages/sst/src/credentials.ts#L1), but doesn't declare it as a dependency. When run with yarn it results in the error `Error: Cannot find module '@aws-sdk/types'`.

I can work around this by manually specifying the dependencies in `.yarnrc.yaml`, but ideally the `package.json` would be updated to appropriately declare the dependencies.

```yaml
packageExtensions:
  "sst@*":
    peerDependencies:
      "@aws-sdk/types": "*"
  "sst-aws-cdk@*":
    peerDependencies:
      "@aws-cdk/cx-api": "*"
      "@aws-cdk/cloud-assembly-schema": "*"
      "@aws-cdk/cloudformation-diff": "*"
      "aws-sdk": "*"
    dependencies:
      "fs-extra": "*"
```

jayair:
Can you post this on #help on Discord? Need some more info on this.

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #3158, opened 2023-07-26
Delete side effect only @aws-sdk/types import

jonmast:
"Bare" imports that don't name what they are importing are never elided by Typescript and end up in the compiled JS. That doesn't make any sense for this package since it only has types.

I don't know why this was originally added, but it still typechecks without it now so I don't think it's necessary.

Partial fix for #3140


jayair:
Oh let me check on this with Dax.

PR #3163, opened 2023-07-28
fix: add next-url to NextjsSite header

khuezy:
As of next 13.4.13+, Nextjs has added `next-url` to the vary headers. https://github.com/vercel/next.js/blob/canary/packages/next/src/client/components/app-router-headers.ts#L10


jayair:
Telling Frank

PR #3165, opened 2023-07-29
feat: astro streaming

adamelmore:
This PR adds support for Astro's [streaming feature](https://docs.astro.build/en/guides/server-side-rendering/#streaming) when deployed with the `astro-sst/lambda` adapter. I've tested this against the current app I'm working on and verified that I can defer component loading with these changes.

Example app:
https://github.com/adamelmore/astro-streaming-example

[With streaming (this PR)](https://d35a996kza03bv.cloudfront.net/):

https://github.com/serverless-stack/sst/assets/2363879/cb6985ac-7598-4a07-8bf9-0157cd845f7c

[Without streaming](https://dq1bgzm27fqgl.cloudfront.net/):

https://github.com/serverless-stack/sst/assets/2363879/571cc20c-c95d-46ba-a46d-8731007f293f

### DIFF
--- a/packages/astro-sst/README.md
+++ b/packages/astro-sst/README.md
@@ -39,8 +39,8 @@ If you prefer to install the adapter manually instead, complete the following tw
 
 You can deploy to different targets:
 
-- `edge`: SSR inside a [Lambda@Edge function](https://aws.amazon.com/lambda/).
-- `lambda`: SSR inside a [Lambda function](https://aws.amazon.com/lambda/edge/).
+- `edge`: SSR inside a [Lambda@Edge function](https://aws.amazon.com/lambda/edge/).
+- `lambda`: SSR inside a [Lambda function](https://aws.amazon.com/lambda/) (supports streaming).
 
 You can change where to target by changing the import:
 
--- a/packages/astro-sst/package.json
+++ b/packages/astro-sst/package.json
@@ -39,7 +39,9 @@
     "build": "tsc"
   },
   "dependencies": {
-    "@astrojs/webapi": "^2.1.0"
+    "@astrojs/webapi": "^2.1.0",
+    "@types/set-cookie-parser": "^2.4.3",
+    "set-cookie-parser": "^2.6.0"
   },
   "devDependencies": {
     "@types/aws-lambda": "^8.10.112",
--- a/packages/astro-sst/package.json
+++ b/packages/astro-sst/package.json
@@ -40,11 +40,11 @@
   },
   "dependencies": {
     "@astrojs/webapi": "^2.1.0",
-    "@types/set-cookie-parser": "^2.4.3",
     "set-cookie-parser": "^2.6.0"
   },
   "devDependencies": {
     "@types/aws-lambda": "^8.10.112",
+    "@types/set-cookie-parser": "^2.4.3",
     "astro": "^2.1.3"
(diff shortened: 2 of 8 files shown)


adamelmore:
after playing with this some more, i'm noticing that successful page renders seem to leave the lambda invocation open until it times out. only noticed this after looking at the logs. from the browser, you can't tell anything was still open. still digging.

adamelmore:
seems like something isn't being cleaned up with the stream handling stuff, but i've made a lot of attempts and can't pin it. if any smart people see this and want to review or brainstorm, lmk!

adamelmore:
> seems like something isn't being cleaned up with the stream handling stuff, but i've made a lot of attempts and can't pin it. if any smart people see this and want to review or brainstorm, lmk!

to be clear, everything works as expected from the client's perspective, just an efficiency thing; i'm not sure why the lambda invocation is hanging on until timeout.

adamelmore:
alright, the issue with timeouts is specific to my app; i've updated the description with a basic astro streaming example, and it correctly ends the invocation after all data has been written to the stream.

however, i've not figured out how to compress (gzip/etc) the response stream without causing the output to be delayed until the final data is read from the astro app stream. adding gzip into the pipeline seems to create a situation where the buffer fills up and it kills the whole purpose of streaming (sending incremental chunks of the html to the client). if anyone has any ideas on how to work around this while still compressing the outgoing stream, i'm all ears. noting that cloudfront doesn't seem to be able to do the normal auto-compression when streaming (which makes sense, i think).

fwang:
Hey @adamelmore, can u quickly try out this PR — https://github.com/serverless-stack/sst/pull/3177.

It's mostly cosmetic changes. I created this `event-mapper.ts` helper file that converts CloudFront, APIG v1, and APIG v2 Lambda events into a standard `Request`. Also convert the `Response` back to CloudFront, APIG v1, and APIG v2 Lambda results. We use a similar event mapper for Svelte and OpenNext. I was hoping to standardize it and abstract it out into a standalone lib down the road.

I did some brief tests and it's working. Once you give a green light, I will release it.

adamelmore:
i'm traveling so it'll be pretty late tonight before i'm able to test this, but i looked through it and it looks great!

do you have any expertise around streaming with compression? i really wanted to introduce gzip, and had it working, but i couldn't get it to flush the stream before astro finished sending data. so, it was a gripped stream but it didn't make it to the client until it was all loaded, negating the streaming benefit.

adamelmore:
oops, didn't at frank @fwang

fwang:
The server cache policy for the CloudFront Distribution has gzip enabled — https://github.com/serverless-stack/sst/blob/052328add2cbf6cb4032e6742be6861693adef2e/packages/sst/src/constructs/SsrSite.ts#L1094

Is that relevant to this?

fwang:
@adamelmore The server cache policy for the CloudFront Distribution has gzip enabled — https://github.com/serverless-stack/sst/blob/052328add2cbf6cb4032e6742be6861693adef2e/packages/sst/src/constructs/SsrSite.ts#L1094

Is that relevant to this?

adamelmore:
@fwang unfortunately doesn't seem to affect streamed responses. it does gzip normal responses, but in my testing it doesn't gzip the streamed response.

PR #3167, opened 2023-07-31
Service construct

fwang:


harriha:
Hi,

This has been merged a while ago, but mentioning anyway, perhaps useful for someone: It seems that the `bind` command's `--site` parameter was removed here rather silently. A bit unfortunate that this breaking change wasn't mentioned in the changelog, but luckily an easy thing to fix.

PR #3170, opened 2023-07-31
[DNM] feat: Astro SSR build props transfer to SST deployment process

bayssmekanique:
Currently, SST uses sane but static defaults for some SSR parameters for Astro. This feature allows those sane defaults to be overridden by props from the Astro build process.

This feature should reduce project on-boarding issues as customizations in build output paths are picked up automatically.

This feature also lays the groundwork for an improved SSR CF Distribution by enabling routing logic from the Astro build process to be used in the CF Function.


bayssmekanique:
@thdxr, what are next steps to getting some eyes on this PR and possibly getting it merged?

PR #3171, opened 2023-07-31
Adding docs to overcome GraphQL N+1 problem with AppSyncApi.

raisinbread:
As discussed in the help topic in Discord: https://discordapp.com/channels/983865673656705025/1131711866129874964

Just adding a small snippet to the end of the doc to cover how you overcome N+1 fetching problems using `BatchInvoke`-configured resolver/handlers.

### DIFF
--- a/www/docs/constructs/AppSyncApi.about.md
+++ b/www/docs/constructs/AppSyncApi.about.md
@@ -734,3 +734,52 @@ new AppSyncApi(stack, "GraphqlApi", {
   },
 });
 ```
+
+### Batching GraphQL Resolvers (N+1)
+
+Using naive GraphQL resolvers may result in what the GraphQL community calls the "N+1 Problem", which occurs when the system makes too many upstream requests for a given query.
+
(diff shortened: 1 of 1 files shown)


www/docs/constructs/AppSyncApi.about.md
```diff
@@ -734,3 +734,52 @@ new AppSyncApi(stack, "GraphqlApi", {
   },
 });
 ```
+
+### Batching GraphQL Resolvers (N+1)
+
+Using naive GraphQL resolvers may result in what the GraphQL community calls the "N+1 Problem", which occurs when the system makes too many upstream requests for a given query.
+
+For example, imagine that an an incoming request is asking for 50 `Company` objects, each with its associated list of `Office` objects. Naive GraphQL resolvers will make a request to fetch the 50 `Company` records first, then repeatedly request the `Office` objects for the each of the 50 `Company` records. This results in 51 requests in total.
```
adikari, line 742: There seem to be 2 "an" before "incoming"

raisinbread: @adikari Thanks for the feedback - fixed.

jayair:
Appreciate this! Thanks for the explanation.

PR #3173, opened 2023-08-01
fix: refactor duplicate code; set next-url for edge

khuezy:
I forgot to set `next-url` for edge mode. Refactored duplicate methods to reduce similar mistakes.


khuezy:
Oh yea forgot about that... thanks for clean up!

fwang:
ah nah.. just cosmetic changes... merging

PR #3176, opened 2023-08-02
feat: Astro hybrid support

### LINKED ISSUE #3139: Astro integration is missing basic functionality
The Astro implementation seems to be missing a few key features to make it usable without significant class overrides. By default, Astro's static file generation assumes the server will auto-append `index.html` to directory requests (`/`, `/blog`, `/blog/how-to-foo`).

For reference, I'll use the following distribution behaviors mapping as the example:
image

It's possible to re-introduce some portion of this functionality by utilizing the `defaultRootObject` property on the `cdk` configuration, but this only solves the issues for the root directory (`/`) and leaves the remaining directories unresolvable (`/blog`, `/blog/how-to-foo`).

It's also possible to add a CloudFront Function to the stack as part of the `defaultBehavior`, but this has 2 issues at the moment:
1) The function will be created, but the function built into the Astro integration will be the default behavior so the function added in the config is ignored and requires going into the console and selecting the function to assign to the default behavior. (See https://github.com/serverless-stack/sst/issues/3120)

image

2) The function will only apply to the `Default(*)` behavior so any routes which match other behaviors will not run this function unless they are manually assigned in the console as well. So the root directory (`/`) and the direct children (`/blog`) are supported, but anything further descending will not be resolvable (`/blog/how-to-foo`).

<hr>

I would assume the proper way to resolve this would be to update the integration to include a default function which supports the `index.html` appending similar to how the `SvelteKitSite` constructor includes a query modification.

https://github.com/serverless-stack/sst/blob/9398de7fc9576f5a46372ce1d65756d0c3d5c990/packages/sst/src/constructs/SvelteKitSite.ts#L19-L39

It would also be necessary to update the wildcard behaviors (`/blog/*`, `/about/*`) to also have the function attached.

bayssmekanique:
This PR leverages the Astro build context transfer from a previous PR to enable building a CF Function to perform routing between static resources and a Lambda SSR function.

Resolves #3139


bayssmekanique:
Updated the implementation to use fallback origin for routing between static assets and SSR. This works perfectly for static content and dynamic content using `GET`, `HEAD`, and `OPTION` methods. There are stil 3 remaining issues to resolve:

- [ ] Handle routes which require the use of `PUT`, `POST`, or `DELETE` methods. These will require special behavior declarations which are not difficult to create in SST, but they will require some sort of configuration in the Astro integration site.
- [ ] Handle redirects in the CF Function to avoid having to invoke Lambda function where possible.
- [ ] Handle 404s using CF Distribution error page rather than having to invoke Lambda function to present 404.

packages/sst/src/constructs/SsrSite.ts
```diff
@@ -513,6 +553,47 @@ export abstract class SsrSite extends Construct implements SSTConstruct {
...
+    
+    
+    this._singletonSsrOrigin = new HttpOrigin(Fn.parseDomainName(fnUrl.url), {
+      readTimeout:
+        typeof timeout === "string"
+          ? toCdkDuration(timeout)
+          : CdkDuration.seconds(timeout),
+    });
+    return this._singletonSsrOrigin;
+  }
```
bayssmekanique, line 578: This is a change from how similar resources have been instantiated, but resources like `S3Origin` are the same independent of Edge or Regional deployment. That's not the case for this `HttpOrigin` since it only exists for Regional deployment. This method allows for essentially JIT creation of resources once they are utilized the first time, and then references the singleton for all future requests.

packages/sst/src/constructs/SsrSite.ts
```diff
@@ -513,6 +553,47 @@ export abstract class SsrSite extends Construct implements SSTConstruct {
...
+
+  /** @internal */
+  protected get cachePolicy() {
+    if (this._singletonCachePolicy) return this._singletonCachePolicy;
+    this._singletonCachePolicy = this.buildServerCachePolicy();
```
bayssmekanique, line 583: The logic for `buildServerCachePolicy` should probably be moved into this getter if this methodology is acceptable.

bayssmekanique:
@jayair, this is a potential fix for behavior limitations as well. Any chance of getting someone to take a look or merge?

adamelmore:
my man...

it's a thing of beauty

fwang:
Thanks @bayssmekanique! Love the idea of having `AstroSite` be more aware of the Astro config.

I wanted to hear ur thought on two CF origin group setups, and how each of the following 3 types of requests are handled in each setup:
1. Static requests not cached by CF
2. Static requests cached by CF
3. SSR/API requests

### Setup 1: Origin group S3 falling back to Lambda
Request 1 will be served directly by S3 => same latency as current setup.
Request 2 will be served directly by CF cache => same latency as current setup.
Request 3 will hit S3 first, then fall back to Lambda => **SLOWER than current setup.**

### Setup 2: Origin group Lambda falling back to S3
Request 1 will hit Lambda first, then fall back to S3 => **SLOWER than current setup.**
Request 2 will be served directly by CF cache => same latency as current setup.
Request 3 will be served directly by Lambda => same latency as current setup.

In Setup 1, SSR/API requests will always see a penalty because requests are not cached. Every request will hit S3 first, then fall back.
In Setup 2, static requests will see a penalty when uncached. Once cached, they will be served by CF cache.

How should we weigh the tradeoff for users?

/cc @adamelmore @thdxr

bayssmekanique:
@fwang, those are great questions. As a bit of pretext to this discussion, the impetus for this PR was largely driven out of a desire to improve developer on-boarding (ie: having routes act the same in SST deployed environment as they do in the dev environment) and allow for more efficient deployment strategies across all frameworks (though I've only really addressed Astro in this PR).

I'd like to expand the 3 scenarios you outlined because think I don't think it's fair to assume that SSR route can't/won't be cached. It's an area that I do think current SSR frameworks don't have a great handle on (maybe because it's assumed to be the implementers responsibility), but I think most experiences utilizing SSR can still benefit from selectively caching the SSR responses by utilizing the response headers. I have middleware in my Astro projects which add cache headers to most responses that aren't sensitive or extremely latency dependent. Combining even relatively short max-age values of 5 minutes can have huge performance improvements when combined with `rel="prefetch-intent"` to essentially prefetch and cache the route before the user navigates. For that reason, I think it's worth adding "4. SSR/API cached by CF".

1. Static requests not cached by CF
2. Static requests cached by CF
3. SSR/API requests not cached by CF
4. SSR/API request cached by CF

I'm also going to add some quantitative values to my thoughts, but please note that these times were what I experienced on my computer doing naïve testing with my test deployment so these are by no means rigorously vetted.

Also worth noting, the difference between Request 1 and Request 2 is imperceivable. The difference between a request with a `Miss from cloudfront` (meaning served by S3) and a `Hit from cloudfront` for S3 content seems to be virtually non-existent in my testing. The "_Waiting for server response_" duration on both request types was consistently ranging from 28ms to 32ms, with misses sometimes having shorter delays than CF hits.

<hr />

### Setup 1: Origin group S3 falling back to Lambda (this PR)
- Request 1 will miss on CF, and hit on S3 (~35ms)
- Request 2 will hit on CF (~35ms)
- Request 3 will miss on CF, miss on S3, hit on Lambda (~1.3s-2s cold, ~200ms-500ms warm)
- Request 4 will hit on CF (~35ms)

Obviously the worst case scenario for this configuration is a non-cached SSR request and I understand the concern is that the S3 miss might be adding overhead to those requests, so I tested this (again, naively). I duplicated the `/blog` route on my test site to `/shmog` and added a behavior to bypass the S3 fallback and directly serve the `/shmog` route from Lambda and just as with the static routes, the difference was imperceptible. There is significant variation request to request for both routes and they both had very similar limits (150ms-260ms) but without collecting a much larger sample size I can't draw any reliable conclusions other than to say that if there is a delay from the miss it is likely in the low single digit ms range which is largely insignificant on a request in the several hundred ms range.

I don't think the difference in the response times warrants a developer's time to optimize away this microscopic difference, but should they desire to, adding routes to the `ssrExlusiveRoutes` parameter would bypass the S3 miss and directly hit Lambda on those requests. There can be an argument made that subjecting SSR/API routes to the same behavior limit quota that was plaguing static routes in the current implementation is putting developers in the same place, but ultimately the difference is that the limit on static content was potentially making content unreachable, while this limitation is limiting only how many routes can be optimized to avoid the S3 hit (or enable `POST`/`PUT` methods).

### Setup 2: Origin group Lambda falling back to S3
- Request 1 will miss on CF, miss on Lambda, and hit on S3 (~1.3-2s cold, ~200-500ms warm)
- Request 2 will hit on CF (~35ms)
- Request 3 will miss on CF, hit on Lambda (~1.3-2s cold, ~200-500ms warm)
- Request 4 will hit on CF (~35ms)

I understand the thinking behind this configuration because once the cache is primed SSR responses theoretically won't have the S3 miss penalty (~1-3ms). The downside is that all static responses have a potential Lambda response penalty (~100-120ms) on misses from the CF cache. Even if the cache is primed by a crawler after deployment, CF doesn't guarantee to hold the content in the cache and will regularly purge infrequently accessed files. Given that most sites make significantly more requests for static content than for potentially dynamic, the risk of one of the 10+ static files having a miss from CF and causing a 100ms delay in rendering doesn't seem like it's worth the much smaller guaranteed miss from S3. As a developer I'm drawn to this solution since it presents the lowest possible response time of the two, but as a product manager I place my support with the first implementation because it's more consistently performant.

A much less broadly applicable limitation to this approach that still bears consideration is that this removes the ability of the Lambda function to handle misses. For a lot of deployments using the standard 404 handling in CF will be fine, but for larger sites 404 pages will often provide links to related content based on the contents of the missed page's URI, which won't be possible with this approach.

<hr />

I'm entirely biased in this discussion as the approach I implemented feels more *natural* to me as someone who tends to favor pre-rendering as much as possible and using SSR in only strictly necessary scenarios so I would invite others to please take part in this dialogue and point out any blind spots I have for either implementation approach.

The difference between the two is literally a matter of switching two lines of code, so it wouldn't be inconceivable to give the developer the ability to configure which approach to deploy, but that does add the requirement of additional documentation and could make future development more difficult.

In all honesty, I don't think anyone outside the four people in this discussion would ever even know the S3 miss penalty exists on those requests, but I do think it's a valid concern and worth investigating further. Are there contacts in the AWS channel that could find out from the Origin Failover engineers what the miss penalty actually is?

adamelmore:
My vote is s3 -> lambda, and that's coming from someone with a largely SSR driven site!

bayssmekanique:
@fwang & @thdxr, what are your thoughts on the above? I can work on merging the recent `master` changes if this looks like it's viable to merge soon.

bayssmekanique:
@fwang, I resolved the conflicts and merged the changes from `master`.

bayssmekanique:
While investigating `Astro v3` support I uncovered an issue with this PR. I assumed that the `build` process happened before the `AstroSite` constructor was called but that's not the case, so the imported build props are actually stale values from the previous build.

bayssmekanique:
I don't see a way to implement this using the CLI binding method. This implementation requires values that are not known to the system until build time and `SsrSite` construct doesn't build the app until very late in the process.

I feel like the proper approach is to drop the CLI binding and instead use the build process hooks to run the SST based on the hooks, but that would require some significant architecture changes to how the constructs work.

I see that other integrations get around this by reading and parsing the config files of the package directly, but that feels like a huge dependency nightmare as it requires reimplementing the config parsing for that package.

fwang:
@bayssmekanique sorry I've been dragging my feet on this PR.

> Setup 1: Origin group S3 falling back to Lambda (this PR)

From the benchmark, this benchmark makes a lot of sense. My hesitation is users having to manually specify `ssrExclusiveRoutes`. It takes the magical "works out of the box" feeling. Is there a way we can figure that out automatically?

> I assumed that the build process happened before the AstroSite constructor was called but that's not the case, so the imported build props are actually stale values from the previous build.

Build has to happen after constructor. But we can lazily add the behaviors to the distribution AFTER build, at which time we know `sst.build-props.json`.

#### Next step

If I make the change to add behaviors after the build, we can merge the CF function rewrite and redirect code from this PR (while we decide on the origin group), right?

bayssmekanique:
@fwang, regarding the `ssrExclusiveRoutes`, in theory the routes market as `endpoint` could be considered to be SSR exclusive, the only problem is that with a hard limit on how many behaviors there are it could quickly run into problems because it's naively assigning behaviors to endpoints which overlap or don't actually need the exclusion. It's really only `PUT` and `POST` methods which need to be added in order to function correctly, but Astro doesn't define the action types in a way it can be determined ahead of time.

IMO these are a bit beyond the "happy path" for most Astro content sites so I can see it being something that requires configuration, but I also agree that it would nice if it just worked for most users that don't have a ton of endpoints. Maybe the mid-point is auto-assigning behaviors to endpoints but having a configuration parameter to essentially eject that feature and custom overriding. Perhaps a name change would also help a bit as `ssrExclusiveRoutes` is very accurate but also a bit confusing for anyone that doesn't understand the limitation. Maybe `ActionRoutes` or `FormRoutes`?

<hr>
As for the configuration issue, there are several configurations that I was pulling from the `sst.build-props.json` file to assist with easing DX which would be required before the Construct runs, but those were certainly more "nice to haves" than strict feature requirements. 

```json
  "props": {
    "customDomain": "preview.zachcardoza.com"
  },
  "buildConfig": {
    "typesPath": "src",
    "serverBuildOutputFile": "dist\\server\\entry.mjs",
    "clientBuildOutputDir": "dist\\client",
    "clientBuildVersionedSubDir": "static"
  },
  ```

Parsing the `astro.config` file doesn't seem like a wise choice either as there are too many run-time configurations which are impossible to predict.

fwang:
>  Maybe the mid-point is auto-assigning behaviors to endpoints but having a configuration parameter to essentially eject that feature and custom overriding.

Sounds good

> As for the configuration issue, there are several configurations that I was pulling from the `sst.build-props.json` file to assist with easing DX which would be required before the Construct runs

Just to clarify, you info u r pulling out from `sst.build-props.json` only affects the CF function and the cache behavior patterns right? If we added the CF function and the cache behaviors AFTER the Astro app build, would that work?

bayssmekanique:
@fwang, I've moved most of the build process inside the `SsrSite` construct into the deferred task and it works. The issue is that calls to the `getConstructMetadata` rely on the `url` and `customDomainUrl` getters which rely on `this.distrubution` which isn't created at the time of first call so the meta object ends up with `undefined` for both attributes. The full site deployment seems to still work correctly, but the CLI output is missing the URL and I'm sure there are likely some secondary effects from this as well.

I'm not certain if there's a better way to get the `url` and `customDomainUrl` values which still support overloading the Distrubution?

bayssmekanique:
Ultimately, the extensible nature of SST has resulted in a very linear build process which is preventing it from using the build process to create an optimized deployment. This PR has reached the end of it's limits without significant changes to the nature of SST constructs which is far beyond the scope of this PR or something I'm able to tackle independently.

It would be ideal if the SST constructs used event hooks to build the deployment in phases. That would allow a `pre-build` hook where configuration parameters could be shared from the SST env into the build environment, then a `post-build` phase where final configurations are returned from the build process and the deployment could proceed using the already built project. I'm sure I'm naively reducing the complexity of the steps involved, but that's what I foresee as being an ideal flow for a tool like Astro or Qwik.

jayair:
I'll share this internally. Thanks for the feedback!

fwang:
@bayssmekanique I still think this is doable. I was planning to create the the distribution inside the constructor, so `url` is available. And add the default and additional cache behaviors in deferred task.

I'll give it a go this week.

Let's keep the PR open. I think it's a pretty key feature.

bayssmekanique:
@fwang, that sounds like it would work. I'd still recommend starting a new branch. There were some significant changes to the SSR construct in the last couple of commits that were really hard to resolve so I think it would be better to reimplement the necessary changes from `main`.

fwang:
@bayssmekanique @adamelmore quick update, I've been working on refactoring `SsrSite` to allow building the distribution AFTER the site is built. And then we can move forward w/ this PR (or a new PR).

Should have something today/tmr. Will update soon.

bayssmekanique:
Once you have the updates merged to `master` I'd be happy to reimplement the core of this PR in a new PR to avoid having merge conflict nightmare!

fwang:
@bayssmekanique I standardized the interface for SSR sites in v2.26.3. Take the [`AstroSite`](https://github.com/sst/sst/blob/master/packages/sst/src/constructs/AstroSite.ts#L26) as an example, it tells `SsrSite` how it should be deployed including:
- CloudFront functions
- Edge functions
- Origins
- Behaviors, including the CF function, edge function, and origin each behavior uses

The goal is to let `AstroSite` own its infrastructure.

Let me know if this interface is flexible enough to achieve our proposed setup above.
^ We would probably need to define a new `s3-to-function` origin type (currently only `s3` and `function`)

fwang:
Lemme loop in @thdxr. Are we going to move the `sst-astro` plugin to it's own repo?

PR #3180, opened 2023-08-03
fix: respect newline char when writing package.json

oyed:
RE: https://discord.com/channels/983865673656705025/1007139657324187729/1135930483851673673

Just a small change to respect the newline chars when overwriting `package.json`s during `sst update` - works when there is a newline and when there isn't one


fwang:
Thanks @oyed!

PR #3186, opened 2023-08-04
Examples: added pnpm into the installation instructions

JPVan:


jayair:
Whoa that's a lot of files. The examples are generated by the presets, so we don't need those changes. Just have the preset ones.

But also I noticed some changes like these https://github.com/serverless-stack/sst/pull/3186/files#diff-beef1a7fe3b5cd540729c86f0773a4fdd228026069355965159e3fff9c349408. Was that intended?

JPVan:
Yeah it's a lot of files, lol. 

I did intentionally change both the presets and the examples because the examples seem to lag the presets and don't seem to always get generated on each build / release. But whatever you think is best. 

The other changes you noticed were intentional, but not intended as part of this same PR. I'm still learning about PRs on GitHub, and would love any link or hint you have to share about how I can separate pull requests. I'm guessing I did something wrong, but don't know what. X) 

On the other hand, if you think the other files are a good change, you could accept them as part of this PR. Again, your call obviously.

JPVan:
This is slightly off topic, but I've also noticed the examples seem to break without getting noticed right away. Would it be worth adding in the ability to build all the examples from a single command that also checks to see if anything breaks, perhaps as part of the example generation process? Bonus points for a solution that can easily be run locally as well as by CI.

JPVan:
I'd be willing to give that a try if you think it'd be useful.

JPVan:
Ok, should be good now.

JPVan:
Sorry, that was messy. Should be good now.

JPVan:
@jayair Thanks. Anything else you'd like me to change?

fwang:
Thanks @JPVan! Looks good 🙏

PR #3190, opened 2023-08-06
fix: docs for Service Dockerfile

tyrauber:
Small issue with the Service docs related to the Dockerfile:
Need to be in the WORKDIR where package.json is to run rpm install.


fwang:
Ah nice catch!

PR #3194, opened 2023-08-07
feat(sst/rds): add additional rds versions

nurulhudaapon:
- feat(sst): add support for additional rds engine versions
- doc(sst/rds): add additional engine version


nurulhudaapon:
For serverless mode not all versions are supported.

PR #3207, opened 2023-08-08
Upgrade fast-jwt

oleghind:
SST fails to install on node 20.5.0 due to outdated package 
 `error fast-jwt@1.7.2: The engine "node" is incompatible with this module. Expected version ">=14 <20". Got "20.5.0"`

### DIFF
--- a/packages/sst/package.json
+++ b/packages/sst/package.json
@@ -80,7 +80,7 @@
     "dotenv": "^16.0.3",
     "esbuild": "0.18.13",
     "express": "^4.18.2",
-    "fast-jwt": "^1.6.1",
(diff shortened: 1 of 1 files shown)


benjamine:
@hannesj nice catch! I checked the 4 places where that is used, and it doesn't look like a function is passed?

https://github.com/search?q=repo%3Asst%2Fsst%20createVerifier&type=code

eg
```
    const jwt = createVerifier({
      algorithms: ["RS512"],
      key: getPublicKey(),
    })(token);
```

it looks like here `getPublicKey()` is called, but `key` is still getting a string (the return value of `getPublicKey`)

PR #3211, opened 2023-08-09
vendor: upgrade constructs library to 10.2.69

berenddeboer:
Just a bump, as this seems to be hanging at an older version.


jayair:
Yup taking a look

fwang:
Thanks @berenddeboer! Closed in favour of #3217

PR #3216, opened 2023-08-11
feat: astro streaming compression

adamelmore:
The actual changes in this PR feel kind of sloppy, tbh, but these changes add compression to streamed responses. The only other issue that I should call out is it's not as efficient in terms of the compression factor as a normal, non-streamed response (because we're compressing each chunk). I think the solution to increase the compression factor would be adding some state and compressing chunks in batches, but that would add some complexity and work against the benefit of streaming in the first place.

### DIFF
--- a/.changeset/calm-suits-breathe.md
+++ b/.changeset/calm-suits-breathe.md
@@ -1,0 +1,5 @@
+---
+"astro-sst": patch
+---
+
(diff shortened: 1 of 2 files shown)


adamelmore:
@fwang any thoughts on this PR?

Issue #3189, opened 2023-08-06
Add Python 3.11 runtime


adolfogc:
Hi, AWS Lambda now supports Python 3.11 (https://aws.amazon.com/blogs/compute/python-3-11-runtime-now-available-in-aws-lambda/).

jayair:
Do you have the CDK version that has this?

adolfogc:
Hi, based on this commit:

https://github.com/aws/aws-cdk/commit/e2b7791072e187f441eb72a5f6d0462748ec5112

I think it is available starting from version v2.88.0

adolfogc:
This PR would allow adding the Python 3.11 runtime: https://github.com/sst/sst/pull/3209

jayair:
Yup taking a look

adolfogc:
Hi @jayair, with https://github.com/sst/sst/pull/3217 now merged, I have created a PR to add the Python 3.11 runtime.

PR #3222, opened 2023-08-14
Add Python 3.11 runtime

adolfogc:
Related issue: https://github.com/sst/sst/issues/3189


adolfogc:
@jayair what do you think? can this be merged?

PR #3227, opened 2023-08-15
feat: cli command for generating types only

jamesgibbons92:


bayssmekanique:
This feature would be very useful!!!

fwang:
Renamed the command to `sst types`

Issue #3213, opened 2023-08-09
RemixSite types for cdk.server.copyFiles


dani-mp:
It works, but TypeScript complains.

<img width="221" alt="image" src="https://github.com/sst/sst/assets/2881251/ea6a27f3-5749-4877-b791-cd5351349dae">

jayair:
What's the error?

dani-mp:
> What's the error?

Just the types of that property. It doesn't expose `copyFiles` there.

jayair:
Do you want to take a crack at this one with a PR? If not then can you open a #help thread in Discord and I'll get the team to take a look.

dani-mp:
@jayair there you go! https://github.com/sst/sst/pull/3242

fwang:
Thanks @dani-mp! `copyFiles` is currently being used internally for `SvelteKitSite`. Can u share more about ur use case?

dani-mp:
@fwang I'm using it in a RemixSite. I need to make sure some files related to Prisma (which I'm using for my db) are included in the build folder of the server lambda. Someone in the sst discord told me about using copy files and it worked, although the types complained.

I also promised to create another PR to explain how to make Prisma work because apparently a lot of folks struggle with that as well. It wasn't straightforward for either.

dani-mp:
What do you say, folks? Should we move forward with the attached PR?

jayair:
Sorry for the delay. Did you mean this PR? https://github.com/sst/sst/pull/3242

I'll ask Frank to take a look.

dani-mp:
@jayair thanks!

PR #3242, opened 2023-08-20
Add copyFiles to RemixSite cdk server

dani-mp:
Fixes https://github.com/sst/sst/issues/3213.


fwang:
Merged. Sorry about the delay @dani-mp

PR #3246, opened 2023-08-21
Service: Add node client and docs

omikader:
This should address some [type errors](https://discord.com/channels/983865673656705025/1142284213207437453/1142284213207437453) that occur when a `Service` construct is introduced to the app


fwang:
Thanks @omikader! Much much appreciated.

PR #3252, opened 2023-08-23
Add dev.command option to SsrSite

BSteffaniak:
Hello, I noticed that the 'Start {framework}' output seems to be hardcoded to an optional cdCmd (based off of the props.path) and then a hardcoded 'npm run dev' string. In my situation the generated 'cd ../ && npm run dev' was not correct. I have the sst folder in a subdirectory for this specific repo, but both the 'npx sst dev' and 'npm run dev' should be executed in the subdirectory, not one in the subdirectory and one in the parent as the generated message suggests to do.

I think this solution of adding an optional devCommand prop solves this specific case, but also would enable users who might not be using 'npm run dev' to run their dev command to customize that.

### DIFF
--- a/packages/sst/src/cli/commands/dev.tsx
+++ b/packages/sst/src/cli/commands/dev.tsx
@@ -307,7 +307,7 @@ export const dev = (program: Program) =>
                         ? ""
                         : `cd ${props.path} && `;
                     const devCommand =
-                      props.devCommand || `${cdCmd}npm run dev`;
+                      props.dev?.command || `${cdCmd}npm run dev`;
                     Colors.line(
                       Colors.primary(`➜ `),
                       Colors.bold(`Start ${framework}:`),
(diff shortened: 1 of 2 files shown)


BSteffaniak:
Here is the current output in 2.24.12:
image

With the change I am proposing it would be configurable to something like this:
image

Which after I move the prop to inside dev could be set up to look like this (using NextjsSite as an example):
```
new NextjsSite(stack, 'bob', {
    path: '../',
    dev: {
        command: 'npm run dev',
    },
});
```

This would also allow for defining the command for other package managers manually (e.g. yarn, etc).

I'll move that prop to inside the dev prop now. 👍

BSteffaniak:
By default, the `cd ${path}` part will still remain.

This PR adds the ability to override the default behavior of adding that. The scenario I described about my current use case for wanting to remove the `cd ${path}` was the motivation for this PR, but I think this could also be useful for the aforementioned reasons (if you use a different package manager, need to run a chain of commands, etc)

jayair:
Yeah let me ping the team about it.

thdxr:

Issue #3195, opened 2023-08-07
SST docs should be more clear on pointing root external domains to cloudfront without route 53


JoltCode:
The docs show the following regarding not using route 53 with custom domains:
![image](https://github.com/sst/sst/assets/46378904/735b1b13-48c2-4602-83e8-180f85b68ec5)

I think the info panel should be changed to add more detail surrounding using CNAME's for root domains, as certain providers (such as cloudflare), do allow this (although through flattened CNAME's). Alternatively, we could modify the current notice to make it clear that it is possible, but it depends on your DNS host. 

---

~~If you've already posted your issue on Discord, make sure to leave a link to it here.~~ N/A - can do if needed :smile:

jayair:
Good point. Would appreciate a quick PR!

JoltCode:
@jayair sorry didn't get on this sooner, here you go! ^

jayair:
Appreciate it!

PR #3256, opened 2023-08-25
Update docs on pointing CNAME's to root domains to CloudFront

JoltCode:
Provides more detail on how to point CNAME's (for root domains) to CloudFront, as the docs currently suggest that this cannot be done.


JoltCode:
Fixes #3195

jayair:
Thanks!

PR #3266, opened 2023-08-29
Fix: Update nodejs 14 -> 16 for custom resource support functions

jamesgibbons92:
In the past few days AWS have scheduled EOL for Lambda NodeJS 14 and sent out the following message to anyone using that version:

```
We are contacting you as we have identified that your AWS Account currently has one or more Lambda functions using the Node.js 14 runtime.
We are ending support for Node.js 14 in AWS Lambda. This follows Node.js 14 End-Of-Life (EOL) reached on April 30, 2023 [1].
```

CDK have updated most of their CR now to use at least node16. So the next CDK bump should pull in those updates automatically, there are a couple of SST custom resources that use node14 however so this PR just bumps them to 16.

I considered node18 but would require a re-write of dns validator function as it doesn't bundle the aws-sdk deps, and node18 lambda does not include aws-sdk v2 in the runtime


fwang:
Thanks @jamesgibbons92 @bobbypiper

bobbypiper:
@jamesgibbons92 @fwang 

https://nodejs.org/en/blog/announcements/nodejs16-eol

LOL!

jayair:
lol

ssrowe:
Are there any plans to update SST to use nodejs18.x for customer resources and handlers?

This-Is-Not-Brad:
Yes, AWS is pinging us about Node 16 EOL, for the Custom Resource Handler and Certificate Request Functions

DianomiJH:
I've been getting Pinged for the CustomS3AutoDeleteObjectsCustomResourceHandler

jayair:
Are these related to the SST Bootstrap stack btw?

DianomiJH:
Yes the Auto Delete Objects one i reported is SST Bootstrap

PR #3271, opened 2023-08-30
RemixConstruct: Encode special characters in the url path

### LINKED ISSUE #3206: 403 response for some  page resources when using Remix
### Problem
I run into an issue where some of my js bundle for a specific page returns `403` upon request, I deployed the same app on vercel but doesn't seem like it's an issue on there. I'm using the [remix-flat-routes ](https://github.com/kiliman/remix-flat-routes/issues) package and using the [nested folders](https://github.com/kiliman/remix-flat-routes#nested-folders-with-flat-files-convention--new-in-v051) which let's you define a folder as a flat route by utilizing `folder_name+/routes.ts` upon inspecting the network calls
it seems only resources with the `+` fail with a `403` the file does exist on s3, but I suspect because `+` is a special character in terms of URL it is being parsed to a space which then fails to find the file.
 Response
CleanShot 2023-08-08 at 10 31 32@2x

### Reproduction
Create a remix app, and set up the remix-flat-routes package. Add a nested folder route or two and deploy

barakcodes:
Fixes #3206
- Encode path to allow  having special characters in files
- Attach Cloudfront functions to all distributions

### DIFF
--- a/packages/sst/src/constructs/SsrSite.ts
+++ b/packages/sst/src/constructs/SsrSite.ts
@@ -1113,6 +1113,9 @@ function handler(event) {
           compress: true,
           cachePolicy: CachePolicy.CACHING_OPTIMIZED,
           responseHeadersPolicy: cdk?.responseHeadersPolicy,
+          functionAssociations: [
+            ...this.buildBehaviorFunctionAssociations(),
+         ]
         }
       );
     }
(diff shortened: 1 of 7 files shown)


packages/sst/src/constructs/SsrSite.ts
```diff
@@ -1113,6 +1113,9 @@ function handler(event) {
           compress: true,
           cachePolicy: CachePolicy.CACHING_OPTIMIZED,
           responseHeadersPolicy: cdk?.responseHeadersPolicy,
+          functionAssociations: [
```
barakcodes, line 1116: This seems to work for now but  I'm not sure if passing a function to all behaviors is a good idea

fwang: Thanks @barakcodes! I added a `clientCFFunctionInjection` field and moved the code in there.

PR #3277, opened 2023-09-01
Add healthcheck props to Service construct

alistairstead:
Added configurable health check to the target group created within the Service construct. Further to the discussion on [Discord](https://discord.com/channels/983865673656705025/991751731824308344/1146893038027817000)

The default health check created assumes that the root path of the service task returns an HTTP 200 response. When deploying 3rd party images or applying security to the service this may not be the case. If the default health check fails the deployment of the Service times out as the deployment internally for CF/Fargate does not complete. This change retains all the default configuration but enables any of the [HealthCheck](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_elasticloadbalancingv2.HealthCheck.html) props to be set. 

I considered placing this configuration inside the existing `cdk` props but concluded that this is actually a config that refers to the Service you are defining.

### DIFF
--- a/.changeset/smart-oranges-change.md
+++ b/.changeset/smart-oranges-change.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Container: support customizing ALB target group (ie. health check)
--- a/packages/sst/src/constructs/Service.ts
+++ b/packages/sst/src/constructs/Service.ts
@@ -66,7 +66,7 @@ import { Platform } from "aws-cdk-lib/aws-ecr-assets";
 import {
   ApplicationLoadBalancer,
   ApplicationTargetGroup,
-  HealthCheck,
+  ApplicationTargetGroupProps,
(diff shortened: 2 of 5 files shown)


nguyendon:
LGTM - are there any relevant tests for this?

alistairstead:
Timely reminder @nguyendon thanks. Tests added.

nguyendon (approved):

fwang:
@alistairstead I added an `cdk.applicationLoadBalancerTargetGroup` prop to allow customizing the ALB target group. You can pass in `healthCheck` in there.

To add some context, we wanted to keep anything that require a CDK import go into `cdk.X`. The thinking is you don't need to import CDK unless u want to customize SST constructs.

And if there's something a lot ppl need to customize, we lift the property out from `cdk.X` to top level. And make it not require CDK imports.

Hope that makes sense.

PR #3283, opened 2023-09-05
feat(python-bundling): allow for excluding files and folders from bundle

### LINKED ISSUE #720: Add `exclude` option for Function to exclude certain directories
When creating a [Function](https://docs.serverless-stack.com/constructs/Function#configure-bundling-a-python-function), there is no option to exclude directories from the `srcPath`. This `exclude` should be able to interact with CDK's [`AssetOption.exclude`](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-s3-assets.AssetOptions.html#exclude) directly

One usage of this would be for python sources that have their `.venv` in the same directory as the `Pipfile`.

Without the `exclude` of `.venv`, the `.venv` is bundled together with the lambda, including the packages installed by the dependency builder on the docker, effectively copying the packages twice. This bloats the python runtime lambda code and can lead to failed deployments due to exceeding the [250MB Deployment package size](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html)

sladg:
Fixes #720 for Python runtime.

### DIFF
--- a/packages/sst/src/constructs/Function.ts
+++ b/packages/sst/src/constructs/Function.ts
@@ -528,6 +528,29 @@ export interface PythonProps {
    * ```
    */
   installCommands?: string[];
+
+   /**
+   * A list of paths to exclude from bundle. Checks if exclusion exists in absolute path string.
+   *
+   * Each path in the array is a command that'll be run. For example:
+   *
+   * @default "[]"
+   *
+   * @example
+   * ```js
+   * new Function(stack, "Function", {
+   *   python: {
+   *     exclude: [
+   *       '/.sst/',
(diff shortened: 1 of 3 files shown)


jayair:
Thanks @sladg!

sladg:
@fwang yeah! that sounds like a reasonable thing, we good with adding `glob` to our dependencies?

fwang:
@sladg Yeeeah! Let's do it!

CestDiego:
;O Thanks for this. Just what I needed. Are there other ways we can exclude compiled cache files?

thdxr:

PR #3285, opened 2023-09-05
SsrSite: Upgrade s3 handler python versions

omikader:
https://github.com/sst/sst/pull/3266 upgrades the remaining nodejs 14 resource functions to nodejs 16 in response to the deprecation warning from AWS. This PR does the same thing but for the Python3.7 deprecation warning

```
We are contacting you as we have identified that your AWS Account currently has one or more Lambda functions using the Python 3.7 runtime.

We are ending support for Python 3.7 in AWS Lambda. This follows Python 3.7 End-Of-Life (EOL) reached on June 27, 2023 [1].
```

The email suggests upgrading to 3.11 which I've done here -- I believe it should be backwards compatible. Also I removed some unused imports I noticed in `s3-upload.py`

### DIFF
--- a/.changeset/forty-ties-retire.md
+++ b/.changeset/forty-ties-retire.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+SsrSite: Upgrade s3 handler python versions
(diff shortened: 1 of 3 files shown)


jayair:
Thanks!

omikader:
Closes https://github.com/sst/sst/issues/3289

PR #3295, opened 2023-09-12
NextJsSite: add support for open-next 2.2

conico974:
This PR aims to add support for revalidate tags of open-next in NextJsSite.
It also add experimental support for streaming.


khuezy (approved):
LGTM!

packages/sst/src/constructs/NextjsSite.ts
```diff
@@ -271,6 +278,64 @@ export class NextjsSite extends SsrSite {
...
+
+    if(fs.existsSync(dynamodbProviderPath)) {
+      const insertFn = new CdkFunction(this, "RevalidationInsertFunction", {
+        description: "Next.js revalidation data insert",
+        handler: "index.handler",
+        code: Code.fromAsset(
+          path.join(this.props.path, ".open-next", "dynamodb-provider")
+        ),
+        runtime: Runtime.NODEJS_18_X,
+        timeout: CdkDuration.minutes(14),
```
khuezy, line 311: Any particular reason why this is 14 minutes?

conico974: No, it's just a typo

packages/sst/src/constructs/NextjsSite.ts
```diff
@@ -271,6 +278,64 @@ export class NextjsSite extends SsrSite {
...
+      partitionKey: { name: "path", type: AttributeType.STRING },
+      sortKey: { name: "revalidatedAt", type: AttributeType.NUMBER },
+    });
+
+    server?.addEnvironment("CACHE_DYNAMO_TABLE", table.tableName);
+    table.grantReadWriteData(server?.role!);
+
+    const dynamodbProviderPath = path.join(this.props.path, ".open-next", "dynamodb-provider");
+
+    if(fs.existsSync(dynamodbProviderPath)) {
```
khuezy, line 303: Should the table created above be moved inside the `if` block? That is, we don't want to create the DDB Table if the app does not store any cache.

khuezy (changes requested):
@fwang we may want to make some code changes and allow users to opt in DDB creation... waiting for conico to respond when he wakes up.

packages/sst/src/constructs/NextjsSite.ts
khuezy: What's cheaper, storing the caches in DDB or in S3?

fwang:
Made a few trivial changes:
- Use `TableV2` instead of `Table` construct. The former is recommended by CDK and supports global table if we ever need it down the road.
- Set removal policy for the table to DESTROY. We want to remove the table if user decide to remove the `NextjsSite`. Default removal policy if RETAIN.
- Moved `props.experimentalStreaming` to `props.regional.experimentalStreaming` since streaming only applies to the `regional` mode.

Questions:
- Curious why the table name is hardcoded?
  > app.logicalPrefixedName("next-revalidation-table")
- Should we wrap `new Table` inside `if (fs.existsSync(dynamodbProviderPath))`?

packages/sst/src/constructs/NextjsSite.ts
fwang: S3 much cheaper. But we can't look up by cache if store in S3, can we?

conico974:
> Questions:
> 
> * Curious why the table name is hardcoded?
>   > app.logicalPrefixedName("next-revalidation-table")
> * Should we wrap `new Table` inside `if (fs.existsSync(dynamodbProviderPath))`?

We can remove the hardcoded name, it was just for testing so that i don't end up with dozens of tables and i just forgot about it.

We can't wrap the new table under the if, people could add tags during runtime and we'd need to handle those cases. But we could add a flag so that people can disable DDB if they don't want to use on demand revalidation.

I'll write a PR with some flags on open-next too to disable DDB and the incremental cache

packages/sst/src/constructs/NextjsSite.ts
conico974: S3 is indeed cheaper, but there is no good way to make tags revalidation works within lambda. 
We could use the same method that they use on standalone of a single json files responsible for all the tags, but it will just work really really bad on lambda.

fwang:
Removed the hardcoded table name.

Should we release this or wait for the disable DDB option for OpenNext?

khuezy:
I think so, let me test and release 2.2.1 before we release this. I think most apps won't be needing the `invalidateTag` or `invalidatePath` in app router if they can achieve the same thing w/ the `pages/api`' `res.revalidate`... so reducing the resources is ideal

packages/sst/src/constructs/NextjsSite.ts
khuezy: I recall discussion EFS. Google has articles stating that it doesn't add coldstarts, but some people say it does... The downside is that it has to be in a VPC and doesn't work on the edge.

packages/sst/src/constructs/NextjsSite.ts
```diff
@@ -111,14 +136,21 @@ export class NextjsSite extends SsrSite {
...
       ...props,
     });
 
-    this.createRevalidationQueue();
-    this.createRevalidationTable();
+    
+    if (!props?.dangerous?.disableIncrementalCache) {
+      this.createRevalidationQueue();
+    }
+    if (!props?.dangerous?.disableDynamoDbCache || !props?.dangerous?.disableIncrementalCache) {
```
khuezy, line 151: If I only want to disable DDB cache, but have Incremental Cache working, do I need to create the DDB table?

conico974: No this is not necessary, it only means that `next/cache` won't work

khuezy: I may be confused on the second or statement. Because if I use `disableDynamoDbCache: true` and leave `disableIncrementalCache` as false, this will create the DDB table?

conico974: You're right, i need to be more careful with copilot sometimes. It's missing parenthesis

khuezy (approved):

fwang:
Made some naming changes:
- `regional.experimentalStreaming` => `experimental.streaming` (default false)
- `dangerous.disableDynamoCache` => `experimental.enableOnDemandRevalidation` (default true)
- `dangerous.disableIncrementalCache` => `experimental.enableIncrementalCache` (default true)

The thinking is that.. for flags that we don't expect most users to touch (and those we are still exploring), let's put them all under `experimental`.

Also wanted to run it across you guys @conico974 @khuezy, does naming the prop `enableOnDemandRevalidation` make sense? If the above makes sense, should we rename `--dangerously-disable-dynamodb-cache` and `--dangerously-disable-incremental-cache` to keep it consistent?

khuezy:
I'm fine w/ those changes.
1) We need to add documentation on SST to emphasize that the streaming is EXTREMELY EXPERIMENTAL... AWS has updated their runtime in the past which broke streaming... it seems stable now but if they do another rugpull, it could cause user apps to break
2) I think it feels more natural have feature sets as "enable" rather than "disable"... but that's semantics.

If @conico974 doesn't have any objects, we can update `open-next` and rename the fields for `2.2.2`

conico974:
It make sense to move everything under experimental and to go with enable.
However i'm not sure about `enableOnDemandRevalidation`, `res.revalidate` works without it. 
Maybe something like `enableTagsRevalidation` would make more sense, and even this could cause confusion since `revalidatePath` use tags under the hoods. 
In open-next, I think we should keep it as disable though, but we could use experimental instead of dangerous and rename dynamodb. 
We could do it later though, no need to create an empty 2.2.2. We could add some alias so that it won't break sst.

PR #3305, opened 2023-09-15
Env vars in .env files should not override existing env vars

### LINKED ISSUE #2523: Environment variables don't override values in .env
According to the documentation (https://docs.sst.dev/config#other-environment-variables), dotenv shouldn't override any existing environment variables. The current implementation has all four invocations of dotenv configured to override (https://github.com/motdotla/dotenv#override).

In `packages/sst/src/cli/sst.ts`:
```
import dotenv from "dotenv";
dotenv.config({
  override: false,
});
dotenv.config({
  path: ".env.local",
  override: false,
});
```

And in `packages/sst/src/project.ts`:
```
  dotenv.config({
    path: path.join(project.paths.root, `.env.${project.config.stage}`),
    override: false,
  });
  dotenv.config({
    path: path.join(project.paths.root, `.env.${project.config.stage}.local`),
    override: false,
  });
```

I'm not sure if this is an mistake in documentation or implementation, but I'm happy to raise a pull request if helpful.

ackinc:
Fixes #2523 

Notes:
- An existing PR (https://github.com/sst/sst/pull/2934) for the same issue was closed before review/merge
- Thought it would be better to keep the override behavior explicit in the code rather than rely on dotenv's defaults. Let me know if you guys really want the `override` attr gone

### DIFF
--- a/packages/sst/src/cli/sst.ts
+++ b/packages/sst/src/cli/sst.ts
@@ -9,11 +9,11 @@ import { Logger } from "../logger.js";
 
 import dotenv from "dotenv";
 dotenv.config({
-  override: true,
+  override: false,
 });
 dotenv.config({
   path: ".env.local",
-  override: true,
+  override: false,
 });
 
 import { bootstrap } from "./commands/bootstrap.js";
--- a/packages/sst/src/project.ts
+++ b/packages/sst/src/project.ts
@@ -203,10 +203,10 @@ export async function initProject(globals: GlobalOptions) {
 
   // Load .env files
   [
-    path.join(project.paths.root, `.env`),
-    path.join(project.paths.root, `.env.local`),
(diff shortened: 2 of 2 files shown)


jayair:
Cool thanks!

fwang:
Thanks @ackinc!

We still want `.env.${stage}` to override those in `.env`. The order from the highest priority to lowest priority is:
- `.env.${stage}.local`
- `.env.${stage}`
- `.env.local`
- `.env`

ackinc:
Ah sorry. I've fixed the order of precedence. @fwang

fwang:
The issue is `sst.ts` is called before `project.ts`. So `.env` and `.env.local` are always loaded first. This might be exposing a flaw in how SST loads dotenvs.

@thdxr any insight on how we can fix this?

thdxr:

PR #3310, opened 2023-09-16
Fix RemixSite enforcing ESM

Nirlah:
Enforcing a `serverModuleFormat` is no longer needed now that both formats are supported.
Fixes the issue raised at  https://discord.com/channels/983865673656705025/1072426878431940758/1152175369823211573


fwang:
Good catch @Nirlah!

PR #3318, opened 2023-09-19
fix: update image optimizer cache policy

gkaskonas:


khuezy (approved):
LGTM. Do we want to allow the user to specify any of the cache policy props for customizations? I don't see a need, but someone might have a use case for it.

gkaskonas:
I added this policy manually to my project and lambda is now failing with

`Cannot destructure property 'url' of 'a' as it is undefined.`

khuezy:
@gkaskonas can you add `"accept"` to your header cache. I think that might be required.

khuezy (changes requested):
I believe we need to add "accept" to the header cache.

gkaskonas:
I don't think you need the accept header in the cache

khuezy:
ok, everything works now?

gkaskonas:
Yeah everything is working with `Query strings - All` cache key settings. I am getting better hit ratio on the images

conico974:
You do need to add the accept headers, otherwise it might send the wrong format to the end user https://github.com/vercel/next.js/blob/ffafad2c35a3f9677bb520f68e989f58f192b931/packages/next/src/server/image-optimizer.ts#L712

Tutai89:
Thanks

gkaskonas:
Added accept header

khuezy:
Thanks @gkaskonas one more thing please: can you add comments w/ a little explanation and the link to the Nextjs repo that Conico linked? That'll help other devs understand the context more in the future.

khuezy:
Just FYI, w/o the `accept` header, the image optimization won't optimize to `webp`

gkaskonas:
Added the comment

khuezy (approved):

gkaskonas:
Does accept need to be uppercase?

khuezy:
I don't think so, the old policy has lower case everything and it works. But if you want to be 100% sure, then change it to "Accept"

gkaskonas:
If it works it works. Who can merge this?

jayair:
Can we fix the conflicts and I'll ask the team to review? Unfortunately, we did a big refactor of the SSR Sites to make it easier to take contributions. So the merge might not be super simple.

gkaskonas:
Not sure if thats the cleanest way to set u p a different policy for image optimization. I am open to suggestions

thdxr:

PR #3323, opened 2023-09-20
Update ExampleStack.ts

ericpyle:
The use of `VITE_APP_AUTH0_DOMAIN` in the frontend assumes that `https://` is not already a part of the VITE_APP_AUTH0_DOMAIN. But the backend expects it, otherwise aws gives the error ""Invalid issuer: Issuer is not a valid URL for JWT Authorizer"

```
getAccessTokenSilently({
        audience: `https://${import.meta.env.VITE_APP_AUTH0_DOMAIN}/api/v2/`,
        scope: "read:current_user",
      });)
```

### DIFF
--- a/examples/api-auth-jwt-auth0/stacks/ExampleStack.ts
+++ b/examples/api-auth-jwt-auth0/stacks/ExampleStack.ts
@@ -7,8 +7,8 @@ export function ExampleStack({ stack, app }: StackContext) {
       auth0: {
         type: "jwt",
         jwt: {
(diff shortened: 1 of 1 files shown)


jayair:
Ah thanks. These files are generated. Can you make the change here instead and submit a new PR? https://github.com/sst/sst/blob/master/packages/create-sst/bin/presets/examples/api-auth-jwt-auth0/templates/stacks/ExampleStack.ts

PR #3326, opened 2023-09-20
[Construct/SSR]: Add optional viewerProtocolPolicy to SSR Functions

alexabidri:
This PR adds an optional `viewerProtocolPolicy` parameter to the SSR functions. This parameter allows the default value for cloudfront functions deployed with `REDIRECT_TO_HTTPS` to be overridden.

To use this new parameter, simply pass in a string value of either `https-only` or `redirect-to-https`. If no value is provided, the default behavior of `redirect-to-https` will be used.

This change should make it easier to customize the behavior of the SSR functions for different deployment scenarios.

_Note: Also added for StaticSite and Service construct_

### DIFF
--- a/packages/sst/src/constructs/Service.ts
+++ b/packages/sst/src/constructs/Service.ts
@@ -346,17 +346,6 @@ export interface ServiceProps {
    */
   customDomain?: string | ServiceDomainProps;
   /**
-   * The viewer protocol policy to use for the cloudfront distribution behaviours. Default is ViewerProtocolPolicy.REDIRECT_TO_HTTPS.
-   * @default ViewerProtocolPolicy.REDIRECT_TO_HTTPS
-   * @example 
-   * ```js
-   * {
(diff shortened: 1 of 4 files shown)


alexabidri:
@fwang Could you please have a check on this PR? Thanks a lot 🙏🏻

bayssmekanique (commented):
Alternative which avoids adding a top level `prop` is to use the attribute already defined on `cdk.distribution.defaultBehavior.viewerProtocolPolicy`.

```ts
viewerProtocolPolicy: cdk?.distribution?.defaultBehavior?.viewerProtocolPolicy ?? ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
```

alexabidri:
@bayssmekanique Thanks for reviewing! I have updated the PR according to your feedback 👍🏻

bayssmekanique (commented):
Looks great to me!

PR #3337, opened 2023-09-22
Update "secrets load" command to support the --fallback option

dfoulger:
Love the `secrets load` command for bulk-loading of secrets, but wanted to be able to quickly bulk-load fallbacks as well and didn't see the option, so here it is.


jayair:
Oh nice. Will ask the team to take a look.

Issue #3264, opened 2023-08-28
Support bun (an all-in-one JavaScript runtime)


dror-weiss:
Bun is growing and its official 1.0 launch is near (09-2023).

**I would love to see sst supporting bun as an AWS `provided` runtime.** 

The gist of my bun lambda setup:

1. Added a bun layer containing a bootstrap script and a binary (based on arm arch).
[see official docs](https://github.com/oven-sh/bun/tree/main/packages/bun-lambda#bun-lambda)

2. Wrote a lambda:
`lambda.ts`
```ts
export default { 
  async fetch(request: Request): Promise<Response> { 
    console.log(request.headers.get("x-amzn-function-arn")); 
    // ... 
    return new Response("Hello from Lambda!", { status: 200, headers: { "Content-Type": "text/plain", }, }); 
  }, 
}
```

3. Wrote an index file with the expected format by AWS:
`index.ts`
```ts
import lambda from "./lambda";

export default {
  js: lambda.fetch
};
```

4. Build the lambda using bun for the `bun` target:
```sh
bun build --target=bun --minify --outfile=index.js index.ts
```

5. Deploy `index.js` with matching runtime and arch.

jayair:
We've definitely looked at this. Thoughts @thdxr?

dror-weiss:
I've started working on a PR for this but it will take me some time to understand the code enough to complete it.

Right now I have encountered a few difficulties in implementing the build/cache process, despite `Bun.build` being heavily inspired by esbuild (what's used in node).

jayair:
Do you think we need to switch the esbuild part of it though?

dror-weiss:
I think the `bun` handler that needs to be added in `packages/sst/src/runtime/handlers` should use [`Bun.build`](https://bun.sh/docs/bundler).

I also think that someone with an intimate understanding of the codebase would be a better fit than me, as it's taking me more time than expected to issue a PR with this support. If there are guidelines for the codebase itself (more detailed than the one for contributors) I can try it.

5o50:
https://github.com/oven-sh/bun/issues/5549

ivawzh:
Any updates on supporting bun in general?

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #3339, opened 2023-09-22
Make SST @babel/core import compat with bun

JoltCode:
Currently, running `sst deploy` fails with the following:

```
import babel from "@babel/core";
       ^^^^^
SyntaxError: The requested module '@babel/core' does not provide an export named 'default'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:1[24](https://github.com/noot-oss/noot/actions/runs/6276959471/job/17047749347#step:6:25):21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)

Node.js v18.17.1
error: "sst" exited with code 1 (SIGHUP)
Error: Process completed with exit code 1.
```

I managed to get sst to work with bun, by just switching the import from a default import to a wildcard.

Perhaps this was only a side fix, and this is fixing the symptom rather than the actual problem? I'd be interested to hear some discussion about this, especially with Bun V1 just releasing 👍


JoltCode:
#3264

JoltCode:
Just looking at recent commits on some files, I think it would be worth giving a CC to @thdxr

PR #3343, opened 2023-09-23
Update ExampleStack.ts

ericpyle:
(moved from https://github.com/sst/sst/pull/3323)

The use of `VITE_APP_AUTH0_DOMAIN` in the frontend assumes that `https://` is not already a part of the VITE_APP_AUTH0_DOMAIN. But the backend expects it, otherwise aws gives the error ""Invalid issuer: Issuer is not a valid URL for JWT Authorizer"

```
getAccessTokenSilently({
        audience: `https://${import.meta.env.VITE_APP_AUTH0_DOMAIN}/api/v2/`,
        scope: "read:current_user",
      });)
```


jayair:
Thank you!

Issue #3344, opened 2023-09-23
The `StaticSite` construct fails if the path contains a space


morficus:
https://discord.com/channels/983865673656705025/1152257795559784478
---
link to discuss ☝️ 

As the title states, if the path to the project contains a space, then the build process for the `StaticSite` construct will fail. This also seems to apply to the `SsrSite` and `NextJsSite` (well, the deprecated one) constructs as well.

For example:
If the project path is `/home/morficus/my cool project/app` then things will fail. But if the path is `/home/morficus/my-cool-project/app` then everything is fine.

morficus:
After doing quite a bit of digging, I was able to determine that the root cause is because `child_process.exec()` (and by extension, `execSync()` do not support spaces in the command path. You can read all about it on this NodeJS issue from 2016: https://github.com/nodejs/node/issues/6803

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #3345, opened 2023-09-23
StaticSite and SsrSite: Fixes issues #3344

morficus:
See issue #3344 for details

After doing quite a bit of digging, I was able to determine that the root cause is because ch`ild_process.exec()` (and, by extension, `execSync()`) does not support spaces in the command path. You can read all about it on this NodeJS issue from 2016: https://github.com/nodejs/node/issues/6803

Based on the comments from the NodeJS issue, the proper solution is to escape (not encode) the path with spaces. On MacOS, this can be done by doing `\\ `, which is not cross-system compatible, but wrapping the path in `" "` is indeed supported on all operating systems.

So this PR is doing just that: wrapping the `script` and `zipPath` values in double quotes.

I checked for other instances of `execSync` in the code, but all the others appeared safe. These seem to be the only ones that failed because they reference an external script (`support/base-site-archiver.mjs`) using an absolute path.


thdxr:

Issue #1068, opened 2021-11-19
Passing a JS/TS lambda handler by reference instead of src path


maciejgoscinski:
Hello!

At the moment, the only way to connect a lambda handler function to an APIGw route (or Function construct) is to pass it as a string path:
`src/blabla/certainLambdasFile.handlerFunction`. This is consistent with how CloudFormation refers to the handler method, but otherwise pretty gimmicky as something to be written "by hand".

When working in an all-typescript project, possibly a monorepo, it would be super handy to pass the handler by reference:

```
import { handlerFunction} from 'src/blabla/certainLambdasFile';

// ...
'GET /myLambda' : handlerFunction
// ..
```

There's been a [discussion on Slack](https://serverless-stack.slack.com/archives/C01JG3B20RY/p1634222371143300) explaining it's rather difficult from the architecture standpoint.

Perhaps it would be possible to bypass the architectural problems, by "cheating" at build-time, swapping the references with their paths?

Also, is a "freely dangling" exported function the only way to export a handler? Would a class method, or the following syntax be possible anyhow?
```
export function MyLambdas() {
   function handler() { }

   return { handler }
}
```

CorentinDoue:
With the serverless framework and CDK, we declare the configuration of the function in the same folder as the handler and use `__dirname` to resolve the path:

src
L_ functions
     L_ myFunction
         L_ config.ts
         L_ handler.ts
L_ stack.ts

```ts
// config.ts

export const myFunction = new NodejsFunction(this, 'Lambda', {
      entry: getCdkHandlerPath(__dirname),
      handler: 'main',
      runtime: Runtime.NODEJS_18_X,
      architecture: Architecture.ARM_64,
});
```

But as SST compiles all the stack in a single `.sst.config.1695484916830.mjs` before executing it, the dirname is not the right one. 

It could probably be done with an esbuild plugin to resolve the path when SST build the `.sst.config.1695484916830.mjs`

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #3348, opened 2023-09-24
feat: resolve __SST_DIRNAME and __SST_FILENAME at built time

CorentinDoue:
## Why
resolve https://github.com/sst/sst/issues/1068

This enables to reference file paths as it can be done with `__dirname` or `__filename` on unbundled files.

This is useful for creating utils to link handler files to their trigger declaration in SST. 

```ts
export const getHandlerPath = (
  dirname: string,
  handlerRelativePath = 'handler.main',
): string => {
  const processRunLocation = process.cwd();
  const relativeDirname = dirname.replace(processRunLocation + '/', '');

  return [relativeDirname, handlerRelativePath].join('/');
};
```

### First example: declare triggers next to the handlers

```
├─ packages
|  └─ functions
|     └─ src
|        └─ myFunction
|           ├─ config.ts
|           └─ handler.ts
└─ stacks
   └─ MyStack.ts
```

```ts
// packages/functions/src/myFunction/handler.ts
export const main = ApiHandler(async (_evt) => {
  return {
    statusCode: 200,
    body: `Hello world. The time is ${new Date().toISOString()}`,
  };
});
```

```ts 
// packages/functions/src/myFunction/config.ts
export const configureMyFunction = ({ api }: MyStackServices): void => {
  api.addRoutes(api, {
    "GET /myFunction": getHandlerPath(__SST_DIRNAME),
  });
}
```

```ts
// stacks/MyStack.ts
export const MyStack = ({ stack }: StackContext) => {
  const api = new Api(stack, "Api");

 const myStackServices: MyStackServices = { api };

 configureMyFunction(myStackServices);
};
```

### Second example: declare handler paths next to the handler

```
├─ packages
|  └─ functions
|     └─ src
|        └─ myFunction
|           ├─ handlerPath.ts
|           └─ handler.ts
└─ stacks
   └─ MyStack.ts
```

```ts
// packages/functions/src/myFunction/handler.ts
export const main = ApiHandler(async (_evt) => {
  return {
    statusCode: 200,
    body: `Hello world. The time is ${new Date().toISOString()}`,
  };
});
```

```ts 
// packages/functions/src/myFunction/handlerPath.ts
export const myFunctionPath = getHandlerPath(__SST_DIRNAME);
```

```ts
// stacks/MyStack.ts
export const MyStack = ({ stack }: StackContext) => {
  const api = new Api(stack, "Api", {
    routes: {
      "GET /myFunction": myFunctionPath,
    },
  });
};
```

In both cases:
- it's easy to navigate from the handler to its trigger and the other way around with CMD+click. 
- no more hard-coded magic string 

We use this util with Serverless Framework and CDK which supports `__dirname`. It's painful to not have an alternative in SST.

## What

Create two new global variables `__SST_DIRNAME` and `__SST_FILENAME`:
- They never really exist
- But they are declared to lure typescript.
- They are replaced by their `dirname` or `filename` at the built time by an esbuild plugin.

## Impact

On my small project.

Without the plugin, the build of the config file takes ~3ms.

The esbuild plugin takes between 1 ms and 5 ms to parse a file. It becomes preponderant in the build process but it remains fast. The parsed files are only the IaC files, files compiled for lambda runtime are not impacted.

For projects with thousands of IaC files, it can be felt. But when you have a thousand lambdas you want to avoid having thousands of magic strings to link handlers to triggers 😜.

As this feature impacts performance, I propose to opt-in it. **WDYT?** **What is the best way to do this?** A new param `advanced.enablePathInjection` in SSTConfig?

## TODO before merge

[] Discuss this with one of the core team @thdxr @fwang @jayair 
[] Validate & implement opt-in if necessary
[] Document. **Where is the appropriate place to document this?**


thdxr:
This is kind of interesting - I'll need to think about the approach a bit.

There are a bunch of cases where importing code from inside sub-packages causes problems but I generally like the idea of a stronger tie between these things.

packages/sst/src/stacks/build.ts
```diff
@@ -80,6 +81,38 @@ export async function load(input: string, shallow?: boolean) {
...
+                plugins: [ts],
+              });
+              babel.traverse(ast, {
+                Identifier(path) {
+                  if (path.node.name === "__SST_FILENAME") {
+                    path.replaceWith(stringLiteral(args.path));
+                  }
+                  if (path.node.name === "__SST_DIRNAME") {
+                    path.replaceWith(
+                      stringLiteral(args.path.split("/").slice(0, -1).join("/"))
```
CorentinDoue, line 104: replace with `path.parse`, `path.join`

CorentinDoue:
Sorry @thdxr I struggled to find time to work on this.

> There are a bunch of cases where importing code from inside sub-packages causes problems but I generally like the idea of a stronger tie between these things.

I agree with you, there are risks of mixing IaC typescript code with runtime typescript, especially if you use barrel files. In the worst-case scenario, you end up with all the CDK code packaged in a lambda or executing runtime cold start code when running an SST command.

To prevent this, we use several countermeasures:
- Use specific, dedicated, and standardized filenames for IaC files in runtime packages: `config.ts`
- [Configure our IDE to use a specific icon](https://code.visualstudio.com/api/extension-capabilities/theming#file-icon-theme) for `config.ts` to make their difference explicit to developers
- Use [no-restricted-imports eslint rule](https://typescript-eslint.io/rules/no-restricted-imports/) to prevent importing `config.ts` in runtime code 
- Use [dependency cruiser](https://github.com/sverweij/dependency-cruiser) to make more specific import rules to avoid any mix. (We actually haven't got it so far, eslint already covers a bunch of use cases. But it's the ultimate weapon against unwanted imports)

On the other hand, it's powerful to use common interfaces between the runtime world and IaC work to ensure they are synced. We are happy to use contract files in both handlers (runtime code) and config (IaC code) to strongly type the handler event and define the trigger according to a single source of truth. 

```mermaid
flowchart TD
    stack[stack.ts] -->|import| config[config.ts]
    config -->|import| contract[contract.ts]
    handler[handler.ts] -->|import| contract
```

Here is a concrete example with [Swarmion API Gateway contract](https://www.swarmion.dev/docs/how-to-guides/use-serverless-contracts/api-gateway)

CorentinDoue:
I don't know how to make it opt-in. I thought about adding a new config option `advanced.resolveBuildTimePaths?: boolean` but the config is resolved from the very built file I'm manipulating.

I thought about adding a third esbuilt plugin to resolve this config option from an ast point of view, but it's very unstable. If someone replaces `resolveBuildTimePaths: true` by `const resolveBuildTimePaths = true;` then use the variable in the config, it breaks the resolver

thdxr:
ah can you resolve conflicts? probably just a pnpm i

CorentinDoue:
> This change looks pretty minimal so looks good to me, i'll merge

Without any documentation ? I think we should write somewhere that those variables are available

CorentinDoue:
> ah can you resolve conflicts? probably just a pnpm i

@thdxr Done ✅

CorentinDoue:
Hi @thdxr 

I resolved the conflicts again. Do you think you could merge? 🙏

thdxr:

PR #3357, opened 2023-09-28
Function: property to disable CloudWatch logs

### LINKED ISSUE #3355: Function: option to disable CloudWatch logs
Many observability tools provide mechanisms to send logs directly to their APIs via Lambda extensions and the Lambda Telemetry API. This has the benefit of avoiding the ingestion and storage costs of CloudWatch. However, disabling CloudWatch logs for a Lambda function requires removing IAM permissions from the function for sending logs to CloudWatch.

I'd like to propose adding a convenience flag (`disableLogs` or `disableCloudWatchLogs`) to the Function construct to explicitly DENY these CloudWatch Logs permissions in the function's IAM role:

- `logs:CreateLogGroup`
- `logs:CreateLogStream`
- `logs:PutLogEvents`

The IAM explicit DENY will always win so this should not conflict with existing permissions added to the role or anything else a user might add to the role.

It is possible to do this manually but it would be convenient for this use case to have it on the Function construct so it can be set as a default function property project wide.

I would be happy to submit a PR for this.

blytheaw:
Resolves #3355


arpadgabor:
lgtm, will this work for `defaults` setting as well to disable all logging for all functions at once?

blytheaw:
@arpadgabor Yes, the function is exposed in all the usual areas. I am investigating an issue though where enabling this flag and disabling is not removing the added permissions. Will have an update soon.

blytheaw:
> @arpadgabor Yes, the property is exposed in all the usual areas (defaults, setDefaultFunctionProps, etc). I am investigating an issue though where enabling this flag and disabling is not removing the added permissions. Will have an update soon.

This issue has been fixed. I missed setting the default value for the property. Tested added and removing both on specific functions as well as higher level defaults and then removing the flag. Permissions were updated and worked as expected.

fwang:
Thanks @blytheaw! Made a trivial change. Renamed the prop to `disableCloudWatchLogs`, defaulting to `false`. Wanted to give it a more explicit name since it's not a commonly used flag.

PR #3363, opened 2023-10-02
feat: Improved Astro SSR Support

### LINKED ISSUE #3139: Astro integration is missing basic functionality

bayssmekanique:
_Continuation of the work started in #3176 and leveraging the improvements made by @fwang in #3340._

PR improves the implementation of Astro SSR for regional deployments (CloudFront + S3 & Lambda). The Astro plugin now produces a meta file at build time containing the routing information for the site which is used to produce a deployment specific CF Function. The deployment strategy also shifts from behavior mapping of static files to a blanket S3 attempt on most routes with a fall-through to the Lambda function on a miss.

**Advantages of this implementation:**
- CF Function performs document postfixing (`/index.html`) for directory routing, so `coolsite.com/about/index.html` now becomes `coolsite.com/about`.
- CF Function performs mapped redirects at the CDN level without activating Lambda function.
- Allows static and dynamic content in the same directory.
- Removes limitation of 25 static files/directories in the root of the site.

**Disadvantages of this implementation:**
- Attempts S3 before Lambda fall-through so theoretical overhead added to all SSR requests.
- All routes only support `GET`, `HEAD`, `OPTIONS` methods by default.

Both of the limitations noted as disadvantages are mitigated through use of the `ssrExclusiveRoutes` property on the SST Config. The route patterns from this property are mapped to behaviors which bypass the S3 static file check and support all request methods (`GET`, `POST`, `PUT`, `DELETE`, etc.). These are also route patterns, so API routes only require declaration of the root directory (`/api/*`).

Example configuration for a simple site:
```ts
app.stack(function Site({ stack }) {
  const site = new AstroSite(stack, `astro-site`, {
    ssrExclusiveRoutes: [
      'feedback', // Feedback page which requires POST method
      'login',    // Login page which requires POST method
      'user/*',   // Directory of user routes which are all SSR
      'api/*'     // Directory of API endpoints which require all methods
    ]
  })
})
```

Resolves #3139

### DIFF
--- a/packages/astro-sst/package.json
+++ b/packages/astro-sst/package.json
@@ -39,12 +39,12 @@
     "build": "tsc"
   },
   "dependencies": {
-    "@astrojs/webapi": "^2.1.0",
+    "@astrojs/webapi": "^3.0.0",
     "set-cookie-parser": "^2.6.0"
   },
   "devDependencies": {
     "@types/aws-lambda": "^8.10.112",
     "@types/set-cookie-parser": "^2.4.3",
-    "astro": "^2.1.3"
+    "astro": "^3.1.4"
   }
 }
--- a/packages/astro-sst/src/edge/adapter.ts
+++ b/packages/astro-sst/src/edge/adapter.ts
@@ -1,4 +1,5 @@
 import type { AstroAdapter, AstroIntegration } from "astro";
+import { BuildMeta } from "../lib/build-meta";
 
 const PACKAGE_NAME = "astro-sst/edge";
 
@@ -14,8 +15,13 @@ export default function createIntegration(): AstroIntegration {
   return {
     name: PACKAGE_NAME,
     hooks: {
-      "astro:config:done": ({ setAdapter }) => {
+      "astro:config:done": ({ config, setAdapter }) => {
+        BuildMeta.setAstroConfig(config);
         setAdapter(getAdapter());
+      },
+      "astro:build:done": async (buildResults) => {
+        BuildMeta.setBuildResults(buildResults);
+        await BuildMeta.exportBuildMeta();
       },
     },
   };
--- a/packages/astro-sst/src/lambda/adapter.ts
+++ b/packages/astro-sst/src/lambda/adapter.ts
@@ -1,4 +1,5 @@
 import type { AstroAdapter, AstroIntegration } from "astro";
+import { BuildMeta } from "../lib/build-meta";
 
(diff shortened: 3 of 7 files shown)


bayssmekanique:
The Changeset could use some consideration. The changes for regional deployments could be fairly drastic since the origins and behaviors are going to be flipped on their heads, but a major bump is certainly not necessary for the other packages.

Most sites should continue to operate without issue as routes which had previously been specified with full paths (containing `/index.html`) will continue to work. The only issues would be if they did not desire to support auto-postfixing for static routes (they want to require specifying `/index.html` on directory calls).

This case could be handled by adding a configuration to enable the postfix functionality and set it to `false` by default to ensure backwards compatibility, but it will be at the cost of improved DX for new users moving forward until it's changed to `true` since the default behavior of the dev server is to postfix the routes.

PR #3368, opened 2023-10-03
Add option for using prebuilt Rust binary

oscartbeaumont:
I am currently messing around with SST and deploying Rust lambda functions. Something Rust is infamously known for is it's slow compile times and I am really feeling this when working with Seed as the Rust must recompile on all stages (dev and for me multiple prod stages).

This PR adds an option called `prebuilt` that can be used to provide a prebuilt binary for the Lambda. I am pairing this with some custom caching code (backed by an S3 bucket) that allows me to only build the Lambda when the Rust code actually changes, letting me reuse the previously built binary when redeploying the same code to a different stage.

It can be used like the following:

```js
new Function(stack, "backend", {
  rust: {
    prebuild: "./target/lambda/main/bootstrap",
  },
});
```

### DIFF
--- a/.changeset/sweet-buttons-cry.md
+++ b/.changeset/sweet-buttons-cry.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Add option for using prebuilt Rust binary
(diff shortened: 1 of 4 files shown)


jayair:
Hmm that's interesting. Let me ask the team to give some feedback.

thdxr:

Issue #3369, opened 2023-10-03
SSR Function ignoring props


bayssmekanique:
As part of the SSR Restructure (#3340) the SSR Function props started ignoring the `memorySize` and `timeout` values passed into the constructor.

https://github.com/sst/sst/blob/c61396b0a9197dcac59bc3ffd5bdbdc308b47f30/packages/sst/src/constructs/SsrFunction.ts#L90-L96

Issue #3366, opened 2023-10-03
SsrFunction overrides some props


dani-mp:
`SsrFunction` is overriding some props (e.g. timeout), hence not respecting what's defined in userland.

I noticed this when I upgraded sst to v2.26.9 and my RemixSite logs showed that all the invocations durations were max 10k ms.

See https://github.com/sst/sst/blob/347666100a43047512b614fb785bedd61e4a911f/packages/sst/src/constructs/SsrFunction.ts#L86-L96

bayssmekanique:
Resolved in #3370

PR #3373, opened 2023-10-04
Respect ssr function props

dani-mp:
Fixes https://github.com/sst/sst/issues/3366
Fixes https://github.com/sst/sst/issues/3369


fwang:
Thanks @dani-mp! Recently fixed in #3370

PR #3379, opened 2023-10-06
[WIP] feature: support remix streaming

michaelgmcd:
Opening this after implementing it in our project. This PR adds support for streaming Remix apps. Unsure how you guys would like to configure it.

### DIFF
--- a/packages/sst/src/constructs/RemixSite.ts
+++ b/packages/sst/src/constructs/RemixSite.ts
@@ -35,7 +35,7 @@ export class RemixSite extends SsrSite {
     const { path: sitePath, edge } = this.props;
 
     const { handler, inject } = this.createServerLambdaBundle(
-      edge ? "edge-server.js" : "regional-server.js"
+      edge ? "edge-server.js" : "regional-server-stream.js"
     );
     const format = this.getServerModuleFormat();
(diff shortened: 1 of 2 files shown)


michaelgmcd:
Finding some issues with this. Would love to have someone on the team take a look. Streaming seems to work, but some responses time out.

mrconnerton-encorehc:
Reposting from discord: 

So I tried my hand at streaming but didn't quite make it. I cobbled this together from a number of previously referenced sources here. After issues with getting it to just load or cookies, I finally have it working "normally" where the site functions as it would non streaming mode. The actual defer / streaming bits are not actually working. I saw some references to api gateway not supporting streaming, but its my understanding that the current setup has cloudfront hit the lambda url direct. 

This is remix 1.19.3  with all the v2 flags enabled. Haven't had time to fix some issues upgrading to v2 caused in our app. However, the streaming bits work fine running locally in dev.

Anyway, here is my wrapper alterations with the pnpn patch if others that are working on this , or more impressively actually have it working could take a look:
https://gist.github.com/mrconnerton-encorehc/ae2a60ad42fc71b7acc1a6ac80b68bdb

thdxr:

PR #3381, opened 2023-10-06
feat: Astro SSR response modes (buffered/streamed)

bayssmekanique:
Adds ability to specify response mode (`buffer` or `stream`) and sets up environment appropriately.

- Adds `responseMode` property to `astro-sst` adapter integration function.
- Moves `serverRoutes` property to `astro-sst` adapter integration function.
- Deprecates `serverRoutes` property on `AstroSite` constructor.

Response mode should be set in the adapter integration function call in `astro.config`:
```js
import { defineConfig } from "astro/config"
import aws from "astro-sst/lambda"

export default defineConfig({
  adapter: aws({
    responseMode: "stream",
    serverRoutes: [
      "feedback", // Feedback page which requires POST method
      "login",    // Login page which requires POST method
      "user/*",   // Directory of user routes which are all SSR
      "api/*"     // Directory of API endpoints which require all methods]
  })
})
```

### DIFF
--- a/packages/astro-sst/src/lambda/entrypoint.ts
+++ b/packages/astro-sst/src/lambda/entrypoint.ts
@@ -1,19 +1,36 @@
 import type { SSRManifest } from "astro";
 import type { APIGatewayProxyEventV2 } from "aws-lambda";
-import type { ResponseStream } from "../lib/types";
+import type { ResponseMode, ResponseStream } from "../lib/types";
 import { NodeApp } from "astro/app/node";
 import { polyfill } from "@astrojs/webapi";
-import { convertFrom, convertTo } from "../lib/event-mapper.js";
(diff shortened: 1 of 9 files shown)


packages/astro-sst/src/lambda/entrypoint.ts
```diff
@@ -50,20 +58,49 @@ export function createExports(manifest: SSRManifest) {
...
+        type: internalEvent.type,
+        response: new Response("Not found", { status: 404 }),
+      });
+    }
+
+    // Process request
+    const response = await app.render(request, routeData);
+    debug("response", response);
+
+    // Stream response back to Cloudfront
```
adamelmore, line 89: nit: "Stream response" -> "Send response"?

Issue #3261, opened 2023-08-27
[Feature request] Include source in published package and set declarationMap to true


steve-taylor:
When command/CTRL-clicking something imported from SST in Visual Studio code, I always end up at the type definition file, which isn't very useful. I can right-click the editor tab and then select _Reveal in Explorer View_, which takes me to the compiled JavaScript, but I have to flick back and forth to see the types associated with the implementation. What I ideally want is to navigate to the source and, fortunately, TypeScript has a nice feature to enable it.

Simply add `"declarationMap": true` to `tsconfig.json`'s `compilerOptions` and include SST's source in the published package. We'll now be able to navigate directly to the implementation source in Visual Studio Code. This will greatly improve the developer experience by easily being able to see what SST is doing under the hood.

jayair:
Oh interesting. Let me share this internally.

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #3386, opened 2023-10-07
DX: TypeScript declaration maps

steve-taylor:
This change adds declaration maps to the `sst`, `astro-sst` and `svelte-kit-sst` packages, so that developers who use these packages can directly navigate to the source within Visual Studio Code (typically via command/ctrl-click) rather than navigating to `.d.ts` files.

To make declaration maps useful, published packages must include their source, so I've configured these packages to include their source. In the case of `sst`, I had to change the location of its build artifacts from the root of the published package to the `dist` directory, enabling both the inclusion of `src` and the correctness of declaration map references to source files.

This addresses #3261.

I also fixed some other issues:
* incorrect `repository.directory` in `packages/sst/package.json` (was `"packages/cli"`)
* redundant `"declaration": true` in [`packages/sst/support/tsconfig.json`](/sst/sst/blob/master/packages/sst/support/tsconfig.json#L4)
* incorrect `astro-sst` version in `pnpm-lock.yaml` (which would have likely resulted in a failed CI build)

### Updated

* I tried a different approach. See [this comment](https://github.com/sst/sst/pull/3386#issuecomment-1761117697).
* The incorrect `astro-sst` version in `pnpm-lock.yaml` seems to have been resolved in the latest master branch commit.


thdxr:
Hey thanks for submitting this.

So the reason we publish out of the dist folder (as annoying as it is) is to allow certain scenarios to import from `sst/node` directly without specifying a sub folder

nextjs for example a lot of people are not using using a configuration that respects the `exports` field

is there a way to do what you want without changing the fact that we publish from dist?

steve-taylor:
> Hey thanks for submitting this.
> 
> So the reason we publish out of the dist folder (as annoying as it is) is to allow certain scenarios to import from `sst/node` directly without specifying a sub folder
> 
> nextjs for example a lot of people are not using using a configuration that respects the `exports` field
> 
> is there a way to do what you want without changing the fact that we publish from dist?

@thdxr I tried a different approach, which involves copying `src` to `dist/src`, so the source is included in the published package, and then compiling from `dist/src` instead of `src` so that the generated declaration maps point to the sources included in the published package.

packages/sst/build.mjs
```diff
@@ -1,7 +1,14 @@
 import esbuild from "esbuild";
 import fs from "fs/promises";
 
-const pkg = await fs.readFile("package.json").then(JSON.parse);
+// Copy package.json to dist, excluding publishConfig
```
steve-taylor, line 4: I'm not really sure why dist/package.json didn't have a `publishConfig` section. I assumed there was a reason, so it's being excluded when copying package.json to dist/package.json.

thdxr:

PR #3390, opened 2023-10-09
feat: Adds layers to CDK props

bayssmekanique:
Adds layers as a CDK prop on the `SsrSite` construct to allow adding layers to Lambda functions.

Reference configurations:
```js
import { AstroSite } from 'sst/constructs'
import type { SSTConfig } from 'sst'
import * as lambda from 'aws-cdk-lib/aws-lambda'

const layerArn = 'arn:aws:lambda:us-west-2:764866452798:layer:chrome-aws-lambda:37'

export default {
  config() {
    return {
      name: 'test-website',
      region: 'us-west-2',
      profile: 'sst'
    }
  },
  stacks(app) {
    app.stack(function Site({ stack }) {
      const site = new AstroSite(stack, 'astro-site', {
        cdk: {
          server: {
            layers: [lambda.LayerVersion.fromLayerVersionArn(stack, 'ChromeLayer', layerArn)]
          }
        }
      })

      stack.addOutputs({
        url: site.customDomainUrl
      })
    })
  }
} satisfies SSTConfig
```

### DIFF
--- a/.changeset/dirty-bees-taste.md
+++ b/.changeset/dirty-bees-taste.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Adds Lambda layer support to SSR Site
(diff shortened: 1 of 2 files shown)


bayssmekanique:
Functionality confirmed with the above noted configuration.

image

alistairstead:
Really keen to see this merged to be able to use astro:assets Image

Issue #3389, opened 2023-10-09
bug: Astro site overriding default S3 caching headers


bayssmekanique:
By default, S3 static files receive `cacheControl` headers from the `createS3OriginAssetFileOptions` method on the `SsrSite` construct. When specifying `fileOptions`, this method gets bypassed in favor of the injected file options (rather than merged).

The default implementation for `AstroSite` assumed these options were merged resulting in files that aren't `.css`, `.js`, or `.html` from being tagged with `cacheControl` headers.

- [x] Revert the `fileOptions` defaults from the `AstroSite` implementation to allow the default operation.
- [ ] Investigate adding a more comprehensive `cacheControl` and `contentType` tagging system for all SSR sites.

PR #3392, opened 2023-10-10
feat: More explicit SSR caching and invalidation

bayssmekanique:
Adds extension based content types along with cache headers setting for SSR sites.

The current issue is that invalidation is not correctly being detected so updated content is continuing to be stale on the server with the current cache headers.  Should content caching be reduced or should invalidation function be improved?

Related to #3389.


bayssmekanique:
This PR has grown significantly based on the discovery of a few oddities in the implementation and feedback from other devs in the Discord channel.

Issue/oddities discovered:
- The decision on whether or not to invalidate the CDN was based on changes to file names only. This means that changes to the contents of file which are not versioned would not cause cache invalidations. This is neither right or wrong, but not necessarily intuitive.
- Non-versioned files are cached for a really long time by default but changes to the files didn't always cause invalidations and the TTL was so long that the cache wouldn't be expected to expire.
- Content types of most files in the cache weren't set which meant that many pages served from cache had broken elements due to text encoding (like Emojis).
- Attempts to correct the above issues via the `fileOptions` parameter meant that you were opting out of all default file metadata tagging in the cache and responsible for setting all metadata.

---

This PR attempts to resolve these oddities and make a more consistent experience for all SSR deployment types by adding an explicit `cache` attribute to the `SsrSite` props. The `cache` attribute allows specifying the `textEncoding` for text based content, as well as cache TTL for both versioned and nonversioned files. There are also options for overriding the default cache strategy with a full cache header for both versioned and nonversioned files. Independent of these props, the distribution will now attempt to add metadata for common web file types in the cache.

The largest change, and likely the most controversial, is the ability to change the invalidation strategy via the `cache.cdnInvalidationStrategy` prop. The default value of `versioned` will only check for changes to the files within the versioned file directory and will only invalidate those files as part of the deployment. The `never` and `always` options are both absolute values and fairly straight forward. The `all` option will perform a full invalidation based on the change of the content of any of the files in the cache. 

Proposed changes to the `SsrSite.props`:
```ts
  cache?: {
    /**
     * Character encoding for text based assets stored in the S3 cache (ex: html, css, js, etc.). If "none" is specified, no charset will be returned in header.
     * @default utf-8
     * @example
     * ```js
     * cache: {
     *  textEncoding: "iso-8859-1"
     * }
     * ```
     */
    textEncoding?: "utf-8" | "iso-8859-1" | "ascii" | "none";
    /**
     * The strategy to use for invalidating the CDN cache. By default, the CDN cache will invalidate on changes any cached file, but this could become slow on very large projects.
     * - "never" - No invalidation will be performed.
     * - "all" - All files will be invalidated when any file changes. (Default, requires checking file content which will increase deployment time)
     * - "versioned" - Only versioned files will be invalidated when versioned files change.
     * - "always" - All files are invalidated on every deployment.
     * @default all
     * @example
     * ```js
     * cache: {
     *   cdnInvalidationStrategy: "versioned"
     * }
     * ```
     */
    cdnInvalidationStrategy?: "never" | "all" | "versioned" | "always";
    /**
     * The TTL for versioned files (ex: `main-1234.css`) in the CDN and browser cache. Ignored when `versionedFilesCacheHeader` is specified.
     * @default 1 year
     * @example
     * ```js
     * cache: {
     *  versionedFilesTTL: '30 days'
     * }
     * ```
     */
    versionedFilesTTL?: number | Duration;
    /**
     * The header to use for versioned files (ex: `main-1234.css`) in the CDN cache. When specified, the `versionedFilesTTL` option is ignored.
     * @default public,max-age=31536000,immutable
     * @example
     * ```js
     * cache: {
     *   versionedFilesCacheHeader: "public,max-age=31536000,immutable"
     * }
     * ```
     */
    versionedFilesCacheHeader?: string;
    /**
     * The TTL for non-versioned files (ex: `index.html`) in the CDN cache. Ignored when `nonVersionedFilesCacheHeader` is specified.
     * @default 1 day
     * @example
     * ```js
     * cache: {
     *  nonVersionedFilesTTL: '4 hours'
     * }
     * ```
     */
    nonVersionedFilesTTL?: number | Duration;
    /**
     * The header to use for non-versioned files (ex: `index.html`) in the CDN cache. When specified, the `nonVersionedFilesTTL` option is ignored.
     * @default public,max-age=0,s-maxage=86400,stale-while-revalidate=8640
     * @example
     * ```js
     * cache: {
     *   nonVersionedFilesCacheHeader: "public,max-age=0,no-cache"
     * }
     * ```
     */
    nonVersionedFilesCacheHeader?: string;
  };
```

The default configuration will assume that versioned content is desired to be long-lived (long TTL), and nonversioned content is desired to have fairly short CDN cache periods with hidden revalidation (aka stale while revalidated) and no user caching, with most deployments resulting in cache invalidation across the full site. This isn't the most "optimized" implementation, but as a default it offers the fewest pain-points since all changes should be visible immediately after deployment.

bayssmekanique:
It's also worth mentioning the list of `commonWebFileExtensions` included in this PR in case there are strong opinions on this list or it's location.

```ts
const commonWebFileExtensions: Record<
  string,
  { mime: string; isText: boolean }
> = {
  txt: { mime: "text/plain", isText: true },
  htm: { mime: "text/html", isText: true },
  html: { mime: "text/html", isText: true },
  xhtml: { mime: "application/xhtml+xml", isText: true },
  css: { mime: "text/css", isText: true },
  js: { mime: "text/javascript", isText: true },
  mjs: { mime: "text/javascript", isText: true },
  apng: { mime: "image/apng", isText: false },
  avif: { mime: "image/avif", isText: false },
  gif: { mime: "image/gif", isText: false },
  jpeg: { mime: "image/jpeg", isText: false },
  jpg: { mime: "image/jpeg", isText: false },
  png: { mime: "image/png", isText: false },
  svg: { mime: "image/svg+xml", isText: true },
  bmp: { mime: "image/bmp", isText: false },
  tiff: { mime: "image/tiff", isText: false },
  webp: { mime: "image/webp", isText: false },
  ico: { mime: "image/vnd.microsoft.icon", isText: false },
  eot: { mime: "application/vnd.ms-fontobject", isText: false },
  ttf: { mime: "font/ttf", isText: false },
  otf: { mime: "font/otf", isText: false },
  woff: { mime: "font/woff", isText: false },
  woff2: { mime: "font/woff2", isText: false },
  json: { mime: "application/json", isText: true },
  jsonld: { mime: "application/ld+json", isText: true },
  xml: { mime: "application/xml", isText: true },
  pdf: { mime: "application/pdf", isText: false },
  zip: { mime: "application/zip", isText: false },
};
```

bayssmekanique:
Another change introduced by this PR is the shift from `exclude` and `include` as properties on the `fileOptions` to a shared `filters` array which allows for the `exclude` and `include` filters to be applied in different orders to allow more specific targeting of files.

Previous file options:
```ts
const fileOptions = [
  {
    exclude: "*",
    include: "*.html",
    cacheControl: "max-age=0,no-cache,no-store,must-revalidate",
  },
  {
    exclude: "*",
    include: ["*.js", "*.css"],
    cacheControl: "max-age=31536000,public,immutable",
  },
];
```

New file options: 
```ts
const fileOptions = [
  {
    filters: [{ exclude: "*" }, { include: "*.html" }],
    cacheControl: "max-age=0,no-cache,no-store,must-revalidate",
  },
  {
    filters: [{ exclude: "*" }, { include: "*.js" }, { include: "*.css" }],
    cacheControl: "max-age=31536000,public,immutable",
  },
];
```
The advantage to this approach is that it allows filters to be specified in the manner that they will be applied in S3 (ref: [Use of Exclude and Include Filters](https://docs.aws.amazon.com/cli/latest/reference/s3/#use-of-exclude-and-include-filters)).

Previously, it was ~~impossible~~ very hard to apply a file option to all extensions in site while excluding a subset that were in a directory because all the `exclude` filters were applied first, followed by the `include`. With this new structure, the order can be specified to allow this sort of filtering.
```ts
const fileOptions = [
  {
    filters: [
      { exclude: "*" },
      { include: "*.csv" },
      { exclude: "public/*" },
    ],
    cacheControl: "private,max-age=0,no-cache,no-store",
  }
];
```

fwang:
@bayssmekanique made a few cosmetic changes along the way while learning the code.

Appreciate the detailed comments. Lemme know when it's ready to merge!

bayssmekanique:
Thanks @fwang. I made one reversion. The `if/else` logic in the invalidation digest section would result in only the versioned files being fed into the hash, when it should really be both in the case of an `all` invalidation strategy.

Ref: https://github.com/sst/sst/blob/10ce93795e4aabef7366570c8bce788895d164f9/packages/sst/src/constructs/SsrSite.ts#L1506-L1532

Everything else looks good. Ready to merge.

adamelmore:
if this can make it into a (snapshot?) release tonight i'd love to play with it tomorrow! cc @fwang

fwang:
> if this can make it into a (snapshot?) release tonight i'd love to play with it tomorrow! cc @fwang

sst@0.0.0-20231012035248

adamelmore:
fyi, i've had some other priorities shifting my attention and won't get back to this for a couple weeks :/

bayssmekanique:
@fwang, doesn't sound like @adamelmore will be able to QA this PR anytime soon. Can we merge for the next minor release?

Issue #3405, opened 2023-10-12
[Next.js] Lambda@Edge Server Function Handler load order


lucasvieirasilva:
I've trying to deploy a Next.js application using `edge: true` property, at first I was getting the following error in the server function.

```log
PermanentRedirect: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
```

after some investigation, I've noticed the `CACHE_BUCKET_REGION` env var was undefined.

Then started investigating how SST adds the process.env with the provided env vars, since the AWS Lambda@Edge doesn't support environment variables.

Then I noticed the `nextjssite-index.mjs` is used to define all the `process.env` before calling the `handler` from the `index.mjs`

Folder structure:
![image](https://github.com/sst/sst/assets/12723594/a86d6e1e-f5c3-42b4-91b9-24ca4adf7969)

After adding some console logs in the functions files, I've noticed the `index.mjs` is being loaded before the `nextjssite-index.mjs`, so, all the `process.env` are undefined.

PR #3406, opened 2023-10-12
fix(nextjs): lambda edge handler import order

lucasvieirasilva:
This PR fixes issue #3405, by removing the `wrapHandler` function, since the open-next already generates the `index.mjs` exporting the `handler`, there's no need for a wrapper.

I've tested it locally and works fine.


khuezy:
This was added to try to get top level await working for Config, see: https://github.com/sst/sst/pull/3411

jayair:
I saw the other issue. But I might be missing some context around this. Was there a Discord discussion around this?

fwang:
Fixed in #3422

PR #3407, opened 2023-10-12
SsrSite / NextJsSite: now expose default server cache policy props as static methods

dkershner6:


packages/sst/src/constructs/NextjsSite.ts
```diff
@@ -126,6 +127,19 @@ type NextjsSiteNormalizedProps = NextjsSiteProps & SsrSiteNormalizedProps;
  * ```
  */
 export class NextjsSite extends SsrSite {
+  public static readonly DEFAULT_CACHE_POLICY_ALLOWED_HEADERS = [
+    "accept",
+    "rsc",
+    "next-router-prefetch",
+    "next-router-state-tree",
+    "next-url",
+  ];
```
fwang, line 136: Does `DEFAULT_CACHE_POLICY_ALLOWED_HEADERS` need to be publicly?

dkershner6: Not explicitly, but I thought there might be a use case for someone wanting only them instead of the full cache policy, like altering some of the other characteristics or building a second cache policy and using it for only some behaviors.

PR #3409, opened 2023-10-12
AstroSite: Fixed invalid format for set-cookie headers in server response

ffxsam:
`set-cookie` headers were being incorrectly sent by the server, in this format:
```
set-cookie: Name=Value; Domain=xyz.com;Name2=Value2; Domain=xyz.com
```
This PR fixes it so they're properly returned in multiple headers:
```
set-cookie: Name=Value; Domain=xyz.com
set-cookie: Name2=Value2; Domain=xyz.com
```
This PR has been fully tested in a live staging environment.

### DIFF
--- a/.changeset/modern-pigs-push.md
+++ b/.changeset/modern-pigs-push.md
@@ -1,0 +1,5 @@
+---
+"astro-sst": patch
+---
+
+AstroSite: Fixed invalid format for set-cookie headers in server response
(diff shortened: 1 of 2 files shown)


bayssmekanique:
Header stringify function appears to have issue.

Using these cookie options:
```ts
Astro.cookies.set('sst', 'awesome')
Astro.cookies.set('age', 'eventual', { expires: new Date('1/1/2030') })
Astro.response.headers.set('Set-Cookie', 'a=b; Path=/;')
```

Results in this single header:
```
Set-Cookie: a=b; Path=/;, sst=awesome, age=eventual; Expires=Tue, 01 Jan 2030 00:00:00 GMT
```

This header does not result in 3 cookies as expected.

fwang:
Fixed in #3420

ffxsam:
Yep, closing! Thanks, @bayssmekanique, for picking up the slack since I was traveling. 🙂 Will test this shortly..

PR #3411, opened 2023-10-13
NextjsSite: fix wrapper function for streaming

fwang:


packages/sst/src/constructs/NextjsSite.ts
```diff
@@ -423,18 +423,25 @@ export class NextjsSite extends SsrSite {
...
-      [
-        `import { handler as rawHandler } from "./index.mjs";`,
-        `export const handler = (event, context) => {`,
-        `  return rawHandler(event, context);`,
-        `};`,
-      ].join("\n")
+      experimental?.streaming
+        ? [
+            `import { handler as rawHandler } from "./index.mjs";`,
+            `export const handler = awslambda.streamifyResponse(async (...args) => {`,
```
conico974, line 435: This will likely not work, or at least cause some issues.
In open-next when the streaming flag is there, we already add `awslambda.streamifyResponse` to the raw handler.

conico974 (commented):
What's the point of the `wrapHandler` function ? It does nothing except calling the `rawHandler` and it's private.

If we want to wrap the handler, i think a better approach would be using open-next plugin system.

packages/sst/src/constructs/NextjsSite.ts
fwang: hmm.. would this work?
```ts
export const handler = (...args) => {
  return rawHandler(...args);
};
```

khuezy: I was able to get this working using `awslambda.streamifyResponse`, it should work b/c it's just an additional callback... but let me confirm again

packages/sst/src/constructs/NextjsSite.ts
```diff
@@ -423,18 +423,25 @@ export class NextjsSite extends SsrSite {
...
+            `  await rawHandler(...args);`,
+            `};`,
```
khuezy, line 437: missing `)` to close the opening `streamifyResponse(`

packages/sst/src/constructs/NextjsSite.ts
khuezy: Before you merge, let me verify the actual streaming response is behaving correctly, eg SSE

fwang: Yeah it's ready to be tested. Do u need a snapshot release?

khuezy: Nope, I'm editing the node_modules directly

khuezy: Ok streaming works, LGTM

Issue #2267, opened 2022-12-20
RemixSite: allow configuration for cache headers and cloudfront invalidation


seanWLawrence:
I'd like to contribute a feature for making the cache headers and cloudfront invalidation options configurable. For example, something like:

## API 

```ts
@default ['/*']
cloudfrontInvalidationPaths: string[],
@default CacheHeaders.DEFAULT
cacheHeaders: {[path: string], header: string}
```
> Note: if the cache headers from Remix will override these, then we may be able to just document that, i.e. `export function headers() {}` will override the default cache control headers [link]

## Steps

 - [ ] Research: how does the team feel about making these options configurable
 - [ ] Research: what's the best API for configuring the cache headers and invalidation options
 - [ ] Implement: make these configurable within the construct
 - [ ] Test: ensure the props output the correct CloudFormation
 - [ ] Doc: add examples and API information for configuring the cache headers and cloudfront invalidation

bitpavel:
Hello, this feature would be extremely useful for us. Our platform has millions of pages that are updated rarely and quite heavy to be generated on the fly. 
We would like to add the cache on FE+CloudFront side (+ enable CloudFront Origin Shield) and use ISR for refreshing the data in background. But since CloudFront cache is invalidated after every new deployment, it isn't acceptable for us.
For now, we use a self-developed workaround to manage this cache (to store and refresh it). But we'll be happy to migrate to this new functionality, when SST provides the possibility to mange invalidation paths.

jayair:
We'll accept PRs for this but let me understand what you are looking for.

You don't want some paths to be invalidated after every deployment?

How are you invalidating them (ISR) currently?

bitpavel:
@jayair Right now we have an nginx-based proxy in front of our SST-based application. We're in the middle of migration from our old FE application to the new NextJS-based version and use this reverse proxy for routing. We've implemented the durable caching at this reverse proxy. 
We understand that it's a bad solution because it eliminates benefits of using CloudFront - CDN capabilities. So once we finish the migration (hope in couple months), we would like to remove this reverse proxy layer and route users directly to SST-based application. In this case, we will lose the caching capability as well.
Good to hear that you're open for PR, it's a working option for us.

bitpavel:
Hey @jayair, I've create a PR: https://github.com/sst/sst/pull/3428
How can we proceed with it?

thdxr:
[...]

PR #3428, opened 2023-10-18
SsrSite: Add possibility to define custom CloudFront invalidation paths

bitpavel:
According to the discussed in this issue: https://github.com/sst/sst/issues/2267

This PR implements possibility to define custom paths for CloudFront invalidation. It introduces a new `defaultInvalidationPaths` parameter, that accepts array of strings with paths. By default, `[/*]` is defined.
Besides adding a new parameter, I've added:
- paths deduplication to reduce number of prefixes when possible
- handler for CloudFront limits. When the limits are exceeded, paths list is spliced into chunks and these chunks are invalidated as individual requests


fwang:
Thanks @bitpavel!

I added an `invalidation` prop, and moved `defaultInvalidationPaths` to `invalidation.paths`.

bitpavel:
Thank you @fwang

PR #3432, opened 2023-10-19
Sites: rewrite S3 deployment

fwang:


packages/sst/support/custom-resources/s3-uploader.ts
```diff
@@ -0,0 +1,189 @@
...
+  // Upload to S3
+  const filesUploaded: string[] = [];
+  for (const fileOption of fileOptions) {
+    const files = globSync(fileOption.files, {
+      cwd: contentsDir,
+      nodir: true,
+      ignore: fileOption.ignore,
+    }).filter((file) => !filesUploaded.includes(file));
+    for (const file of files) {
+      await s3.send(
```
bayssmekanique, line 176: I suspect between the file I/O and whatever networking is involved in the `s3.send`, this operation would probably greatly benefit from utilizing a bulkhead with maybe 3-5 concurrent requests and changing to an async `readFile`.

packages/sst/src/constructs/StaticSite.ts
```diff
@@ -674,96 +660,65 @@ interface ImportMeta {
...
-                ])
-              )
-              .flat(2),
-            cacheControl ? ["--cache-control", cacheControl] : [],
-            contentType ? ["--content-type", contentType] : [],
-            contentEncoding ? ["--content-encoding", contentEncoding] : [],
-          ].flat();
-        }),
-        ReplaceValues: this.getS3ContentReplaceValues(),
+        fileOptions: [
```
bayssmekanique, line 705: The `fileOptions` are parsed first to last so the specificity of the `files` glob should be most specific to least specific (reverse of current implementation).

packages/sst/src/constructs/SsrSite.ts
```diff
@@ -1217,96 +1233,52 @@ function handler(event) {
         zip: { mime: "application/zip", isText: false },
       };
 
-      copy.forEach((files) => {
-        if (!files.cached) return;
+      copy.forEach(({ cached, to, versionedSubDir }) => {
+        if (!cached) return;
+
+        // Create a default file option for: unversioned files
```
bayssmekanique, line 1239: Both the `versionedSubDir` and `nonVersionedSubDir` patterns should be added to the end of the `fileOptions`.

packages/sst/src/constructs/StaticSite.ts
bayssmekanique: Can be resolved by reversing file order.

PR #3435, opened 2023-10-20
[Construct/SSR]: Add optional viewerProtocolPolicy to SSR Functions

alexabidri:
This PR adds an optional viewerProtocolPolicy parameter to the SSR functions. This parameter allows the default value for cloudfront functions deployed with `REDIRECT_TO_HTTPS` to be overridden.

To use this new parameter, simply pass in a string value of either `https-only` or `redirect-to-https`. If no value is provided, the default behavior of redirect-to-https will be used.

This change should make it easier to customize the behavior of the SSR functions for different deployment scenarios.


fwang:
Thanks @alexabidri!

`cdk.distribution.defaultBehavior.viewerProtocolPolicy` should get applied only to the default behavior. Moved it to `cdk.viewerProtocolPolicy`.

PR #3438, opened 2023-10-23
fix: support warming requests in remix server

michaelgmcd:
Open to suggestions on a better way to do this, but
https://github.com/sst/sst/blob/master/packages/sst/support/remix-site-function/regional-server.js#L84
throws an error when it receives a request from a warming function (event.headers is undefined)

### DIFF
--- a/packages/sst/support/remix-site-function/regional-server.js
+++ b/packages/sst/support/remix-site-function/regional-server.js
@@ -153,6 +153,16 @@ const createApigHandler = (build) => {
   const requestHandler = createNodeRequestHandler(build, process.env.NODE_ENV);
 
   return async (event) => {
+    if (event.type === "warmer") {
+      return {
+        statusCode: 200,
+        headers: {},
+        cookies: [],
+        body: JSON.stringify({ message: "warmed" }),
+        isBase64Encoded: false,
+      };
+    }
+
     const request = convertApigRequestToNode(event);
     const response = await requestHandler(request);
     return convertNodeResponseToApig(response);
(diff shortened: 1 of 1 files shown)


fwang:
Thanks @michaelgmcd.

The warmer function expects a predefined return response format — https://github.com/michaelgmcd/sst/blob/98db2fae5a7c84cf0875f62c5188bfb05a58a362/packages/sst/support/ssr-warmer/index.ts#L69

Also added a `setTimeout` to ensure the warming requests takes some time to respond, so a lambda instance don't end up handling multiple warming requests.

Can you give it a try and see if it works?

packages/sst/support/remix-site-function/regional-server.js
```diff
@@ -153,10 +153,22 @@ const createApigHandler = (build) => {
...
     const request = convertApigRequestToNode(event);
     const response = await requestHandler(request);
     return convertNodeResponseToApig(response);
   };
 };
 
+function formatWarmerResponse(event) {
+  return new Promise((resolve) => {
+    setTimeout(() => {
+      resolve({ serverId });
```
michaelgmcd, line 169: `serverId` here is undefined. Looking through the code, I don't see where it's supposed to come from. Is it `event.warmerId`?

jayair:
Pinging Frank. Feel free to mention me later in the week if he doesn't respond.

fwang:
Thanks @michaelgmcd I ended up wrapping the lambda function to handle the warming request - https://github.com/sst/sst/commit/391e8f83ea5eda20fc6e8e555d4aa62ba1c0ab71

This should fix it for Remix and other sites.

Fixed in v2.36.6

PR #3452, opened 2023-10-27
sst: Create a `configPrefix` global option for the dev command

mikeki:
I'm working on a monorepo that will define different configs in the same directory, and I noticed that we only allow 1 config.

This change will provide the option of specifying a prefix to the config file:
e.g.
```
sst dev --configPrefix web/
```
The previous command would look for the config at `web/sst.config.ts` instead of only ever having the option of `sst.config.ts`.

This way we can run the `sst dev` command from the root level and work on any sub-package

### DIFF
--- a/packages/sst/src/cli/program.ts
+++ b/packages/sst/src/cli/program.ts
@@ -23,11 +23,15 @@ export const program = yargs(hideBin(process.argv))
     type: "string",
     describe: "ARN of the IAM role to use when invoking AWS",
   })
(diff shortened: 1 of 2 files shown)


jayair:
Hmm this is an interesting pattern. Can you open a help thread in Discord? I've got a couple of questions.

mikeki:
@jayair I don't have a discord account, let me create one in the next couple days so that we can discuss.

thdxr:

Issue #3451, opened 2023-10-27
`sst bind` fails on stacks containing a Service with bound constructs


zvictor:
_Discord discussion: https://discord.com/channels/983865673656705025/1166372275185000598_

---

## Description:

When running the `sst bind` command on a stack that contains a `Service` with constructs bound through the `service.bind` call, a TypeError is thrown: `Cannot read properties of null (reading 'addEnvironment')`. This error seems to occur because `this.container` is `null` when running locally.

```haskell
Error: Cannot read properties of null (reading 'addEnvironment')

Trace: TypeError: Cannot read properties of null (reading 'addEnvironment')
    at Service.addEnvironmentForService (file:///node_modules/.pnpm/sst@2.32.1_@types+react@18.2.16/node_modules/sst/constructs/Service.js:567:24)
    at file:///node_modules/.pnpm/sst@2.32.1_@types+react@18.2.16/node_modules/sst/constructs/Service.js:554:64
    at Array.forEach (<anonymous>)
    at file:///node_modules/.pnpm/sst@2.32.1_@types+react@18.2.16/node_modules/sst/constructs/Service.js:554:33
    at Array.forEach (<anonymous>)
    at Service.bindForService (file:///node_modules/.pnpm/sst@2.32.1_@types+react@18.2.16/node_modules/sst/constructs/Service.js:551:47)
    at Service.bind (file:///node_modules/.pnpm/sst@2.32.1_@types+react@18.2.16/node_modules/sst/constructs/Service.js:323:14)
    at EmptyStack.WhatsappListenerStack (file:///deployment/.sst.config.1698353557905.mjs:599:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.stacks [as fn] (file:///deployment/.sst.config.1698353557905.mjs:630:5)
    at async synthInRoot (file:///node_modules/.pnpm/sst@2.32.1_@types+react@18.2.16/node_modules/sst/stacks/synth.js:56:13)
    at async Module.synth (file:///node_modules/.pnpm/sst@2.32.1_@types+react@18.2.16/node_modules/sst/stacks/synth.js:19:16)
    at async buildApp (file:///node_modules/.pnpm/sst@2.32.1_@types+react@18.2.16/node_modules/sst/cli/commands/bind.js:84:13)
    at async Object.handler (file:///node_modules/.pnpm/sst@2.32.1_@types+react@18.2.16/node_modules/sst/cli/commands/bind.js:51:9)
```

[...]

PR #3453, opened 2023-10-29
fix: Bind to service only when deployed

zvictor:
Possible fix to #3451. Please check #3451 for context.

### DIFF
--- a/.changeset/five-bananas-sell.md
+++ b/.changeset/five-bananas-sell.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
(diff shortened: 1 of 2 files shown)


jayair:
Hmm let me pass this along.

zvictor:
@jayair how can I make this PR better?

I have solved the problem in the way that seemed correct, but I don't have the full execution context in mind. I'm open to changing the solution if given more guidance.

fwang:
Thanks @zvictor! I corrected the underlying types to catch for the issue.

PR #3454, opened 2023-10-29
Add alb cdk property for Service construct

suil:
In the case where `cloudfrontDistribution` is set to be `false`, we need a way to get alb's domain name. This requires to expose alb as a property under `cdk` for the Service construct


jayair:
Thanks!

fwang:
Thanks @suil! Will be in the next release.

PR #3456, opened 2023-10-30
fix: Edge Function deployment for windows bug

### LINKED ISSUE #3455: bug: EdgeFunction construct fails when deployed from Windows
The handler path for the EdgeFunction entrypoint is invalid with deployed from a Windows machine, likely because a `path.join` or similar function is used to generate the path.

Invalid path:
Screenshot 2023-10-29 165958

Resulting error:
Screenshot 2023-10-29 170022

bayssmekanique:
Normalize `handler` path to posix for all environments to correct path issue when deploying from Windows.

Resolve #3455


jayair:
Thanks!

PR #3459, opened 2023-10-30
sst/examples/create-sst-rds: Add dynamic JSDoc database type to migration template

stevenschmatz:
# The problem

The `migration.ejs.t` file uses JSDoc to provide some autocompletion to the `db` type, but it doesn't make use of the auto-generated types in `sql.generated`.

This has led to some folks in the SST Discord (and myself) trying to set up TypeScript compilation for these files. But there is a simpler solution.

# The solution

JSDoc supports imported types from other files. This makes for some really nice autocompletion without requiring a separate build step or having to worry about anything happening with runtime of the migration itself.

CleanShot 2023-10-30 at 17 30 19@2x


jayair:
Oh that's interesting. Thanks for tracking down.

thdxr:

PR #3461, opened 2023-10-31
fix: verifier on future/auth requires AUTH_ID

justindra:
When using `future/auth` and using `session.use`, the publicKey is never returned as `process.env.AUTH_ID` is no longer defined. It used to be defined in previous auth through `auth.attach` which is no longer in the new version.

I thought it would be nicer to have it as part of the binding instead. So that we can just do the following and it should work:

```ts
new Api(stack, 'api', {
defaults: {
      function: {
        bind: [auth],
      },
    },
})
```

instead of currently which would be:

```ts
new Api(stack, 'api', {
defaults: {
      function: {
        environment: {
          AUTH_ID: auth.id,
        },
      },
    },
})
```

### DIFF
--- a/packages/sst/src/constructs/future/Auth.ts
+++ b/packages/sst/src/constructs/future/Auth.ts
@@ -194,6 +194,10 @@ export class Auth extends Construct implements SSTConstruct {
           type: "plain",
           value: this.url,
         },
+        authId: {
+          type: "auth_id",
+          value: this.id,
+        },
       },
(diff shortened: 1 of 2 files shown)


jayair:
Thanks!

justindra:
@jayair just wondering if there's anything I can do to get this merged in?

jayair:
Oh let me ping Dax again.

PR #3463, opened 2023-11-01
Added WASM mime type to S3 uploader

wpoynter:
The recent change to the S3 uploader script dropped support for WASM files by default. I believe adding this line in to the extensions array will address the issue.


jayair:
Thanks! Reviewing.

fwang:
Merged, will be in the next release.

Thanks @wpoynter!

PR #3467, opened 2023-11-02
Added some more details on creating a Python Lambda handler.

rupertlssmith:
Added some important details needed to deploy some Python code as a Function.


jayair:
Thanks!

PR #3472, opened 2023-11-03
Fix source maps documentation 

RafalWilinski:
`bundle` is not a valid property on `FunctionProps` object. Instead, there's `nodejs`


jayair:
Ah yes

Issue #2532, opened 2023-02-21
AppSync with multiple co-dependent schemas


melissakintz:
Hello,
I would like to set up an appsync api, by splitting my graphql schemas, however I encounter the following problem:

I specify my different schemas in an array, no merge problem, but as soon as I want to add a dependency from one schema to another, an error throw like the type that comes from another schema does not exist.

Is there an other solution than all in the same graphql file? 

First schema extract: 
```graphql
type Query {
  brand(slug: String, brandId: ID): Brand
}

type Brand {
  id: ID!
  name: String!
  description: String
  slug: String!
  offers: [Offer]
}
```

Second schema extract : 
```graphql
type Offer {
  id: ID!
  name: String!
  link: String
  brandId: String!
  createdAt: AWSDate!
}
```

Throwed error : `Unknown type "Offer"`

AppsyncApi : 
```ts
  const api = new AppSyncApi(stack, "AppSyncApi", {
    schema: [
      "src/server/appsync/graphql/offer/schema.graphql",
      "src/server/appsync/graphql/brand/schema.graphql"
    ],
    dataSources: {
      brands: "appsync/functions/brand.handler",
      offers: "appsync/functions/offer.handler"
    },
    resolvers: {
      "Query    brand": "brands",
      "Brand    offers": "offers"
    }
  });
```

galfrevn:
Facing the same issue here!

melissakintz:
I worked around the problem by merging the schema myself to indicate a single file already merged to the api, using : 

```ts
  import { loadFilesSync } from "@graphql-tools/load-files";
  import { mergeTypeDefs } from "@graphql-tools/merge";
  import { print } from "graphql";
  import * as fs from "fs";

  {...}

  const loadedFiles = loadFilesSync(`src/server/appsync/graphql/**/*.gql`);
  const typeDefs = mergeTypeDefs(loadedFiles);
  const printedTypeDefs = print(typeDefs);
  fs.writeFileSync(`src/server/appsync/graphql/generated.graphql`, printedTypeDefs);

  const api = new AppSyncApi(stack, "AppSyncApi", {
    schema: ["src/server/appsync/graphql/generated.graphql"],
    {...}
   }
```

Issue #2104, opened 2022-09-28
AppSyncApi v1 construct doesn't support AWS directives


smacker:
Discovered it during migration from v0 to v1.

**Steps to reproduce:**

Create 2 files, e.g. `mutations.graphql` and `queries.graphql`. Use any [AWS directives](https://docs.aws.amazon.com/appsync/latest/devguide/security-authz.html) such as `@aws_iam`, `@aws_cognito_user_pools`, `@aws_api_key` or [scalars](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html) such as `AWSDate` or `AWSJSON`.

Pass schemas as an array:

```ts
new sst.AppSyncApi(this, `AppSyncApi`, {
      schema: [
        'schema/mutations.graphql',
        'schema/queries.graphql',
      ],
}
```

Run `sst build/start/deploy`.

Result:
```
Error: Unknown directive "@aws_iam".

Unknown directive "@aws_cognito_user_pools".

Unknown directive "@aws_api_key".

Unknown type "AWSDateTime".

    at assertValidSDL (/some/path/node_modules/graphql/validation/validate.js:107:11)
    at buildASTSchema (/some/path/node_modules/graphql/utilities/buildASTSchema.js:45:34)
    at buildSchema (/some/path/node_modules/graphql/utilities/buildASTSchema.js:112:10)
    at Array.map (<anonymous>)
    at AppSyncApi.createGraphApi (file:///some/path/node_modules/@serverless-stack/resources/dist/AppSyncApi.js:242:26)
    at new AppSyncApi (file:///some/path/node_modules/@serverless-stack/resources/dist/AppSyncApi.js:48:14)
    at new GraphQlStack (file:///some/path/server/.build/lib/index.js:83:23)
    at Module.main (file:///some/path/server/.build/lib/index.js:533:24)
    at file:///some/path/server/.build/run.mjs:96:22
```

**Problem:**

When schema is defined in more than 1 file `sst` processes them with [graphql.buildSchema](https://github.com/serverless-stack/sst/blob/b51cd3ac640eddb73fb391e57a0e313f44747826/packages/resources/src/AppSyncApi.ts#L686).

This function doesn't know about AWS directives or scalar types like `@aws_iam`, `@aws_cognito_user_pools`, `@aws_api_key`, `AWSDateTime`.

Current workaround is to combine all schemas into a single file and defined schema as a string instead of an array in the construct to bypass `graphql.buildSchema`.

[...]

PR #3475, opened 2023-11-03
AppSyncApi: Improve multiple schema functionality.

BradDobson2:
Currently, when using the `AppSyncApi` construct with multiple schema files, there's a few limitations due to passing each individual schema into the `buildSchema` function from `@graphql-tools/merge`.

- Sharing common types and enums etc between multiple schemas isn't possible as `buildSchema` treats each schema file separately
- `buildSchema` does not allow AppSync directives such as `@aws_iam`

Removing the call to `buildSchema` will fix both of the above limitations.

Related issues:
- https://github.com/sst/sst/issues/2104
- https://github.com/sst/sst/issues/2532

### DIFF
--- a/.changeset/four-dingos-greet.md
+++ b/.changeset/four-dingos-greet.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+AppSyncApi: improve multiple schema functionality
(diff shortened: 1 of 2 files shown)


jayair:
Thanks, reviewing.

fwang:

Thanks @BradDobson2!

Issue #3479, opened 2023-11-03
Providing CDK id to Auth construct throws error


jackschu:
The docs [mention](https://docs.sst.dev/constructs/Auth#cdkid) you can provide a cdk id to the auth construct.

But doing so results in an error during `sst build`

`It looks like you may be passing in Cognito props to the Auth construct. The Auth construct was renamed to Cognito in version 1.10.0`

(mostly filing this as a token issue as i open a PR b/c thats what the contributing guidelines suggest)

PR #3480, opened 2023-11-03
Auth: Prevent throwing on CDK id

jackschu:
The docs [mention](https://docs.sst.dev/constructs/Auth#cdkid) you can provide a cdk id to the auth construct.

But doing so results in an error during `sst build`

`It looks like you may be passing in Cognito props to the Auth construct. The Auth construct was renamed to Cognito in version 1.10.0`

Seems like cdk id was added as a prop after this throw and the throw was never cleaned up.

fixes: #3479

### DIFF
--- a/.changeset/poor-cheetahs-approve.md
+++ b/.changeset/poor-cheetahs-approve.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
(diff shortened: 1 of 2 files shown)


jayair:
Thanks taking a look.

PR #3484, opened 2023-11-05
Update default NodeJS runtime to 18.x

adityavm:
Solves the issue of default NodeJS being set to 16.x (causes `fetch` to fail since it was [introduced in 18.x](https://nodejs.org/en/blog/announcements/v18-release-announce#fetch-experimental))


fwang:

Thanks @adityavm!

PR #3485, opened 2023-11-05
Fix `No path found at` for StaticSite Construct

suil:
For building StaticSite, `buildCommand` will create a build folder. so we should check the existence of the build folder after running the `buildCommand`.


packages/sst/src/constructs/StaticSite.ts
```diff
@@ -525,15 +525,6 @@ interface ImportMeta {
...
 
-    // validate site path exists
-    if (!fs.existsSync(sitePath)) {
-      throw new Error(
-        `No path found at "${path.resolve(sitePath)}" for the "${
-          this.node.id
-        }" StaticSite.`
-      );
-    }
-
```
fwang, line 536: hey @suil this is checking the path to the site, not the output generated by the build command.

thdxr:

PR #3493, opened 2023-11-06
feat: Astro Assets via Sharp by default

bayssmekanique:
This PR adds the automatic creation of a Sharp layer for `AstroSite` deployments using Sharp for image processing. As part of this update, the `squoosh` compatibility has been dropped from the adapter as there is not a readily available Squoosh layer pakcage.

This PR also fixes the issue of the adapter export (`aws()`) not having types due to the `types.d.ts` definition file not being included in the `dist` output.

All changes in this PR should be backwards compatible and all configurations can be overwritten.


packages/sst/src/constructs/AstroSite.ts
```diff
@@ -174,6 +177,56 @@ export class AstroSite extends SsrSite {
           origin: "staticsServer",
         });
       } else {
+        if (buildMeta.imageService === "sharp") {
+          const pathToSharpLayerZip = createRequire(import.meta.url).resolve(
+            "astro-sst/layers/sharp"
+          );
```
fwang, line 183: Does it make sense to have `astro-sst` spit out sharp into `dist`, and `AstroSite` can find it at `join(sitePath, "dist", "image", "sharp")`.
```
my-astro-app/
  dist/
    server/
    client/
    image/
```

bayssmekanique: I had considered this and was part of the way to implementing it that way but I decided it was doing a lot of work for no benefit. The layer is not exactly tiny (15.6 MB), so copying it over on every rebuild seems a bit excessive. Also, the phase when it would be copied over happens before the final CDK evaluation so the layer could be moved over, but then not used because a different server config or layer is used. It just seemed a bit cleaner to leave it in place.

All that being said, it does feel odd to be deploying a file directly from an `npm` package.

fwang: Makes sense.

Taking a step back, currently when building `AstroSite`'s server function, SST already detects if `sharp` is being used. If detected, `sharp` gets automatically packaged into the function bundle.
https://github.com/sst/sst/blob/b79a68a3d1f01a6983337cd950ba1b7ae838a045/packages/sst/src/runtime/handlers/node.ts#L209-L219

ie. I just ran `sst build`, and the server bundle has a `node_module` folder w/ `sharp` in it
Screen Shot 2023-11-06 at 7 26 58 PM


I don't think we need the layer?

bayssmekanique: The layer does still help for startup time (supposedly), but the biggest thing is that it prevents a lot of really common deployment issues because the runtime binaries for `sharp` need to be Linux and most developers are using OSX or Windows, so this is a common issue that results in devs uninstalling and reinstalling packages before each deployment, and reverting again afterwards.

I'd certainly like to avoid any duplication of code/resources on the instance, so it's worth looking into further. I've tested and confirmed that both the layer and the `nodejs.esbuild.external["sharp"]` declaration are necessary. Removing either or both results in 500 with an error about not finding `sharp`.

bayssmekanique: @fwang, you are correct that the `sharp` dependencies were being built into the bundle, but they weren't working in many cases because the architecture of the binaries did not match the runtime.

I've fixed this part of this PR, so now both layers and runtime bundles are supported.

bayssmekanique:
This PR has wondered through the capability space and has finally settled on adopting a more dynamic and DX friendly approach with an easy "advanced" user path via the docs.

- Adds support for `sharp` image process for `regional` and `edge` deployments without any config changes.
- Adds docs for how to add a layer to `regional` deployments for a more optimized runtime.
- Adds `architecture` as a `SsrSite` prop with `arm_64` as the default architecture for SSR functions.
- Removes `squoosh` as a supported asset optimizer from the adapter.

fwang:
@bayssmekanique oops I forgot to send this after typing it up

From what I gathered so far (I might be wrong), 3 ways to get Sharp to work:
1. Use the bundled version as in: with ur fix to respect the architecture, the bundled version should just work. Is the downside having to `npm i sharp` on every deploy?
2. Bundle a pre-existing version of `sharp` from `astro-sst` as a `node_modules` folder. Is the downside having to upload this `node_modules` folder on every deploy?
3. Bundle a pre-existing version of `sharp` from `astro-sst` as a layer. There were some thoughts in the OpenNext community that the image optimizer should be deployed to the edge. It is not right now b/c deploying an edge function is slow in CloudFront (takes minutes). But I want to keep this a possible by not using a layer. Also using layer doesn't actually reduce the cold start.

I'm leaning towards 1 or 2 more.

bayssmekanique:
@fwang,

1) Bundled version (with this fix): *This is my preference for a default.* It works for regional and edge deployment, doesn't introduce dependency locking, and has a clean path for overriding. The only downside to this is there is a lot of extra junk from the `npm` package that doesn't get used by the runtime but is still loaded into the environment. I believe `npm` should cache the dependency in the local cache so it shouldn't have to download on each rebuild (just on first build).

2) Bundled...?: I'm not entirely sure how this differs from (1). Sounds like the same as but with a bundled version in `astro-sst`. I would avoid doing anything that adds dependencies when not necessary as they will eventually cause headaches when they fall too far behind. That's also my thinking for resisting the layers approach.

3) Bundling a pre-built layer: This was my initial implementation strategy and in theory it should be more performant, not because of anything intrinsic to the layer, but because the prebuilt layers have been stripped of all "fluff" so it makes the entire deployed code smaller. This, however, introduces a version dependency within the `astro-sst` which will require maintenance and it doesn't work for edge deployments.  For these reasons, I think it makes more sense to treat this as the "advanced" strategy made available through the documentation but not a default configuration.

All in all, I think strategy (1) is the best approach, and that's basically what's represented in this PR. It works on the most environment and has the most maintainable path forward. Bundling in a version of `sharp` might be more performant from a build and possibly even runtime standing, but it will cause headaches eventually when there is a demand for an update to the bundled `sharp` version but the update will break backwards compatibility so you end up doing a major release just for a bundled dependency.

packages/astro-sst/src/adapter.ts
```diff
@@ -88,6 +87,7 @@ export default function createIntegration(
             updateConfig({
               output: "server",
             });
+            config.output = "server";
```
bayssmekanique, line 90: FWIW, this is ugly and I have a PR with Astro open to make it so the config can be returned after an update to the config: https://github.com/withastro/astro/pull/9013

fwang:
@bayssmekanique I'm a bit hesitate on adding a top level `SsrSite.architecture` prop, as it's not supported by L@E (as u mentioned in the comment). Currently, you can set `SsrSite.cdk.server.architecture` and that only applies to the server function. So I pushed a change to use `SsrSite.cdk.server.architecture` instead.

Let me know if it looks good, and I will merge it.

bayssmekanique:
@fwang, I understand the concern, though I would expect that before long there will likely be support for alternate runtime architectures on L@E.

These reversions nullify the changes that fixed the asset handler. The issue is that without an architecture defined somewhere in the `SsrFunction` scope the bundler doesn't know which architecture to use and uses the default (which on my machine is `x86_64`) which doesn't necessarily match the default configuration used by Lambda (which at the moment is `arm_64`).

I've pushed another commit which essentially meets in the middle but adding a top level default to the `SsrFunction` of `ARM_64` while still allowing override. This will give the bundler something concrete to work off instead of hoping the defaults align.

jlarmstrongiv:
I just merged Astro _image endpoint support to unpic in https://github.com/ascorbic/unpic/pull/94, which means as soon as this PR lands, we’ll have support for Astro assets in all ui frameworks (react, preact, svelte, vue, solidjs, lit, qwik) with SST.  Combined, this will be a game changer 😄

PR #3500, opened 2023-11-07
Make sourcemap order in CF templates deterministic

rubythulhu:
When multiple lambdas in a stack have sourcemaps, the order of these sourcemaps within the CF template changes between runs. This means that:

* `sst diff` will show these as changes 
* `sst deploy` will needlessly redeploy these
* `sst start` also takes longer to fully deploy its stubs

Sorting the list of sourcemaps fixes this problem. This fix can be either applied inside `constructs/Function.ts` (in `useFunctions().sourcemaps.forStack()`) or `constructs/App.ts` (during the creation of the SourcemapUploader CustomResource). It seems cleaner to me to fix it in Function.ts, so this is the approach I took.

### DIFF
--- a/.changeset/nervous-rats-invite.md
+++ b/.changeset/nervous-rats-invite.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Make sourcemap ordering deterministic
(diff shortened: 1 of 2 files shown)


jayair:
Hmm interesting. Let me share it with the team.

rubythulhu:
Small update:

Using `srcKey` as the sorting key for this is insufficient. Changes in lambdas will change this hash, which will change the ordering, causing more false changes.

Instead this PR now uses `props.handler` as a sort key, which should keep the order deterministic in more cases.

PR #3504, opened 2023-11-10
WebSocketApi: Support returnResponse in route definitions

jaydeebee:
We found need for two-way communications with our WebsocketApi's, and this seems to do the trick.  CDK had the wiring already, this PR exposes it via the routes.  Example:
```typescript
      const websocket = new WebSocketApi(stack, 'websocket', {
        routes: {
          echo: {
            function: {
              handler: 'server/websocket/echo.func'
            },
            // new:
            returnResponse: true
          },
        }
      });
```

### DIFF
--- a/.changeset/lazy-starfishes-lie.md
+++ b/.changeset/lazy-starfishes-lie.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+WebSocketApi: support route returnResponse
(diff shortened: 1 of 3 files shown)


jayair:
Oh interesting. Thank you!

fwang:

Thanks @jaydeebee!

PR #3505, opened 2023-11-12
Use new kyesly website URL in docs

ggrochow:
Noticed some links weren't taking me to the correct pages, google showed some fancy new kysely site. Lets use that.

Replace all usages of the two old kysely URLs in documentation
[1](https://kysely-org.github.io/kysely/) [2](https://koskimas.github.io/kysely/#migrations)

With the new website found at https://kysely.dev/


jayair:
Nice! Cool new website!

PR #3507, opened 2023-11-14
Add _events and cb to the callback function in EventHandler

tetranoir:
Add _events and cb to the callback function in EventHandler to make reflection easier.

### DIFF
--- a/packages/sst/src/node/event-bus/index.ts
+++ b/packages/sst/src/node/event-bus/index.ts
@@ -146,7 +146,7 @@ export function EventHandler<Events extends Event>(
     }[Events["type"]]
   ) => Promise<void>
 ) {
-  return async (
+  const fn = async (
     event: EventBridgeEvent<string, any> & { attempts?: number }
   ) => {
     await cb({
@@ -156,4 +156,7 @@ export function EventHandler<Events extends Event>(
       attempts: event.attempts ?? 0,
     });
   };
+  fn._events = _events;
+  fn.cb = cb;
+  return fn;
 }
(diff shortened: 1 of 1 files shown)


thdxr:
can you achieve what you need to by wrapping the EventHandler function? this does feel like a bit of a hack to put in

tetranoir:
Yea, this is what I have now
```
/**
 * @file Drop in replacements for SST's EventHandler and
 * an event bus createEventBuilder instance.
 */
import * as SST from 'sst/node/event-bus';
// Extracted SST types
import { SstEvent, SstEventCallback, SstEventHandler } from './sst-types.js';
// My custom publishEvent function
import { publishEvent } from '@local-lib/local-event-bus.js';


/**
 * Proxies SST EventHandler. Adds a `events` to the return object.
 * @param _events uses _ because that's what SST.EventHandler calls it.
 */
export const EventHandler = <E extends SstEvent>(
  _events: E[],
  cb: SstEventCallback<E>,
): ReturnType<SstEventHandler<E>> & { events: E[], cb: SstEventCallback<E> } => {
  const events = Array.isArray(_events) ? _events : [_events];

  return Object.assign(SST.EventHandler<E>(events, cb), { events, cb });
};

/**
 * Use this to define events that mean to work with SST event bus.
 */
export const createEvent: ReturnType<typeof SST.createEventBuilder> = (name, shape) => {
  const event = SST.createEventBuilder({
    bus: 'Bus', // Can't use the dynamic resource name EventBus.Bus.eventBusName.
  })(name, shape);

  // If local, monkeypatch publish to instead resolve locally.
  // TODO: BAD
  if (process.env.EVENT_BUS === 'local') {
    event.publish = async (props) => {
      publishEvent(event, props);
      // Mocks AWS PutEventsCommandOutput
      return {
        $metadata: { httpStatusCode: 200 },
        FailedEntryCount: 0,
        Entries: [],
      };
    };
  }

  // Leaving as any because the generics in createEventBuilder don't transfer well.
  return event as any;
};
```

thdxr:

PR #3513, opened 2023-11-17
sst/NextjsSite: Update NextjsSite.about.md

byF:
When we applied the config in its original form, the Next dev mode reported:

> Warning: Reverting webpack devtool to 'eval-source-map'. Changing the webpack devtool in development mode will cause severe performance regressions.

The Next.js documentation suggests to avoid changing the devtool in dev mode: https://nextjs.org/docs/messages/improper-devtool


jayair:
Oh I see. Thank you!

PR #3520, opened 2023-11-22
Add Lambda URL stream option - InvokeMode

jlanzarotti:
Adding bool attribute `stream` to `FunctionUrlProps`  to enable streaming response.

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.InvokeMode.html


jayair:
Ah I like it. I'll share internally.

fwang:
Thanks @jlanzarotti!

Renamed the prop to `streaming` and added docs.

Will be in the next release.

PR #3525, opened 2023-11-23
Appsync: Fix Custom Domain Certificate Region

dgocoder:
Currently support for app sync custom domain appears to be broken. App sync requires a cert to be created in us-east-1 due to cloudfront. I have changed the code to mirror the same functionality as cf distro dns creation.

AWS Docs: https://docs.aws.amazon.com/appsync/latest/devguide/custom-domain-name.html#appsynclong-configure-custom-domain

Docs state that "To set up a custom domain name as the API's hostname, the API owner must provide an SSL/TLS certificate for the custom domain name. To provide a certificate, do one of the following:

Request a new certificate in ACM, or import a certificate issued by a third-party certificate authority into ACM in the us-east-1 AWS Region (US East (N. Virginia))."

Here is the error I currently see when trying to use a custom domain.
image

Work around for now is to create cert separately and pass it down through the cdk prop or only deploy in us-east-1, however this is not as ideal.


fwang:
Thanks @dgocoder!

Tweaked it for backward compatibility. Will be in the next release.

PR #3526, opened 2023-11-23
docs: Update events.md to include reference to createEventBuilder

DianomiJH:
I couldn't find any reference to createEventBuilder anywhere in the docs site, and it took me spelunking in the Console repo to find out what i was missing creating Events

I don't really know the full ins and outs of how createEventBuilder worked so i added it as an example here rather than give a full section for it.

### DIFF
--- a/www/docs/events.md
+++ b/www/docs/events.md
@@ -60,7 +60,11 @@ In your application you can define events. This definition provides validation u
 
 
 ```ts title="packages/core/src/todo.ts"
-import { event } from "./events";
+import { createEventBuilder } from "sst/node/event-bus";
+
+const event = createEventBuilder({
+  bus: "bus",
+});
 
 export const Events = {
   Created: event("todo.created", {
@@ -68,6 +72,11 @@ export const Events = {
   }),
 };
 ```
+
+:::info
+Note that the `"bus"` string passed to the `createEventBuilder` is the same ID as provided in the EventBus construct in MyStack.ts above
+:::
+
 ---
 
 ## Define subscriber


DianomiJH:
Also, while i'm here, we seem to have found a weird edge case here when trying to publish multiple events in a Promise.all, see the discussion on [discord](https://discord.com/channels/983865673656705025/1177185534427271208). but it breaks down to creating the payload inside the async loop sends the same payload 3 times, but creating the payload and then looping to publish works fine.

```ts
// This example ends up executing the event handler lambda function 3 times with the same payload

async function publishEvent () {
  await Events.Created.publish({
    id: crypto.randomUUID()
  })
}

export async function produceEvents () {
  await Promise.all([
    publishEvent(),
    publishEvent(),
    publishEvent()
  ])
}
```

whereas this example works:

```ts
// this example works as expected, where the event handler lambda function runs 3 times, but each time with a different event

async function publishEvent (eventPayload) {
  await Events.Created.publish(eventPayload)
}

export async function produceEvents () {

  const eventPayloads = [
    { id: crypto.randomUUID() },
    { id: crypto.randomUUID() },
    { id: crypto.randomUUID() },
  ]

  await Promise.all(eventPayloads.map(payload => publishEvent(payload)))
}
```

thanks to @srodriki (can't find their github handle, sorry) for helping identify the issue 

@srodriki thought this was unlikely to be a SST issue, but i thought it might be nice to add a warning somewhere in this doc (or perhaps somewhere else?) explaining the edge case and how to avoid?

thdxr:

PR #3529, opened 2023-11-23
Add optional resource type to function binding

floriangosse:
Introducing an optional property called `resourceType` on function binding props so that we can explicitly define the resource type that should be used for type generation. This allows us to build custom SST constructs that can be injected into SST's type generation and variable population.

An example would be a custom database construct that can be bound to functions as well. Right now it would mean that we have to call the construct's class `RDS` just so that the types will be generated correctly and the variables are available via the `sst/node/rds` package. With the new property `resourceType` I can name my custom database construct however I want and can just set the `resourceType` in `getFunctionBinding` to `RDS` and the variables I define will be available via `sst/node/rds`.

### DIFF
--- a/packages/sst/src/constructs/Api.ts
+++ b/packages/sst/src/constructs/Api.ts
@@ -996,6 +996,7 @@ export class Api<
   public getFunctionBinding(): FunctionBindingProps {
     return {
       clientPackage: "api",
+      resourceType: 'Api',
       variables: {
         url: {
           type: "plain",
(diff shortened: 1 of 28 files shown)


jayair:
Oh this is interesting. This PR is a bit complicated but I like the underlying problem you are trying to solve. Let me share this with the team.

floriangosse:
I thought about a simple solution and I've decided to just use a `string` type. Another idea I had was to work with a enum or other predefined constants for `RDS`, `Function`, etc.

Let me know if I can help you with any more details.

floriangosse:
I found another thing that needs to be adjust to work properly: in case of the RDS construct there is the warmer that uses RDSDataAPI to talk to the database.

I've "misused" the RDS namespace for my own Database construct which isn't relying on the DataAPI. So, the warmer throws an error after some time.

There are probably more parts of the whole tooling that might not work as expected I don't know about it.

thdxr:

PR #3542, opened 2023-11-30
feat(service): support configuration of Fargate Task ephemeral storage

omikader:
This PR gives users of the `Service` construct the ability to configure the Fargate task [ephemeral storage](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/fargate-task-storage.html).

The pulled, compressed, and the uncompressed container image for the task is stored on the ephemeral storage. I would like to use a 3rd party image that is larger than the default 20 GB ephemeral storage. 

```
CannotPullContainerError: ref pull has been retried 1 time(s): failed to extract layer sha256:7c524a8642865c29df8d3973cefe8430653ba7fea85bd386e24e522601c30227: write /var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/86/fs/home/notebook-user/.local/lib/python3.10/site-packages/numpy/linalg/__pycache__/linalg.cpython-310.pyc: no space left on device: unknown
```

As a result, being able to configure this value would be awesome

### DIFF
--- a/.changeset/cuddly-kiwis-rhyme.md
+++ b/.changeset/cuddly-kiwis-rhyme.md
@@ -1,0 +1,6 @@
+---
+"sst": patch
+"@sst/docs": patch
+---
+
+feat(service): support configuration of ephemeral storage
--- a/.changeset/cuddly-kiwis-rhyme.md
+++ b/.changeset/cuddly-kiwis-rhyme.md
@@ -3,4 +3,4 @@
 "@sst/docs": patch
 ---
 
-feat(service): support configuration of ephemeral storage
+Service: support configuration of ephemeral storage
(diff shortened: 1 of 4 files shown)


omikader:
Good morning @jayair -- sorry to ping but would it be possible to have a look at this?

I'm having a lot of trouble manually adjusting the task storage settings in the AWS console without clobbering the service in SST.

Thank you again for all the hard work and support!

fwang:
Thanks @omikader!

Made a minor change. You can't actually set storage to 20 GB, the minimal allowed value by CloudFormation is 21. Have to leave the field undefined to set to 20. Weird.
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.CfnTaskDefinition.EphemeralStorageProperty.html#sizeingib

omikader:
@fwang oh good call, thanks! btw I think the test descriptions for the "too small" and "too large" cases are flipped 🙂

fwang:
lol good call. fixed

PR #3545, opened 2023-12-01
IoT: Fix connection closed after 5 minutes

coronapl:
Hi everyone,

My team and I have been encountering issues when running our SST project locally. After 5 minutes of not sending requests to our API, the lambda functions are no longer triggered. We identified the following logs:

```
2023-11-30T16:06:15.046Z +311599ms [debug] IoT error {"errno":-54,"code":"ECONNRESET","syscall":"read","type":"error" ...
2023-11-30T16:06:15.057Z    +11ms [debug] IoT closed
2023-11-30T16:06:16.058Z  +1001ms [debug] IoT reconnected
2023-11-30T16:06:17.460Z  +1402ms [debug] IoT connected
```

We came across discussions on Discord suggesting that the Node version might be a potential problem. However, we've updated all our Lambda functions to Node 18.x, and we're using the same version locally. We also experimented with the '--increase-timeout' flag, but this did not resolve the issue.

It appears that others are facing similar problems: [link](https://discord.com/channels/983865673656705025/1083422760111439973/threads/1095568175585837107)

I found a solution in a GitHub issue from the aws-iot-device-sdk-js repository: [link](https://github.com/aws/aws-iot-device-sdk-js/issues/280). Setting the keepalive option to 60 seconds seems to solve the problem. I tried it locally and was able to maintain a connection for hours without any issues. However, it would be great if someone else could also test this.

Here is what the SDK API documentation says about the keepalive option:

```
keepalive: used to specify the time interval for each ping request. The default is set to 300 seconds to connect to AWS IoT.
```


jayair:
Thank you for this!

PR #3554, opened 2023-12-05
fix: init project doesn't replace stage

justindra:
So I noticed that when calling `initProject` twice, it seems to replace some values but not all of it.

Here's how to re-create it:
```ts
import { useProject, initProject } from 'sst/project';

await initProject({ stage: 'prod' });
const project = useProject();

console.log(project);

await initProject({ stage: 'justin' });
const project2 = useProject();

console.log(project2);
```

What I see logged out is that everything is the same between the 2 stages except for the `config.profile`. But if we deleted the `SST_STAGE` env variable before initiating the project a 2nd time, it works as expected. I think if we are passing in a config, it should take precedent before the `SST_STAGE` env variable. Either way, it's weird that the profile changes but not the stage.

My use case is: I'm writing a script that let's me migrate data between different environments. E.g. Migrate User Xs data from prod to dev, so that I can debug the issue.

### DIFF
--- a/.changeset/calm-olives-joke.md
+++ b/.changeset/calm-olives-joke.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Calling init project does not replace stage if it was called previously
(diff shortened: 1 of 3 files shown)


jayair:
I'll send it to Dax

PR #3557, opened 2023-12-06
sst/cli: improve iot reconnect message

jon-spaeth:
update IoT reconnection message from `IoT reconnected` to `IoT reconnecting..` to better represent the event.


jayair:
Thanks!

PR #3566, opened 2023-12-07
fix: cjs being thrown in to provider

### LINKED ISSUE #3565: CommonJS in Node runtime
The recent Apple future/auth provider addition included some CommonJS for `require('querystring')` - just moved it to `import` syntax, was killing my local ESM-only set-up.

oyed:
closes #3565


jayair:
Thanks! Taking a look.

PR #3567, opened 2023-12-07
Open-next V3

conico974:
This PR add supports for open-next V3.
It should support all the features of open-next V3.
This PR also added some new things in the cloudfront function to improve cache hit ratio by a lot
It requires OpenNext v3, latest at the time of writing is `3.0.0-rc.13`

Here is a simple example to use this construct: 
```ts
import {NextjsSite} from "sst/constructs"

import type {OpenNextConfig} from "./open-next.config";

const site = new NextjsSite<OpenNextConfig>(stack, "site", {
  openNextVersion: "3.0.0-rc.13",
  logging: "combined", // per route logging is not implemented yet
  waitForInvalidation: false,
  cdk: {
    servers: {
      pageSsr: {
        cpu: "0.5 vCPU"
      },
      ssr: {
        memorySize: 512,
      },
      default: {
        memorySize: 1536,
      },
      middleware: {
        memorySize: 256
      }
    },
  
  }
});
```

This is mostly done, there is still some things that need to be figured out before release
- Do we want the external middleware to be in front of cloudfront ?
- ~~Right now it works with ECS as well as lambda, do we want to allow other deployment method (EC2, App Runner, EKS)~~ This has been removed, could be added again later
- Should we move more logic into `SSRSite` ?
- Reimplement per route logging
- ~~Makes warmer work with multiple lambda~~ This should work now, but it might have broken other framework (Not tested)


packages/sst/src/constructs/NextjsSite.ts
```diff
@@ -245,179 +302,200 @@ export class NextjsSite extends SsrSite {
...
+    return {
+      type: "function" as const,
+      constructId: `${key}ServerFunction`,
+      function: {
+        ...baseServerConfig,
+        handler: fn.handler,
+        bundle: path.join(sitePath, fn.bundle),
+        runtime: "nodejs18.x" as const,
+        architecture: Architecture.ARM_64,
+        memorySize: 1024,
```
khuezy, line 338: We should default the lambdas to 1536MB, see: https://docs.aws.amazon.com/lambda/latest/operatorguide/computing-power.html

tl;dr - 1536 MB is almost as cheap as 128 MB but you also get max speed.

khuezy (approved):
LGTM, I haven't deployed this latest version yet though.

packages/sst/src/constructs/NextjsSite.ts
```diff
@@ -38,7 +38,7 @@ import { SsrFunction } from "./SsrFunction.js";
 import { Asset } from "aws-cdk-lib/aws-s3-assets";
 import { useFunctions, FunctionProps } from "./Function.js";
 import { useDeferredTasks } from "./deferred_task.js";
-import { Logger } from "../logger.js";import { SsrContainer, SsrContainerProps } from "./SsrContainer.js";
+import { Logger } from "../logger.js"; import { SsrContainer, SsrContainerProps } from "./SsrContainer.js";
```
khuezy, line 42: New line?

packages/sst/src/constructs/NextjsSite.ts
matthew-gladman-oua: I don't believe that's strictly true. The AWS table provided is an example, they explain that by increasing memory you also increase vCPU given and that for CPU bound operations it will be cheaper as the function will run faster. 

But **if** it doesn't complete faster, it'll just be more expensive. That's why they've got a complicated tool to measure and profile your services ideal memory:
- https://docs.aws.amazon.com/lambda/latest/operatorguide/profile-functions.html

khuezy: That's fair. It's a bit tricky b/c there's been several people getting low latency from certain dependencies like prisma and increasing the memory helped minimize it. But everyone's mileage may vary.

packages/sst/src/constructs/NextjsSite.ts
```diff
@@ -161,17 +177,8 @@ export interface NextjsSiteProps extends Omit<SsrSiteProps, "nodejs"> {
...
-const DEFAULT_CACHE_POLICY_ALLOWED_HEADERS = [
-  "accept",
-  "rsc",
-  "next-router-prefetch",
-  "next-router-state-tree",
-  "next-url",
-  "x-prerender-bypass",
-  "x-prerender-revalidate"
-];
+const DEFAULT_OPEN_NEXT_VERSION = "3.0.0-rc.15";
```
conico974, line 180: I forgot about that, but we should bump this one to 3.0.1 before final release

fwang: yup!

fwang:
@conico974 is this getting called?https://github.com/conico974/sst/blob/cf85d5dd087f1e68436d7049fc324ca82ec84619/packages/sst/src/constructs/NextjsSite.ts#L249

I did a brief test and `x-open-next-cache-key` wasn't listed in allowed headers. It seems only `SsrSite.buildDefaultServerCachePolicyProps()` is called.

conico974:
> @conico974 is this getting called?https://github.com/conico974/sst/blob/cf85d5dd087f1e68436d7049fc324ca82ec84619/packages/sst/src/constructs/NextjsSite.ts#L249-L253
> 
> I did a brief test and `x-open-next-cache-key` wasn't listed in allowed headers. It seems only `SsrSite.buildDefaultServerCachePolicyProps()` is called.

Looks like you're right, I think we can just remove this function, and we'll need to pass the allowedHeaders into the plan directly

PR #3570, opened 2023-12-07
[DNM] feat: Reduce Astro CloudFront Function size and runtime

bayssmekanique:
Remove unused routes
Remove unused properties
Reduce strings to int enums where possible
Replace property names
Remove white-spaces


bayssmekanique:
Originally I reduced the matching set to only routes which were redirect or pages since those are the only routes where we do something to the request, but I didn't think about the fact that the matching works from the top down, so removing an endpoint or a page that was prerendered might cause the request to match for the wrong route.

bayssmekanique:
This is a very simple implementation that loops through all the routes until the first match is found. This works well for small sites, but can take a long time on bigger sites. For that reason, there's a PR with a more optimized approach which should be merged instead of this PR if testing is successful.

v2 PR: #3572

PR #3574, opened 2023-12-08
[SST] patch container to allow live lambda on WSL

dylanirion:
This PR adds a util for detecting if sst is running on a WSL host, and extracting the IP of that host. The WSL IP is used in place of `host.docker.internal` in the `AWS_LAMBDA_RUNTIME_API`

NB: Manually replacing `host.docker.internal` with the IP of my WSL host previously worked flawlessly, but more recently I've had to make changes to the Docker Engine [settings ](https://github.com/docker/for-win/issues/9168#issuecomment-771971994) to get this to work


dylanirion:
left an erroneous comment in there 😬

fwang:
Thanks @dylanirion!

Will be in the next release.

PR #3578, opened 2023-12-12
docs: the route path was to .ts rather than the function name

nhamilton1:
The API route for `/` and `/session` went to `.ts` rather than the name of the function which is `handler`

Also, I am not 100% sure if `"GET /": "packages/functions/src/auth.ts"` in the API Construct are needed for these examples.


jayair:
Oh yeah that's weird. Thanks!

jayair:
Removed the extra route as well.

PR #3579, opened 2023-12-12
Examples: remove Thundra APM as it's out of service

floriangosse:
Reference: https://www.catchpoint.com/press-releases/catchpoint-invests-to-advance-api-cloud-functions-and-microservices-monitoring


jayair:
Thank you!

PR #3584, opened 2023-12-13
fix: properly calculate content-length for cfn custom resource responses

auvred:
Ref: https://github.com/aws/aws-cdk/pull/24501

> Custom Resources need to write their response into a S3 object.
> This is implemented as a PUT request to a pre-signed URL and has to specify the `content-length` of the response object.
> Previously the CustomResource code would use `responseBody.length`.
> However this returns the number of graphemes, not bytes.
> If any utf8 characters with `graphemes != bytes` are part of the response, CloudFormation would fail the deployment with a `Response is not valid JSON` error.

### DIFF
--- a/.changeset/odd-windows-lick.md
+++ b/.changeset/odd-windows-lick.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Properly calculate content-length for cfn custom resource responses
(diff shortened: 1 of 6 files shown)


jayair:
Oh interesting. Let me ask the team to take a look.

PR #3586, opened 2023-12-14
feature: Add cache options for python bundling

BLucky-gh:
Currently the python bundling docker build is completely uncached, wasting time on every build just to download the same dependencies every time, but cdk allows passing caching options to the `fromBuild` function that's used in the bundle function, so I just passed that through to the input. I am not sure how to generate the necessary docs for this, or if that's handled by CI, but I did add a docstring where I felt like it is appropriate

I thought of adding the same to container runtime but seems like that one directly calls `docker run` instead of cdk's `DockerImage.fromBuild`, and just passes build args from `buildArgs`, which lets the user achieve the same result anyways


jayair:
Oh interesting. Taking a look.

BLucky-gh:
Hi, it's been a while. Have you had a chance to take a look? any updates?

jayair:
Oh sorry, we've been away for the holidays. But I'll ping the team again.

BLucky-gh:
By the way, I've tested the patch already, I have patched our own installation of sst in the meantime and it's working

jayair:
Oh good yeah. The team should get to this soon.

fwang:
Thanks @BLucky-gh! I moved the option to `Function.python.dockerBuild` to accomodate more docker build options down the road.

Will be in the next release.

PR #3591, opened 2023-12-18
feat: update render call to support all Astro versions

bayssmekanique:
The `app.render` call deprecates the third argument in favor of an object with the `routeData` as a property.

See https://github.com/withastro/astro/pull/9199 for more information.


bayssmekanique:
The initial `v4` update actually happened as part of #3647, but this PR addresses the scenario where sst is used with older versions of Astro (ex: `v3`) which don't support the change in the `render` method argument.

jayair:
Oh do we need to look into this?

bayssmekanique:
@jayair, yes, @fwang brought up the idea in a thread on the Astro channel. This would preserve functionality for older versions of Astro.

PR #3602, opened 2023-12-31
fix: Astro greedy routes

### LINKED ISSUE #3601: Greedy Astro routes break static documents
This issue was brought up in this [Discord Thread](https://discord.com/channels/983865673656705025/1189921596715053107).

The issue is that some libraries (ie i18n libraries) use slugs for lazy routing of locales, and often base the locale slug element at the root of the site (ex `example.com/fr/example-site`). This results in a route in the `sst.buildMeta.json` which matches against all root elements.

```json
    { "route": "/[locale]/", "type": "redirect", "pattern": "/^\\/([^/]+?)\\/$/", "redirectPath": "/${1}" },
    { "route": "/[locale]", "type": "page", "pattern": "/^\\/([^/]+?)$/", "prerender": true },
```

If this route is SSR, this shouldn't be an issue. When it's prerendered, this results in all root level elements not in the manifest (all those in the `/public` directory and any created at build time) to be treated like locale slugs.

**Example:**

`/fr` => `/fr/index.html`
`/sitemap-index.xml` => `/sitemap-index.xml/index.html`

The simplest solution would be to parse the slug for file extension and ignore index postfix if file extension is detected. This is not ideal, and I would prefer if Astro was able to handle this scenario better, but this issue keeps appearing and this is a solution that addresses the reality of the current implementation.

bayssmekanique:
This PR resolves #3601, preventing Astro route paths with very greedy RegEx values from capturing requests for static files by testing for a file extension in the CF Function and preventing rewrites if an extension is present.

Example routing before implementation of this PR:
`/fr` => `/fr/index.html`
`/sitemap-index.xml` => `/sitemap-index.xml/index.html`

Example routing after implementing this PR:
`/fr` => `/fr/index.html`
`sitema-index.xml` => `/sitemap-index.xml` (*unchanged*)

The only breaking change I can foresee with this change is if someone were to have implemented a false mime-type endpoint using the static rendering function (`/file.png` => `/file.png/index.html`) to serve an HTML page from a URL that appears to point to an asset. I can't see the value in doing this, so I'm assuming it's safe to abandon this functionality to resolve issues with i18n libraries.

### DIFF
--- a/.changeset/honest-vans-itch.md
+++ b/.changeset/honest-vans-itch.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Astro: fix greedy routes bug
--- a/packages/sst/src/constructs/AstroSite.ts
+++ b/packages/sst/src/constructs/AstroSite.ts
@@ -57,7 +57,7 @@ export class AstroSite extends SsrSite {
     
   var matchedRoute = findMatch(request.uri, routeData);
   if (matchedRoute) {
-    if (!matchedRoute[1]) {
+    if (!matchedRoute[1] && !/^.*\\.[^\\/]+$/.test(request.uri)) {
       ${
         pageResolution === "file"
           ? `request.uri = request.uri === "/" ? "/index.html" : request.uri.replace(/\\/?$/, ".html");`


jayair:
Thanks!

PR #3608, opened 2024-01-05
Fix typo in Cognito.about.md

brno32:
when specifying a specific trigger, the method is called `attachPermissionsForTrigger`, not `attachPermissionsForTriggers`


jayair:
Thanks!

Issue #3607, opened 2024-01-03
NextjsSite Construct does not forward VPC configuration to the Image Optimization Function lambda


jplwood:
Lower environments of my application talk to internal-only DNS/resources. For this reason, I have configured my `NextjsSite` SST construct to attach my application to my VPC like so:

```ts
const vpc = Vpc.fromLookup(stack, 'VPC', { vpcId: '<MY ID>' });

const site = new NextjsSite(stack, name, {
  cdk: {
    server: { vpc },
  }, 
  //...
});
```

This works properly for my application code. It is able to talk to services behind private-network DNS. However, my app also loads images from private DNS. As far as I can tell, the application is unable to reach these images. According to the open next documentation, Next.js is routing images loaded via the `Image` component through the [Image Optimization Function](https://open-next.js.org/advanced/architecture#image-optimization-function) (see below)
 
![image](https://github.com/sst/sst/assets/2601724/825336d7-3297-47ab-9916-946c23f43568)

This function can't seem to access images only available via the VPC (requests time out, DNS not resolvable). This implies that the VPC configuration is only given to the Server Function app lambda, and not forwarded to the Image Processing Function lambda.

[AlcaponeYou in discord confirmed this](https://discord.com/channels/983865673656705025/1187522846939959377/1187524780472160366) to be the case, that `vpc` props are not forwarded to that lambda: [/packages/sst/src/constructs/NextjsSite.ts](https://github.com/sst/sst/blob/master/packages/sst/src/constructs/NextjsSite.ts#L313-L329)

---

See original thread in Discord: https://discord.com/channels/983865673656705025/1187522846939959377

jayair:
Thanks! Would appreciate a PR!

kaue-delmonte-acn:
Any solutions?

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #3612, opened 2024-01-08
[SsrSite/NextjsSite]: Add cdk server configuration to ImageOptimizationFunction in SsrSite base construct

jplwood:
`enhancement` (?)

See https://github.com/sst/sst/issues/3607, this provides a fix for that issue in the NextjsSite construct where VPC configuration is only applied to the origins of `type: 'function'` (see [this](https://github.com/sst/sst/blob/master/packages/sst/src/constructs/SsrSite.ts#L1098C1-L1100) and [this](https://github.com/sst/sst/blob/master/packages/sst/src/constructs/SsrSite.ts#L1010)), but not the image optimization function (see [this](https://github.com/sst/sst/blob/master/packages/sst/src/constructs/SsrSite.ts#L1101-L1102) and [this](https://github.com/sst/sst/blob/master/packages/sst/src/constructs/SsrSite.ts#L1057-L1070)).

All constructs inheriting from the `SsrSite` construct will inherit this change. Hoping for confirmation from a reviewer that this is desirable.

### DIFF
--- a/packages/sst/src/constructs/SsrSite.ts
+++ b/packages/sst/src/constructs/SsrSite.ts
@@ -1067,6 +1067,7 @@ function handler(event) {
           }),
         ],
         ...props.function,
+        ...cdk?.server,
       });
 
       const fnUrl = fn.addFunctionUrl({


jayair:
Ah cool. Thank you!

thdxr:

PR #3628, opened 2024-01-19
[SsrSite/NextjsSite]: Add optional disableCloudFrontFunctions to SSRSite

alexabidri:
This PR aims to add an optional parameter `disableCloudFrontFunctions` to avoid creating cloudfront functions when it is not needed. One of the use-case is that we may use SSR deployment with another CDN than Cloudfront (Cloudflare). In this case, host header is not needed in Cloudfront

### DIFF
--- a/packages/sst/src/constructs/NextjsSite.ts
+++ b/packages/sst/src/constructs/NextjsSite.ts
@@ -278,7 +278,7 @@ export class NextjsSite extends SsrSite {
     this.removeSourcemaps();
     return this.validatePlan({
       edge: edge ?? false,
-      cloudFrontFunctions: {
+      cloudFrontFunctions: cdk.disableCloudFrontFunctions ? undefined : {
         serverCfFunction: {
           constructId: "CloudFrontFunction",
           injections: [this.useCloudFrontFunctionHostHeaderInjection()],
--- a/packages/sst/src/constructs/SsrSite.ts
+++ b/packages/sst/src/constructs/SsrSite.ts
@@ -464,6 +464,17 @@ export interface SsrSiteProps {
      * ```
      */
     viewerProtocolPolicy?: ViewerProtocolPolicy;
+    /**
+     * Disable the CloudFront functions that are used to implement SSR.
+     * @default
+     * By default, the CloudFront functions are enabled for host header injection.
+     * @example
+     * ```js
+     * cdk: {
+     *  disableCloudFrontFunctions: true,
+     * }
+     */
+    disableCloudFrontFunctions?: boolean;
     server?: Pick<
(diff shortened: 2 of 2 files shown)


jayair:
Hmm this is a bit tricky. Cos disabling it removes some of the routing we do. Let me check on this.

alexabidri:
@jayair I understand your potential concern. For your information, I have already a fork of that improvement on another repo and everything works properly.

fwang:
hey @alexabidri, wouldn't it be more ideal if `NextjsSite` didn't create the CloudFront Distribution at all? And you can wire up Cloudflare to the Lambda function URLs and the S3 bucket.

alexabidri:
@fwang I understand your point, but in my cases, i still want to have some benefits of the cloudfront distribution, even tough it doesn't seems very standard to have double CDN at first. It involves more complex considerations that I will avoid to expose in this discussion hehe 😃 

The SST logic provides robust and reliable functionality with Cloudfront. For now I prefer to be defensive and still relies on it. 

So if i can juste disable the CF functions that will be perfect (I am already doing it on a fork version but I would prefer to use the official and bring this little improvement)

fwang:
Thanks for the details @alexabidri. This is a pretty unique use case. Introducing a top level prop `disableCloudFrontFunctions` is probably not ideal.

Instead, I [added a `cdk.transform` prop](https://github.com/sst/sst/commit/b29ce15fa94305cc50e4a40b95476cf8bd316bca), which takes a callback that lets your customize the site.

To add a bit of context, each site has this concept of a `plan` that defines the origins, edge function, CF functions, etc that are going to be created. [See `NextjsSite`'s plan here](https://github.com/sst/sst/blob/b29ce15fa94305cc50e4a40b95476cf8bd316bca/packages/sst/src/constructs/NextjsSite.ts#L248). With this callback, you can do something like this:
```ts
await createSite({
  cdk: {
    transform: (plan) => {
      // remove CF functions from the plan
      delete plan.cloudFrontFunctions;

      // add Edge functions to `plan.edgeFunctions`
      ...

      // loop through each `plan.behaviors`, set `edgeFunction` and remove `cfFunction`
      ...
    },
  },
});
```

This will be in the next release. Let me know if it makes sense.

alexabidri:
@fwang Thanks for the suggestion, it is good to me! We can close this PR in favor of https://github.com/sst/sst/commit/b29ce15fa94305cc50e4a40b95476cf8bd316bca

fwang:
@alexabidri fwiw here's an example of using `transform` to implement Basic Auth - https://docs.sst.dev/constructs/NextjsSite#configuring-basic-auth

PR #3630, opened 2024-01-20
Allow setting nodejs version of container image in jobs construct

mathisobadia:
This PR updates the docker images used for jobs, I searched the latest available ARM image with nodejs 18 and 20. 
this PR sets the new default as nodejs18 but I could revert it back to nodejs 16 if necessary so as not to create a potentially breaking change

### DIFF
--- a/.changeset/lazy-tools-float.md
+++ b/.changeset/lazy-tools-float.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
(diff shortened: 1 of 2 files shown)


jayair:
Oh interesting. Let me ask the team to take a look.

fwang:
Thanks @mathisobadia!

^ I added the "nodejs" option back to avoid a breaking change.

Will be in the next release.

PR #3633, opened 2024-01-22
docs: Update to the new way of importing an existing SNS Topic

neo:
according to https://docs.sst.dev/constructs/Topic#importing-an-existing-topic


jayair:
Thanks!

PR #3637, opened 2024-01-25
fix: Converting layer example from nodejs.install to nodejs.esbuild.external

kieronjmckenna:
The current example for using sharp and chrome layers uses the install option of the Function construct

This causes the function to install the package, which is not correct as we're looking to mark the function as external and use the code from the layers.

This PR converts from using nodejs.install in the function config to nodejs.esbuild.external


jayair:
Thanks!

PR #3650, opened 2024-01-31
Add support for RemixSite with vite

SpencerDuball:
# This PR adds support for using Vite with the RemixSite construct.

**Description of changes:**

- First we add logic to detect the `vite.config.js` or `vite.config.ts` file. If this is detected we will handle a few items differently.
- A new `edge-server-vite.js` and `regional-server-vite.js` need to be added to the `support/` files for Remix because the import for the server bundle is located in a new location now.
- The public assets have been changed and should be copied from a new location. Previously the `build/` was added to the actual `public/` folder of the source code, and then this whole folder was copied to S3. Now Vite pulls all static files and puts them in the `build/client` folder and we just pull from there instead.
- Finally we no longer need the polyfill as Vite takes care of this for us as well.

---

**Feedback Areas / Points to Consider**

- In the `getServerModuleFormat()` we are reading the config file with `require(configPath)` when using the `remix.config.js`, it's not possible to read the `vite.config.ts` file as we can't use a dynamic import and this file probably won't be written in CJS format (or rather it would significantly complicate things if we tried). I think this should be fine however, but is a point to note. We already don't check and just assume ESM if we have issues reading the config file.
```js
let userConfig;
try {
   userConfig = require(configPath);
 }
 catch (e) {
    return "esm";
 }
```
- I added logic to look for both the `vite.config.js` and `vite.config.ts` however Remix's docs only contain reference to the `vite.config.ts` - should we bother looking for the JS file?
- I have omitted the logic to use the `versionedSubdir` in the s3 config. This is because all build files are not located neatly in 1 folder anymore, plenty of build files live in the root of the folder. Is this needed?
- Is there any way around two new support files? There is no issue with this, but only changed one line in each file and figured this might be the cleanest approach. Open to ideas here. I had to add the:
    - `edge-server-vite.js` (along side the `edge-server.js`)
    - `regional-server-vite.js` (along side the `regional-server.js`)


SpencerDuball:
closes https://github.com/sst/sst/pull/3650

jayair:
Oh cool!

Asking the team to take a look.

packages/sst/src/constructs/RemixSite.ts
```diff
@@ -66,6 +71,13 @@ export class RemixSite extends SsrSite {
       },
     };
 
+    // The path for all files that need to be in the "/" directory (static assets)
+    // is different when using Vite. These will be located in the "build/client"
+    // path of the output. It will be the "public" folder when using remix config.
+    const publicAssetsPath = this.usingVite
+      ? path.join("build", "client")
```
dkAndFed, line 78: Hardcoding these values makes it extremely difficult to adopt to new defaults that Remix might choose to introduce in the future.

Perhaps introducing some configurables on the RemixSite construct might be a better way. That way we don't have to care whether it's Vite or not as all SST needs to know is where to read from.

As for the `*-server.js` support files, it might be better to treat them as templates as they are not executed until they are in the lambda. That way in combination with aforementioned configurables it can be easier to control where the import (`import * as remixServerBuild from "./server/index.js";`) happens from.

fwang: hmm.. in Astro, there's this adapter concept. And the `astro-sst` adapter writes all the build config (ie. assets build, server build) to a manifest file, and the `AstroSite` construct then read from the manifest file.

Next.js also generates these manifest files.

Is there something similar for Remix?

SpencerDuball: For the templates we can probably just use the `esbuild: inject` where I had removed the polyfills. From there we can just inject the single line of where to import the server build from. This will allow us to have a single template and dynamically specify where the actually server entry is located.

Also hardcoding is what we do currently, if you check the code when we examine for the `remix.config.js` file we ensure that these values are either implicit or hardcoded exactly. I do agree giving more of the options would be good, but thinking maybe that should come in a separate PR. At least we could add Vite support for time being. I actually believe we will need a little bit of a rewrite as Remix offers the `serverBranchBundles` concept now - this allow splitting the app into chunks for multiple lambdas. I don’t believe the `SsrSite` itself allows this, IMO we should consider expanding the config and including this detail at the same time.

SpencerDuball: So to clarify on the inject:
1. We can remove the import from the template.
2. We can check is using Vite, and then inject a line of code that does the `import * as remixServerBuild` in the `createServerLambdaBundle` function.

SpencerDuball: > hmm.. in Astro, there's this adapter concept. And the `astro-sst` adapter writes all the build config (ie. assets build, server build) to a manifest file, and the `AstroSite` construct then read from the manifest file.
> 
> Next.js also generates these manifest files.
> 
> Is there something similar for Remix?

Will check for manifest files tonight after getting back from office and see if this is possible. This would solve the biggest hurdle of not being able to dynamically import the `vite.config.ts` file, we also can’t use require because it’s not guaranteed to be written in CJS format.

fwang: > splitting the app into chunks for multiple lambdas
In `plan()`, you can define multiple server functions and define which behaviors use which function. None of the SSR site constructs is doing it.. but should be supported. But agreed that should be a separate PR.

SpencerDuball: > > splitting the app into chunks for multiple lambdas
> 
> In `plan()`, you can define multiple server functions and define which behaviors use which function. None of the SSR site constructs is doing it.. but should be supported. But agreed that should be a separate PR.

Oh terrific, did not know this! Will have to investigate for a PR on this after this one.

woble: Regarding the manifest file, starting with [Remix 2.6.0](https://github.com/remix-run/remix/blob/main/CHANGELOG.md#v260) it respects the explicit setting in `vite.config.ts` for generating the manifest.

SpencerDuball: Ok have gotten deploys working and removed the two added `support/` files. Only had to change the `createServerLambdaBundle` but am now doing the injection in the procedure itself. Probably fine, but would like an opinion if you guys think this looks good for this PR?

(Section that changed is the `buildInjector` section.)

```ts
  private createServerLambdaBundle(wrapperFile: string) {
    // Create a Lambda@Edge handler for the Remix server bundle.
    //
    // Note: Remix does perform their own internal ESBuild process, but it
    // doesn't bundle 3rd party dependencies by default. In the interest of
    // keeping deployments seamless for users we will create a server bundle
    // with all dependencies included. We will still need to consider how to
    // address any need for external dependencies, although I think we should
    // possibly consider this at a later date.

    // In this path we are assuming that the Remix build only outputs the
    // "core server build". We can safely assume this as we have guarded the
    // remix.config.js to ensure it matches our expectations for the build
    // configuration.
    // We need to ensure that the "core server build" is wrapped with an
    // appropriate Lambda@Edge handler. We will utilise an internal asset
    // template to create this wrapper within the "core server build" output
    // directory.

    // Ensure build directory exists
    const buildPath = path.join(this.props.path, "build");
    fs.mkdirSync(buildPath, { recursive: true });

    // Copy the server lambda handler and pre-append the build injection based
    // on the config file used.
    //
    // Note: When using Vite config, the output build will be "server/index.js"
    // and when using Remix config it will be `server.js`.
    const buildInjector = [
      `// We have to ensure that our polyfills are imported prior to any other modules`,
      `// which may depend on them`,
      `import { installGlobals } from "@remix-run/node";`,
      `installGlobals();`,
      ``,
      `// Import the server build that was produced by 'remix build'`,
      `import * as remixServerBuild from "${
        this.usingVite ? "./server/index.js" : "./server.js"
      }";`,
      ``,
    ].join("\n");
    fs.writeFileSync(path.resolve(buildPath, "server.js"), buildInjector);
    const wrapperFileContents = fs.readFileSync(
      path.resolve(__dirname, `../support/remix-site-function/${wrapperFile}`),
      { encoding: "utf8" }
    );
    fs.appendFileSync(
      path.resolve(buildPath, "server.js"),
      wrapperFileContents
    );

    // Copy the Remix polyfil to the server build directory
    //
    // Note: We need to ensure that the polyfills are injected above other code that
    // will depend on them when not using Vite. Importing them within the top of the
    // lambda code doesn't appear to guarantee this, we therefore leverage ESBUild's
    // `inject` option to ensure that the polyfills are injected at the top of
    // the bundle.
    const polyfillDest = path.join(buildPath, "polyfill.js");
    fs.copyFileSync(
      path.resolve(__dirname, "../support/remix-site-function/polyfill.js"),
      polyfillDest
    );

    return {
      handler: path.join(buildPath, "server.handler"),
      inject: this.usingVite ? [] : [polyfillDest],
    };
  }
```

SpencerDuball:
Have cleaned this up per the suggestions, @woble and @dkAndFed! Have removed the support files and now programmatically adding the import based upon if this is for Vite or Remix config. Doing a diff from before these commits and to the latest should show that things have been simplified a bunch.

@fwang If any other suggested improvements please let me know - thanks!

dkAndFed (approved):
Hopefully someone from the team can have a look and merge/release it.

jayair:
We'll take a look early next week. If we don't get to it, feel free to mention me again.

dkAndFed:
@jayair Did you guys have time to review this yet?

Heher:
@jayair Just a new ping to see if there's any updates with this

fwang:
Thanks @SpencerDuball! Will be in the next release.

PR #3651, opened 2024-02-01
Function/Job/Service: Add buildSsh, cacheFrom, cacheTo to container props

adrianisk:
# Summary

Exposes the `buildSsh`, `cacheFrom` and `cacheTo` properties from the CDK [AssetImageCodeProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.AssetImageCodeProps.html#interface-assetimagecodeprops) interface when building a docker image for a Function, Job, or Service

* `buildSsh`: SSH mounts are [recommended by Docker](https://docs.docker.com/build/building/secrets/#ssh-mounts) for passing secrets like Github deploy keys to the build process
* `cacheFrom`/`cacheTo`: Allows users to speed up their CI/CD build times by using the (experimental) [Github actions cache](https://docs.docker.com/build/cache/backends/gha/) or ECR who [recently announced support](https://aws.amazon.com/blogs/containers/announcing-remote-cache-support-in-amazon-ecr-for-buildkit-clients/)

I added a test using all of the new container props that only use inline/local caches, so I'm assuming it should work but I haven't gotten tests running locally so not positive


jayair:
Thanks! Taking a look.

adrianisk:
@jayair taking another look at this, the test I added failed due to a bug in the Docker flag generating code I added. I'll push a commit shortly to fix

adrianisk:
Fixed the tests, though I had to remove the `buildSsh: 'default',` container prop because the Docker build was failing with

```bash
Error: Failed to build function "test/constructs/container-function"
Error: Command failed: docker build -t sst-build:c8ed9dbe9a800c218512b749c670e088ecf3a53842 --ssh default --cache-from=type=inline --cache-from=type=local,src=.,mode=max --cache-to=type=inline --platform linux/amd64 .
ERROR: invalid duplicate ID default
```

which doesn't make sense to me because the generated build command looks fine. I'm not sure where the `invalid duplicate ID default` comes from, I couldn't find anything online. Specifying the `--ssh default` should be valid though, it's how they do it [in their docs](https://docs.docker.com/build/building/secrets/#ssh-mounts)

mjperrone:
@jayair can you take another look at this? I would love to have this PR merged

jayair:
Oh let me check on this again.

PR #3654, opened 2024-02-02
[future/auth] use query params on success too

benjamine:
bringing back this behavior from v2.24.22
https://github.com/sst/sst/blob/766436883ee7b7b23637cf9d3effd08a07b861fe/packages/sst/src/node/future/auth/handler.ts#L300 so `state` query param is also copied from query params on success step too.

the regression observed is that `?state=` is empty when doing the final redirect after success (at least tested with google adapter) as `state` arrives as a query param to the auth handler.

link to discord with more details: https://discord.com/channels/983865673656705025/1082462837710012508/1203065839591489566

### DIFF
--- a/packages/sst/src/node/future/auth/handler.ts
+++ b/packages/sst/src/node/future/auth/handler.ts
@@ -485,7 +485,10 @@ export function AuthHandler<
               }
             );
 
-          const { client_id, state } = useCookies();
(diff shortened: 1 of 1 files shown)


jayair:
Taking a look.

thdxr:

Issue #3658, opened 2024-02-06
Rust lambda compilation fails on copying files.


ZakkProjects:
When compiling rust lambda functions on Windows platform, there is bug when sst runtime handler for rust is trying to copy files into the artifacts folder.

Running `cargo build` on windows creates files with .exe extension.
![image](https://github.com/sst/sst/assets/26041819/0ad1a84f-7aa1-4617-a34a-9b7c3abd4e44)

Copying files into artifacts folder then failes because it cant find "main" file. 

In deploy mode, there is no need to add ".exe" extension to "bootstrap" file since `cargo lambda build` is set up correctly to build with linux environment.

---

[If you've already posted your issue on Discord, make sure to leave a link to it here.](https://discord.com/channels/983865673656705025/1202981470160035850)

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #3659, opened 2024-02-06
Runtime: Fix rust local-lambda build step on windows #

ZakkProjects:
When compiling rust lambda functions on Windows platform, there is bug when sst runtime handler for rust is trying to copy files into the artifacts folder.

Running cargo build on windows creates files with .exe extension.
image

Copying files into artifacts folder then failes because it cant find "main" file.

In deploy mode, there is no need to add ".exe" extension to "bootstrap" file since cargo lambda build is set up correctly to build with linux environment.

[Discord thread](https://discord.com/channels/983865673656705025/1202981470160035850)

### DIFF
--- a/.changeset/fuzzy-apes-fetch.md
+++ b/.changeset/fuzzy-apes-fetch.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+Runtime: Fix rust local-lambda build step on windows #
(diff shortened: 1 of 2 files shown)


ZakkProjects:
closes #3658

jayair:
Taking a look!

PR #3668, opened 2024-02-12
Docs: Fix the cron job rate to be once a day

deriegle:
The remix starter example includes a cron job to delete uploaded files once per day, but the schedule was set to once per minute. This commits fixes the cron rate to be once per day to match the rest of the documentation.


jayair:
Ah thanks!

PR #3670, opened 2024-02-13
[DNM] feat: add llrt runtime

### LINKED ISSUE #3669: Allow use of LLRT runtime
AWS released an _experimental_ runtime called **LLRT** (Low Latency Runtime) which promises to be faster at cold boots than even some of the compiled runtimes containers like Go. The full details are here: https://github.com/awslabs/llrt

As an experiment, sst should support using the LLRT runtimes. The best place to start would likely be with the `Function` construct, and then if it's stable and workable apply the changes to `SsrFunction` construct as well.

The runtime should be an opt-in only runtime for the foreseeable future, and I expect there will be annoying bugs and limited support until the runtime hits `v1`, so this should be considered an extremely dumb thing to do in a production environment.

bayssmekanique:
This PR implements the [llrm](https://github.com/awslabs/llrt) runtime as an experiment.

Current implementation adds support to `Function`, `SsrFunction`, `SsrSite`, and `AstroSite` constructs.

Resolves #3669


bayssmekanique:
The runtime currently doesn't support top-level awaits, which is a blocker for the `AstroSite` support, but there is a possibility of this being resolved soon.

Ref: https://github.com/awslabs/llrt/issues/124

bayssmekanique:
There is also an issue with the `static` property on `Class` definitions that's been reported.

Ref: https://github.com/awslabs/llrt/issues/148

bayssmekanique:
The top level await issue can be resolved now with a plugin to abstract away the await. (Ex: [`vite-plugin-top-level-await`](https://www.npmjs.com/package/vite-plugin-top-level-await))

The next issue is the number of APIs that aren't natively supported by the runtime and require polyfills. They are not joking when they say it is not a drop-in replacement for node as there is a fairly large number of common utilities which need to be polyfilled.

rfoel (approved):

PR #3671, opened 2024-02-14
feat(link-adapter): allow configuration of magic link expiry

omikader:
This PR makes it so that users can configure how long they want the magic link to be valid for, in minutes. Right now the default value is 10, which is preserved if a value is not explicitly provided going forward.

Resolves this [issue](https://discord.com/channels/983865673656705025/1204465728623743066) raised in the Discord


jayair:
Thanks!

jayair:
Taking a look.

omikader:
@jayair thanks!

One thing I thought about doing was changing it to accept an SST duration, e.g. `"10 minutes"` or `"2 days"` but I didn't want to bring in CDK dependencies into the node modules, even though they're technically available.

Lemme know what you or Dax prefer.

PR #3672, opened 2024-02-14
Docs: add missing required iam roles

benry1:
These permissions are required, but not found in the docs:

`iam:PassRole` is required for deploying any stack.
`s3:GetBucketLocation` is required to use an existing CDK bootstrap bucket


jayair:
Thank you!

PR #3674, opened 2024-02-17
Function: Don't overwrite Lambda description for Live Lambda development, just suffix label

mikeyaa:
[This commit](https://github.com/sst/sst/commit/36bf1b86d285fe742579a7c9bad3179e5b5df1e1#r138742925) overwrites any custom description provided on a Function with "SST Live Lambda handler" during development. In the AWS console, when looking at many functions who previously had a description, this isn't as helpful. It would be nice to maintain the user provided description but also let them know that it is being run via Live Lambda by suffixing their description instead of overwriting it completely. I chose to suffix the label but could alternatively do a prefix.


jayair:
Thanks!

fwang:
Thanks @mikeyaa! Will be in the next release.

PR #3676, opened 2024-02-17
[sst/cli]: Fix hang on empty secrets list

### LINKED ISSUE #3675: `sst secrets list` hangs if there are no secrets set in a stage
To reproduce:

```
sst secrets list --stage not-a-real-stage
```

The output is "No secrets set", and the process never terminates. This is a problem when using the command in a script.

tobz1000:
Fixes #3675.


jayair:
Oh taking a look.

PR #3683, opened 2024-02-27
Update Service construct to be able to create an http -> https listener on the load balancer itself.

BradyRyun:
Internally, my team needs to be able to create a load balancer with HTTP redirecting traffic to HTTPS. This is an initial start. There needs to be additional configuration available here, but this unblocks myself and should allow others to be able to do the same with a basic setup.

I am open to feedback on how to pass the variables to the construct. I put them at the top-level, but I considered adding them to the applicationLoadBalancer object in the cdk options.

### DIFF
--- a/packages/sst/src/constructs/Service.ts
+++ b/packages/sst/src/constructs/Service.ts
@@ -561,11 +561,11 @@ type HttpsListenerProps = {
    * @example
    * ```js
    * {
-   *   https: true,
+   *   httpsRedirect: true,
    * }
    * ```
    */
-  https?: boolean;
+  httpsRedirect?: boolean;
   /**
    * This option attaches these certificate to the https load balancer. This must be specified
    * @default undefined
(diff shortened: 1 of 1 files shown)


jayair:
Thanks! Yeah maybe a cdk prop would make more sense. Let me check with the team.

BradyRyun:
@fwang Updated now. JetBrains user so it didn't find the VSCode formatting options preconfigured. Looks good now. :)

BradyRyun:
@fwang @jayair bump. :)

jayair:
I'll ping Frank

fwang:
@BradyRyun sorry about the delay.

I'm hesitate on adding the http to https redirect as a built-in feature. Let's wait and see if more people need it, and we can support it as a top level feature. ie.

That said, I just published `2.41.2` to support customizing the listener. This should make adding http to https redirect easily. Here's an example — https://docs.sst.dev/constructs/Service#configuring-application-load-balancer-http-to-https-redirect

Let me know if that works for you.

BradyRyun:
@fwang This looks like it will work for us. I will let you know once I get a chance to get back to this.

PR #3686, opened 2024-02-28
Fix non functional template for Prisma

eguenou:
Fixed:
- update Prisma files location (remove the obsolete one - .prisma)
- rmSync and cpSync instead of removeSync and copySync
- add 'recursive' option while copying prisma files to avoid the error 'Path is a directory: cp returned EISDIR'
- remove the local only deployment to make it work with 'sst dev' mode
- load database URL from .env instead of hard coded one
- added the "linux-arm64" architecture if you are using AWS Graviton 2 (source: https://www.prisma.io/docs/orm/prisma-client/deployment/serverless/deploy-to-aws-lambda)
Some of the updates above are based on this repo: https://github.com/geauser/sst-prisma-example
Tested OK with Node 20 LTS.


jayair:
Appreciate it!

PR #3701, opened 2024-03-10
modified render options to add cookies

lvindotexe:
cookies that were set with the  `Astro.cookie.set()` method were not passed on to the CloudflareResult. and some cookies were also duplicated before they were converted into specific AWS Responses


jayair:
@bayssmekanique does this make sense to you? Or is this more for Frank to look at?

bayssmekanique:
@lvindotexe, could you provide some context to the original issue this PR addresses. I see the code and I think it's harmless enough, but I'm not sure I understand the original issue. Is it possible to get a reproduction step or example of some sort?

thdxr:

PR #3713, opened 2024-03-15
Fix: add x-prerender-revalidate to allowed headers in cache policy

sommeeeer:
Adding `x-prerender-revalidate` header as one of the default allowed headers in cache policy will solve some inconsistencies with executing `res.validate()`

This is the **HEAD** request issued by open-next : [revalidate.ts L33-L51](https://github.com/sst/open-next/blob/a475e9dbc48b5f83f8735599baea4206e1738b4c/packages/open-next/src/adapters/revalidate.ts#L33-L51)


jayair:
I'll ask the team to take a look! Thanks!

PR #3726, opened 2024-03-24
Update WebSocketApi.about.md

sashalikesplanes:
I tried to use the SST Auth `useSession` with the websocket api and could not get it to work. Had to read the source code. This is an update to the docs showing the example which worked for me.

See line 33-38 of `st/src/node/auth/session.ts`


jayair:
Thanks!

PR #3727, opened 2024-03-25
- Update Open Next Version for NextJSSite construct

neilsaccardo:
- Update Open Next Version for NextJSSite construct to 2.3.8: This is to include the fix included in https://github.com/sst/open-next/pull/387


jayair:
Yup taking a look

jayair:
Oh we might've missed this

PR #3729, opened 2024-03-26
[Function]: Update go runtime

jamesgibbons92:


jamesgibbons92:
Dunno if rust should be updated too 🤷‍♂️

jayair:
Let me see.

henrycrib:
Rust should be updated too

PR #3755, opened 2024-05-01
fix: Update Remix.run -> remix.config.js

bartoszgolebiowski:
esm does not work out of the box, so I just changed it to cjs


lukasjsk:
"esm" module type seems to be required: https://docs.sst.dev/constructs/RemixSite#using-aws-services 
Here you can find fix to be able to use "esm" format: https://github.com/sst/sst/pull/3804

thdxr:

PR #3759, opened 2024-05-08
add postgresq13.12 as an option - RDS.ts

adminCitify:
There are cases of people that have deployed postgresql11.13, which now appears as deprecated, and can't be deployed anymore. For those cases, they can modify by rds console to version 13.12 and the deploy postgresql13.12. To do that we need to add this option.


jayair:
Taking a look soon

fwang:
Thanks @adminCitify! Will be in the next release.

PR #3764, opened 2024-05-14
chore: update aws-cdk-lib to 2.141.0 from 2.132.1

berenddeboer:
Bump aws-cdk-lib version.


krmorse:
+1 thank you!  looking forward to this getting released, as we're stuck on a pretty old version of `sst` due to the `cdk diff` bug in `2.132.1`

berenddeboer:
@krmorse we need this because it makes it easier to set resolver and depth limits on the graphql api instead of resorting to the escape hatch.

berenddeboer:
And fixed the commit, as I had switched to lockfile version 9 here, kept it at 6 now.

fwang:
Thanks @berenddeboer!

PR #3780, opened 2024-06-02
fix: google and apple adapter

vidalmatheus:
I keep getting `config.issuer.Client is not a constructor` (for example GoogleAdapter) in SST v^2.42.0.
In the previous version it does work SST v2.41.5.
I looked into the implementation and I think this `Proxy` for resolving issuers might not be working as expected.

[Discord](https://discord.com/channels/983865673656705025/1082462837710012508/1246779198132060192)

### DIFF
--- a/.changeset/great-moles-lie.md
+++ b/.changeset/great-moles-lie.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
+fix: google and apple adapter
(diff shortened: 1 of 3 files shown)


jayair:
Thanks

vidalmatheus:
Hey @jayair! I just added an `AppleConfig` type to make the `issuer` field optional in code completion.
https://github.com/sst/sst/pull/3780/commits/16a93aed810dbbc71c3978214744525eeb33e249

oyed:
Would be great to get this fix merged! Currently really bugging me :)

PR #3785, opened 2024-06-07
fix: obey runtime and memorySize configs in `NextjsSite`

relsunkaev:
Introduced in https://github.com/sst/sst/commit/96b99e08f3cfbbbdf9f25f920be73e368b0599d0


fwang:
Thanks @relsunkaev! Will be in the next release.

Issue #3789, opened 2024-06-11
`secrets list env` breaks with multi-line secrets


dan-turner:
we've been sst secrets list env to clone secrets across stages.  Had worked well for us until we put a cert chain in a secret.

The fix was easy, we simply patch sst to wrap the value to console output in single quotes.  Will submit a PR now.

---

Discord thread:
https://discord.com/channels/983865673656705025/1083422760111439973/threads/1250022517578403914

dan-turner:
See this SO answer for more detail on why this should be done:

https://stackoverflow.com/a/27701642

PR #3790, opened 2024-06-11
fix: wrap output in single quotes to allow for multi-line secrets

dan-turner:
Closes: https://github.com/sst/sst/issues/3789


jayair:
Thanks!

PR #3794, opened 2024-06-13
fix(node): support npm versions 10 and above

gronxb:
backward compatibility: 
https://github.com/sst/ion/pull/549

## Environment
M1 Mac
npm version 21

## Problem
Deployments succeed, but the application is not accessible. This issue occurs because the node_modules directory uploaded to AWS Lambda is built for Darwin architecture. Starting from npm version 9, the arch and platform options have been changed to cpu and os.

https://docs.npmjs.com/cli/v9/commands/npm-install#cpu


jayair:
Thanks for submitting it here too!

fwang:
Thanks @gronxb!

PR #3799, opened 2024-06-26
enhancement: make magic link expiry configurable

blair55:
by using an optional parameter on `LinkConfig` interface


blair55:
👋 @jayair

a simple change but anticipating different opinions on the implications. 

if this isn't inline with your thinking then just close 🙏

PR #3802, opened 2024-07-03
[Service] Allow reuse of cache policies for Service constructs

cgcompassion:
AWS allows a maximum of 20 Cache Policies per account. This limit is hit quickly if every service generates its own cache policy. We need a way to reuse cache policies across multiple services.

Something similar was done for Nextjs sites here: https://github.com/sst/sst/issues/899


fwang:
Thanks @cgcompassion!

Made `cdk.cachePolicy` take an ICachePolicy object instead of the ID.

PR #3804, opened 2024-07-04
Remix template update

lukasjsk:
Creating new project with Remix template results in code that is not working. This PR fixes those issues.

```
npx create-sst@latest --template=examples/quickstart-remix
``` 

Fixes:
* switched to Vite (following Remix' migration guide)
* `isbot` import was wrong
* `Response` no longer exists in `@remix-run/node` - instead standard `Response` is used, but that requires to wrap `body` into `createReadableStreamFromReadable`


jayair:
Oh thanks for this. These files are actually generated from this template: https://github.com/sst/sst/tree/master/packages/create-sst/bin/presets/examples/quickstart-remix/templates

Would appreciate a PR to these files instead.

lukasjsk:
Thanks for pointing to correct place. I will fix it. Would it be OK to switch the template to Vite as part of the same PR?

jayair:
Yup thanks!

Issue #3800, opened 2024-06-27
Dependabot reports high security vulnerability due to ws package


richardcho-sr:
Advisory: https://github.com/advisories/GHSA-3h5v-q93c-6h6q

Latest version of sst still has this vulnerable package.

Currently unable to resolve the vulnerability due to transitive dependencies relying on a vulnearbale version of the `ws` package:

```bash
yarn why v1.22.22
[1/4] Why do we have the module "ws"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "ws@8.14.2"
info Has been hoisted to "ws"
info Reasons this module exists
   - Hoisted from "sst#ws"
   - Hoisted from "jsdom#ws"
   - Hoisted from "sst#ink#ws"
   - Hoisted from "sst#@aws-sdk#signature-v4-crt#aws-crt#@httptoolkit#websocket-stream#ws"
info Disk size without dependencies: "180KB"
info Disk size with unique dependencies: "180KB"
info Disk size with transitive dependencies: "180KB"
info Number of shared dependencies: 0
=> Found "websocket-stream#ws@3.3.3"
info This module exists because "sst#aws-iot-device-sdk#websocket-stream" depends on it.
info Disk size without dependencies: "196KB"
info Disk size with unique dependencies: "284KB"
info Disk size with transitive dependencies: "284KB"
info Number of shared dependencies: 3
=> Found "mqtt#ws@7.5.9"
info This module exists because "sst#aws-iot-device-sdk#mqtt" depends on it.
info Disk size without dependencies: "168KB"
info Disk size with unique dependencies: "168KB"
info Disk size with transitive dependencies: "168KB"
info Number of shared dependencies: 0
=> Found "aws-crt#ws@7.5.9"
info Reasons this module exists
   - "sst#@aws-sdk#signature-v4-crt#aws-crt#mqtt" depends on it
   - Hoisted from "sst#@aws-sdk#signature-v4-crt#aws-crt#mqtt#ws"
info Disk size without dependencies: "168KB"
info Disk size with unique dependencies: "168KB"
info Disk size with transitive dependencies: "168KB"
info Number of shared dependencies: 0
```

---

[Link to Discord post](https://discord.com/channels/983865673656705025/1252516046447841380/1252516046447841380)

jgangelBLNG:
[...]

PR #3807, opened 2024-07-09
chore: update ws to 8.18.0 from 8.11.0.

MaTTaDox:
Bump ws because of high security vulnerability see #3800


iloewensen (approved):

MWein:
This may not close the alert. ws 3.3.3 is still in here subordinate to `websocket-stream`

image

MaTTaDox:
@MWein thanks, you are right. If I see it correctly, the solution for this would be to migrate to the `aws-iot-device-sdk-v2` what I think will not be done because it's already part of ion? 

Anyway, is it possible to still merge this PR to resolve 2 out of 3 alerts?

PR #3816, opened 2024-07-17
Fix: cdk bootstrap s3 bucket objects with extra tags - deploy fix

arkadiuszbartnik:
This change fixes the deployment issue when using the resource auto-tagging system on S3. If there are any additional tags on *.map files in the S3 cdk-bootstrap, the deployment will fail without this change.


jayair:
Thanks!

PR #3829, opened 2024-07-30
fix: query params with no values should be strings

scarperharper:
This change fixes the aws v4 signature when query string params with no values are used. This lambda is used when the option `regional.enableServerUrlIamAuth = true` is set in `SsrSite` (in my case `RemixSite`).

For example, posting to Remix index routes hits an endpoint such as:

https://xxxxx.cloudfront.net/?_data=routes%2F_index&index=

This version of the URL works as expected following the change

### DIFF
--- a/.changeset/clean-kiwis-greet.md
+++ b/.changeset/clean-kiwis-greet.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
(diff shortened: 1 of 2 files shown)


jayair:
Thanks

fwang:
Thanks @scarperharper! Will be in the next release.

PR #3832, opened 2024-08-01
docs: update going-to-production.md

pnodet:
The default console Maximum session duration is 1h. Which will cause the ci to fail when copy pasting this code.


jayair:
Thanks!

Issue #3833, opened 2024-08-03
Received response status [FAILED] from custom resource. Message returned: The operation cannot be performed at this time. An update is in progress for resource: arn:aws:lambda:us-east-1:xxxxx


sahilrc:
Running a deployment in CloudFormation. 

Custom::SSTEdgeLambdaVersion xxx   CREATE_FAILED 

Received response status [FAILED] from custom resource. Message returned: The operation cannot be performed at this time. An update is in progress for resource: arn:aws:lambda:us-east-1:xxx:function:xxxLogs: https://sa-east-1.console.aws.amazon.com/cloudwatch/homeevents/ (RequestId: xxx)

The Custom Resource `Custom::SSTEdgeLambdaVersion` is attempting to deploy a layer without waiting for the underlying Lambda to finish deployment.

CloudTrail shows error:

```
    "errorCode": "ResourceConflictException",
    "errorMessage": "The operation cannot be performed at this time. An update is in progress for resource: arn:aws:lambda:us-east-1:xxx:function:xxx",
    "requestParameters": {
        "functionName": "xxxx"
    },
```

While CloudFormation marks the `Custom::SSTEdgeLambda` as UDPATE_COMPLETE, it is has not finished updating and CloudFormation fails and rolls back.

SST needs to check if lambda is truly finished deployment using `$ aws lambda get-function --function-name MY_FUNCTION_NAME --query 'Configuration.[State, LastUpdateStatus]'`

An alternative would be to add a delay before deploying the `Custom::SSTEdgeLambdaVersion` resource.

sahilrc:
I have work around the issue by extracting the CloudFormation template, and adding a dummy resource such as "DynamoDB::GlobalTable" to allow time between the `Custom::SSTEdgeLambda` and `Custom::SSTEdgeLambdaVersion`.

jlarmstrongiv:
Yes, over the past 48 hours, this has broken all of my stacks that deploy edge functions.

@sahilrc could you explain your workaround in more detail?  How do I extract the cloudformation template?

sahilrc:
You can grab the template from CloudFormation console. You will need to be quick - because CloudFormation will roll-back and the template you see will be the old/rolled back template. Alternatively you can ask AWS Support to grab the latest template for you.

[...]

PR #3834, opened 2024-08-04
Fix/hacky sleep for lambda version update: https://github.com/sst/sst/issues/3833

neilsaccardo:
This is just to add a hacky sleep to ensures that the lambda function has finished before it goes to try to create the lambda function version... this is a definite bandaid solution to be able to continue to deploy our apps to envs. adding this as a PR so that others in the same boat have access to a solution if needed
The issue is discussed here:
https://github.com/sst/sst/issues/3833

Issue:
It seems that the creation of the lambda@edge function is not probably awaited/checked before moving onto the next step. it is still in a "pending" state when creating the version, which causes this issue. it seems that at on Thursday, it seems that cloudfront edge lambda functions started taking a few seconds longer to deploy, causing this issue to manifest


jlarmstrongiv:
That’s great!  For a more permanent solution, it looks like we can ping the Lambda API action GetFunction to check a function's state before invocation to verify that it is active, not pending.

It’s also possible to use [`patch-package`](https://www.npmjs.com/package/patch-package) to edit `node_modules/sst/support/edge-function/edge-lambda-version.mjs` to add the timeout to deploy for the time being:

```js
import { setTimeout } from "node:timers/promises";import { createRequire as topLevelCreateRequire } from 'module';const require = topLevelCreateRequire(import.meta.url);
```

```js
;async function sge(t){le("createVersion() called with functionName",t);await setTimeout(30000);let e=await sr.publishVersion({FunctionName:t}).promise();
```

why-el:
I would love to get this fast-merged. I confirmed it worked with a few colleagues. I have not seen it yet in our production CI workflows, possibly because no one has touched these edge functions yet, but once they do we won't be able to deploy. 

Regardless of the origins of this issue (I personally believe a performance regression in lambda@edge provisioning), there is currently no waiting on the lambda versions and this PR fixes that, and we got a few confirmations it works. This will unblock all of us.

jayair:
I don't think we'd be okay with merging this since it's a bit hacky. Though happy to review a proper fix for this.

juliuscarvajal:
Truncated response from AWS tech support team:

---

From the case description I understand that the CloudFormation stack fails to deploy due to a failed response status from the Lambda@Edge function. Please feel free to correct my interpretation of your issue.

This response message seems to be related to the Expansion of AWS Lambda states to all functions [1]. This change impacts workflows that attempt to invoke or modify a function shortly after a create or an update action. To summarize, once you update the Lambda function/code, the Lambda state switches into InProgress update, in which no changes are able to be done until it switches back to LastUpdateStatues: Successful.

As of this change, before making API calls to a function that was recently created or modified, you must confirm it is first in the Active state, and that the LastUpdateStatus is Successful.
If you are updating your functions, a lambda function can only be updated if they are in the ACTIVE or INACTIVE state. Thus, you need to modify your workflow to check the functions state first.
Like so:
- Use the GetFunctionAPI to check that the state is ACTIVE or INACTIVE
- If it is not ACTIVE or INACTIVE, you need to wait some time. 10 seconds, 20 seconds, 1 minute as per your requirements.
- Then try to update your function again

The same goes if you are deploying then invoking
- Deploy
- Use the GetFunction API to check that the state is ACTIVE,
- Else continue to wait and check again after some time; 10 seconds, 20 seconds, 1 minute, the choice is yours.
- Invoke the function

---

Pseudo code:
```
const timeouts = [1000, 2000, 60000];
for (timeout in timeouts) {
  const isActiveAndLastUpdateSuccess = // Get function status
  if (isActiveAndLastUpdateSuccess) {
     break;
  }
  // Wait before next timeout
};

```

med8bra:
Hello, @jayair do you think polling for lambda status would be a fine solution as suggested by @juliuscarvajal ? If so I can try to push an impl or @neilsaccardo you can update this PR

wootencl:
Much appreciated @jlarmstrongiv! Above solution of using a patch package approach (I used `yarn patch`) worked great. Similarly thinking this issue should be prioritized at some level above medium/high. Understand the shift towards ion but this breaking current sst usage seems sufficient enough cause to have the sst team minimally respond to folks that have interest in contributing a solution (many thanks to the people on this thread).

thdxr:

PR #3838, opened 2024-08-16
docs: fix IAM role code example import statements

BrunnerLivio:
Not sure what is the prefered way to import aws-cdk-lib according to SST but I figured it should consistent with how it's described here:
https://docs.sst.dev/constructs/Bucket#importing-an-existing-bucket


jayair:
Thanks!

PR #3840, opened 2024-08-20
ECS (Service): allow disabling default autoscaling

zxan1285:
Hi!
There's a tiny change i need to setup my ECS autoscaling. I need to scale it based on SQS metric, and i also don't want to scale based on default params (cpu, memory).
With this tiny change i can disable default autoscaling by explicitly set scaling to null (`scaling: null`) and then use 
`service.cdk.fargateService.autoScaleTaskCount()` to create any custom scaling scenarios i want. 
Thanks!

### DIFF
--- a/packages/sst/src/constructs/Service.ts
+++ b/packages/sst/src/constructs/Service.ts
@@ -1102,7 +1102,6 @@ export class Service extends Construct implements SSTConstruct {
       minCapacity: minContainers ?? 1,
       maxCapacity: maxContainers ?? 1,
     });
-
     scaling.scaleOnCpuUtilization("CpuScaling", {
       targetUtilizationPercent: cpuUtilization ?? 70,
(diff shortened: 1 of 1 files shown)


jayair:
Can you open an issue for this? I'll assign it to the team and have it looked at.

We might want to decide if we want it to be `null`.

zxan1285:
> Can you open an issue for this? I'll assign it to the team and have it looked at.
> 
> We might want to decide if we want it to be `null`.

Thank you!
#3843 created

I do not insist on `null`, i'm just too lazy to add types for ts to accept anything else :)

zxan1285:
@jayair i've been thinking on alternatives for that `null`. I guess, it could be something like `scaling: { disabled: true }` or `scaling: { enabled: false }`. Though i'm not sure it is better than just `null`.

thdxr:

PR #3859, opened 2024-09-24
Switch npm install command to use `--omit=optional` not `--no-optional`

kopertop:
By default NPM will install optional dependencies, even if you include --no-optional.

```
npm warn config optional Use `--omit=optional` to exclude optional dependencies, or
npm warn config `--include=optional` to include them.
npm warn config
npm warn config       Default value does install optional deps unless otherwise omitted.
```

See https://docs.npmjs.com/cli/v10/commands/npm-install#omit

### DIFF
--- a/.changeset/cuddly-pugs-bow.md
+++ b/.changeset/cuddly-pugs-bow.md
@@ -1,0 +1,5 @@
+---
+"sst": patch
+---
+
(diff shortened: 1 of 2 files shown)


jayair:
This is the `nodejs.install` packages?

fwang:
Thanks @kopertop!

Issue #3858, opened 2024-09-16
SST warmer not compatible with open-next>=3.1.0 with streaming enabled


mateogianolio:
Using open-next >= 3.1.0 with streaming _and_ warmer enabled at the same time causes "Cannot set properties of undefined (setting 'callbackWaitsForEmptyEventLoop')". Disabling the warmer fixes the issue. So something is wrong with the warmer code injection.

---

[open-next discord thread](https://discord.com/channels/983865673656705025/1284191893730623508)

revmischa:
Getting this too

JQuezada0:
I _think_ [this diff fixes it](https://github.com/sst/sst/pull/3864/files#diff-e03d87c9f58a197b51d4687b0190e372f0f1167ef3ad5aeaa75726427d6223e5R427). I ran into this in a different situation (enabling streaming, warming, and using Sentry's `wrapHandler` function for lambda), and the error was gone after installing this patch and deploying. The root of it is that the internal handler function (i.e. the one that the SST wrapper calls) expects the three streaming arguments but only gets the first two.

PR #3864, opened 2024-10-01
Fix streaming + warming causing internal handler to lose access to context object

JQuezada0:
Fixes https://github.com/sst/sst/issues/3858


jayair:
Thanks.

fwang:
Thanks @JQuezada0!

PR #3865, opened 2024-10-02
fix(create-sst): use sst v2 for create-sst

AdditionAddict:
- Install sst@^2 for create-sst presets
 - Update docs create-sst@latest to create-sst@two


jayair:
Thanks!

PR #3613, opened 2024-01-09
[NextjsSite construct] add basePath support back into NextjsSite construct

jplwood:
Partially addresses https://github.com/sst/sst/issues/2964

The [original, now-deprecated version of the `NextjsSite` construct](https://github.com/sst/sst/blob/d6917d8ee581f016072155b1898e7c50ecc141e2/packages/sst/src/constructs/deprecated/NextjsSite.ts) (as far as I can tell, it's deprecated because the non-deprecated construct [uses SsrSite as a base](https://github.com/sst/sst/blob/d6917d8ee581f016072155b1898e7c50ecc141e2/packages/sst/src/constructs/NextjsSite.ts#L174), whereas [the deprecated one doesn't](https://github.com/sst/sst/blob/d6917d8ee581f016072155b1898e7c50ecc141e2/packages/sst/src/constructs/deprecated/NextjsSite.ts#L217)?) supported `basePath`(see [deprecated code](https://github.com/sst/sst/blob/d6917d8ee581f016072155b1898e7c50ecc141e2/packages/sst/src/constructs/deprecated/NextjsSite.ts#L1406-L1411), and [Next.js docs on `basePath`](https://nextjs.org/docs/app/api-reference/next-config-js/basePath)), a crucial feature for hosting apps behind a reverse proxy (and a major convenience when linking around if all of your app's routes are all heavily nested). When the new version of `NextjsSite` construct was released, `basePath` support seemed to have been removed. 

This PR takes that original functionality and adds it back to the new NextjsSite construct. 

Unclear if this is safely additive or if it's a breaking change.

Issue #2964, opened 2023-06-08
SSR Site construct support for base path option


Ankcorn:
All the SSR frameworks support rendering on a base path like 

example.com/docs

* svelte - https://kit.svelte.dev/docs/configuration#paths
* astro - https://docs.astro.build/en/reference/configuration-reference/#base
* next - https://nextjs.org/docs/pages/api-reference/next-config-js/basePath

I can't get this working with sst  :(

jayair:
Hey Thomas, can you post this on #help and we'll take a look at supporting it.

jplwood:
@jayair Any updates on this? Searching discord just turns up workarounds and callouts that `basePath` isn't currently supported. We use `basePath` extensively  with multiple apps behind a reverse proxy, so not having having `basePath` is certainly a hindrance!

jayair:
I think this is supported for Astro and maybe Next.js. Which one are you looking for?

jplwood:
Next.js, yeah, we use it in our app's Next config, but when deployed with SST, that basePath isn't used as a prefix in cloudfront (specifically the app looks for static bundles at `<basePath>/.next/*`, but the cloudfront route has it as `.next/`. That was at least the obvious one, not sure if there are other places it should apply. 

I've got a workaround for now, but certainly would be nice if the next construct used it when defining the CF routing rules!

jayair:
Let me check on the Next.js one with the team.

jplwood:
I derped/lied. My workaround doesn't actually solve the issue. Regarding what I said before: 
> We use basePath extensively with multiple apps behind a reverse proxy

Without base path support, my next app will request all static bundles and such from the root path `/`. When behind a reverse proxy, the routing rules will have no idea that the `/` root requests should be routed to my app without putting all requests behind a `basePath`. Those requests will be sent to the default routing rule.

So this remains a big hindrance.

jplwood:
@jayair since I saw how it was previously done, went ahead and opened a suggested PR ☝️ to address at least the Next.js part of this issue.

eschaefer:
Just wanted to bump this, since using `basePath` seems to result in `403` forbidden errors for static assets for me. Just curious if anyone else using `basePath` experiences this?

thdxr:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #3866, opened 2024-10-03
[SvelteKitSite construct] add basePath support into SvelteKitSite construct

chronotc:
Addresses: https://github.com/sst/sst/issues/2964

Taking inspiration from https://github.com/sst/sst/pull/3613 and updating the behaviours and S3 copy paths.


jayair:
Thanks! We'll take a look.

PR #4941, opened 2024-10-22
docs: s3 lifecycle policy example

agamm:
Added an exmaple to expire s3 bucket files.


jayair:
We need to turn this into an app in the `examples/` dir. Leaving this open for when I get to it.

vimtor:
now this is built-in #6209

PR #4942, opened 2024-10-22
docs: Add node/python lambda examples

agamm:


jayair:
I haven't had a chance to document Python yet. I'll look at this when I get to it.

agamm:
> I haven't had a chance to document Python yet. I'll look at this when I get to it.

A big reason I chose sst was the ability to integrate with AWS using whatever best tech I can use. Having an example is nice, but I'll also look into integrating them more tighly via the SDK, maybe allowing me to call resources like the python lambda from a nodejs app (given there is a VPC/public network between them). This could be a real game changer.

vimtor:
closing in favor of #6061

PR #4943, opened 2024-10-22
docs: Add docs for cluster dev gotcha

agamm:
I didn't see https://github.com/sst/ion/pull/1327 in the docs so adding here again, maybe because the repo changed.


jayair:
Yeah I think I moved this to the right spot in that doc.

PR #4950, opened 2024-10-22
Add support for providers to Postgres.get.

izakfilmalter:
Currently `sst.aws.Postgres.get` will error out / return the host for primary db (if located within the default region).

This pr updates `sst.aws.Postgres.get` to work more like it's pulumi counter part, passing `opts?: ComponentResourceOptions` as the third param.


fwang:
Supported in v3.2.60 - https://github.com/sst/sst/commit/43606b520ce6b5add65a3342cff71af053387d40

PR #4955, opened 2024-10-23
docs: Add tip for WSL login

agamm:
In windows the browser via WSL is clunky, so I figured out a way to open it in the native browser of the host machine.


jayair:
Maybe for these cases it just makes sense to copy-paste and navigate manually to the page in the browser?

agamm:
> Maybe for these cases it just makes sense to copy-paste and navigate manually to the page in the browser?

Such a DX Loss, and a pretty nice fix in my opinion, it just opens a browser automatically - for something that will happen every 12 hours is important.

jayair:
Yeah that's true. I'll take a look.

PR #4962, opened 2024-10-24
Update AWS Python example config

valgaze:
- Update aws-python example's `pyproject.toml` to point to SST's `dev` branch instead of `walln/python-fixes`

CC: @walln


jayair:
Thanks.

PR #4965, opened 2024-10-24
Add new removal policy forbid

danielfyhr:
what do you think of adding a guard rail option to [removal](https://sst.dev/docs/reference/config#removal) ?
{
  removal: input.stage === "production" ? "forbid" : "remove"
}

the idea is to prevent accidental resource removal in production environments. for example, if I mistakenly run a removal command targeting production, i’d prefer the operation to fail and crash, rather than proceed to remove all resources except S3 and DynamoDB. the crash could include a message about updating the removal policy to proceed. i think using retain-all keeps all resources, but i then need to do extra work to sync those resources back into sst. a guard rail could simplify this by preventing the issue in the first place.

### DIFF
--- a/cmd/sst/remove.go
+++ b/cmd/sst/remove.go
@@ -1,8 +1,9 @@
 package main
 
 import (
+	"fmt"
 	"strings"
-
+	"github.com/sst/ion/internal/util"
 	"github.com/sst/ion/cmd/sst/cli"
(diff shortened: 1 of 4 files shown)


jayair:
Yeah let's create an issue for this as well?

fwang:
In v3.3.62, Added `protected: true` flag in sst config.
https://sst.dev/docs/reference/config#protected

PR #4967, opened 2024-10-24
fix: use correct paths for pyproject and related files

michaelgmcd:
Python bundling for sst v3 fails because the target folders for copying files are wrong. They append the absolute path to the target directory. Something like:
/Users/myuser/my/project/path/.sst/artifacts/ChatApiRouteWwasrbHandler-src/Users/myuser/my/project/path/filename

### DIFF
--- a/platform/src/runtime/python.ts
+++ b/platform/src/runtime/python.ts
@@ -51,8 +51,8 @@ export async function buildPythonContainer(
 		await limiter.acquire(name);
 
 		// Find the closest pyproject.toml
-		const pyProjectFile = await findAbove(parsed.dir, "pyproject.toml");
-		if (!pyProjectFile) {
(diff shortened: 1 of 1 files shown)


cosmic-ascendant:
I cannot get the simple examples/aws-python-container to work given this MR. 

Error:
container/.sst/artifacts/MyPythonFunction-src/Dockerfile: no such file or directory

I see in fact that the Dockerfile is at MyPythonFunction-src/src/Dockerfile not MyPythonFunction-src/Dockerfile.

Does work on v3.22.5

pixelcatgg:
This is also causing issues for us, same error as above "Dockerfile: no such file or directory"

jayair:
Open a new issue with your project structure?

Issue #3925, opened 2024-10-20
Allow add custom healthcheck to Service


ubinatus:
#### Description:
When creating a Cluster Fargate service, it would be awesome to allow the user to pass a custom healthcheck block for the target groups associated to the load balancer.

This feature is super helpful if you need to specify a custom endpoint, like when your service depends on other services (e.g., spinning up a database socket or anything that takes some time). It’s also handy for API services where the root route might not even exist yet.

#### Location:
- https://github.com/sst/ion/blob/8cad056b60deda591a55f8b4281992b8d035a6ad/platform/src/components/aws/service.ts#L452-L472

zammitjames:
I was able to get this working using transform:

```
transform: {
    target: {
      healthCheck: {
        enabled: true,
        path: '/health',
      },
    },
  },
```

PR #4968, opened 2024-10-25
Cluster: load balancer health check

zammitjames:
## Scope

This allows us to update the health check settings (including the path) of a service via:

```
cluster.addService('service', {
  healthCheck: {
    enabled: true,
    path: '/health',
  },
 ...
})
```

Solves https://github.com/sst/sst/issues/3925


michaelgmcd:
It looks like the "healthCheck" option is configuring both the target group health check and the container level one. In addition, the type doesn't seem to include "args.healthCheck.command" but uses it. I opened a similar PR for container-level health checks since the target group check can be defined via transform.

fwang:
@zammitjames so there are 2 types of health checks
- container health checks
- load balancer health checks

Looking at the PR, `path`, `retries`, `healthyThreshold`, and `unhealthyThreshold` are load balancer health checks arguments. But they are set on the container.

So to clarify, u r looking to set the load balancer health checks right?

zammitjames:
Yeah exactly - I’d like to modify the health check for the load balancer.

fwang:
Got it. Lemme move it around.

fwang:
Released in 3.2.69. Docs here https://sst.dev/docs/component/aws/cluster#loadbalancer-healthcheck

fwang:
Released v3.2.73 with a minor tweak in how health check is specified - https://sst.dev/docs/component/aws/cluster#loadbalancer-health

/cc @zammitjames @michaelgmcd

PR #4976, opened 2024-10-25
fix: point to sst repo in global install script

### LINKED ISSUE #4975: Bug: global install script fails for version 3.2.41 onwards after the repo move
```
curl -fsSL https://sst.dev/install | VERSION=3.2.56 bash
Downloading SST ❍ ion version: 3.2.56 ...
######################################################################## 100.0%

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

Was going to push a PR just changing the repo in the install script but would break peoples CI if they are pinned to a pre 3.2.41 version which is maybe fine? Or do something more complex and use the old ion repo if its a pre 3.2.41 version.

aatecey:
resolves https://github.com/sst/sst/issues/4975

This will break pre-3.2.41 pinned installs but probably not the worst as a once off.

### DIFF
--- a/install
+++ b/install
@@ -39,15 +39,15 @@ INSTALL_DIR=$HOME/.sst/bin
 mkdir -p "$INSTALL_DIR"
 
 if [ -z "$requested_version" ]; then
-    url="https://github.com/sst/ion/releases/latest/download/$filename"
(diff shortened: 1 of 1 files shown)


subvertallchris:
It's 2:43 AM and I am trying to fix a critical production bug. This did break my CI running 3.1.51. I am very sad.

PR #4981, opened 2024-10-26
docs: Update Hono examples

garysassano:
Update the examples:

- Swapped `Response` with `c.text()` [as recommended in Hono docs](https://hono.dev/docs/api/context#body).
- Added `Content-Type` for the upload operation so that it's actually retrieved when hitting the `/` route with GET method.


jayair:
Thanks! I'll try it out.

PR #4984, opened 2024-10-27
docs: update CLI command to `sst shell`

Newbie012:
Update testing docs to replace `sst load-config` with `sst shell`

The `sst load-config` command is no longer available, as it's been replaced by `sst shell`.


jayair:
Hmm this is one of our old blog posts for sst v2. Did you find this through a google search?

Newbie012:
Oh you're right, I guess I did. It seems like there's no reference on how to test code that uses SST resources.

jayair:
Yeah I'll close this for now.

Though we probably should add a warning to these posts that they are for an older version.

PR #4990, opened 2024-10-28
Support container level health checks in ECS

michaelgmcd:


jayair:
Frank added support for this right, so we can close this?

michaelgmcd:
Frank added support for just "load balancer" / "target group" health checks. The discussion on the PR goes into more detail, but this is container-level. More of a nice to have if you're using a load balancer, but can't hurt.

fwang:
Thanks @michaelgmcd!

Renamed the field to `health`.

Will be in the next release.

PR #4995, opened 2024-10-29
fix: allow false when true is excepted

lodmfjord:
This pr changes ```Input<true>``` to ```Input<boolean>```

## Reason

Writing conditional is weird when only true is accepted, for example:

```
deletionProtection: isFeatureDeployment ? undefined : false,

```

or 

```
const extraValues = isFeatureDeployment ? {} : {deletionProtection: true}


// Later in code
{
...extraValues
}
```

instead of doing something like:

```
deletionProtection: !isFeatureDeployment
```

## Why this might be a bad idea?

There might be a reason why false is not allowed…


FridrikMProgramm (approved):
LGTM

fwang:
Thanks @lodmfjord! Will be in the next release.

PR #5029, opened 2024-11-06
Ensure softwareTokenMfaConfiguration is not set when not enabled on cognito user pool

### LINKED ISSUE #5022: AWS Cognito: Invalid MFA configuration given
```
✕  Failed    
   UserPool sst:aws:CognitoUserPool → UserPoolUserPool aws:cognito:UserPool
     sdk-v2/provider2.go:520: sdk.helper_schema: setting Cognito User Pool (us-east-1_...) MFA configuration: operation error Cognito Identity Provider: SetUserPoolMfaConfig, https response error StatusCode: 400, RequestID: 71d38a30-c519-4d9f-9c65-b8850d691caa, InvalidParameterException: Invalid MFA configuration given, can't turn off MFA and configure an MFA together.: provider=aws@6.57.0
```
Stepped versions of sst back from 3.2.76 and was able to successfully redeploy an existing project with 3.2.73.

Don't have the time to dig into it myself this week, but opening the ticket so y'all know about the issue.  Will work on it next week if still open.

thelegendtubaguy:
Fixes #5022

### DIFF
--- a/platform/src/components/aws/cognito-user-pool.ts
+++ b/platform/src/components/aws/cognito-user-pool.ts
@@ -1,4 +1,4 @@
-import { ComponentResourceOptions, all, output } from "@pulumi/pulumi";
+import { ComponentResourceOptions, Output, all, output } from "@pulumi/pulumi";
 import { Component, Prettify, Transform, transform } from "../component";
 import { Input } from "../input";
 import { Link } from "../link";
@@ -420,7 +420,7 @@ interface CognitoUserPoolRef {
  */
 export class CognitoUserPool extends Component implements Link.Linkable {
   private constructorOpts: ComponentResourceOptions;
-  private userPool: cognito.UserPool;
+  private userPool: Output<cognito.UserPool>;
 
   constructor(
     name: string,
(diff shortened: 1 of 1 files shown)


platform/src/components/aws/cognito-user-pool.ts
```diff
@@ -543,11 +543,9 @@ export class CognitoUserPool extends Component implements Link.Linkable {
             ),
             smsAuthenticationMessage: args.smsAuthenticationMessage,
             smsConfiguration: args.sms,
-            softwareTokenMfaConfiguration: output(args.softwareToken).apply(
-              (v) => ({
-                enabled: v ?? false,
-              }),
-            ),
+            softwareTokenMfaConfiguration: args.softwareToken
```
brunocleite, line 546: `args.softwareToken` is of type `Input<boolean`.

This means that if it is set as either `true` or `false` it would be truthy. 

In this case seems to me that `enabled` should be set to softwareToken, like so:

`output(args.softwareToken).apply((token) => ({ enabled: token }))`

brunocleite (approved):

brunocleite:
I have tested here and looks good, thanks

fwang:
Thanks @thelegendtubaguy!

Made a minor change to handle `args.softwareToken` being false.

PR #5041, opened 2024-11-08
Update `bun.mdx`

NivaldoFarias:
this pr may change the trajectory of humanity

### DIFF
--- a/www/src/content/docs/docs/start/aws/bun.mdx
+++ b/www/src/content/docs/docs/start/aws/bun.mdx
@@ -72,13 +72,13 @@ bunx sst init
 bun install
 ```
 
-This'll create a `sst.config.ts` file in your project root and install SST.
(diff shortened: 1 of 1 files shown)


jayair:
Haha thanks

Though I think the a vs an might actually be okay in these cases.

www/src/content/docs/docs/start/aws/bun.mdx
```diff
@@ -72,13 +72,13 @@ bunx sst init
 bun install
 ```
 
-This'll create a `sst.config.ts` file in your project root and install SST.
+This'll create aa `sst.config.ts` file in your project root and install SST.
```
nurulhudaapon, line 75: @jayair aa --> an 😄

NivaldoFarias: HOW DID I NOT SEE THIS 😭

Issue #3970, opened 2024-10-13
BUG: Mention Cloudflare Pitfalls and Discrepancies in SST docs.


prabal-rje:
PITFALL 1:

Please mention somewhere in SST docs that initializing outside of *a* function in Cloudflare workers is going to cause issues.

```javascript
/*
This is merely an EXAMPLE to demo a problem - do NOT publish your secrets on a public API
*/

// CORRECT WAY

function get_some_secret() {
    return Resource.SomeSecret.value;
}

export default {
    async fetch(request: Request, env: any, ctx: any) {
        return new Response(JSON.stringify({"config_val": get_some_secret()})), {
            headers: { 'Content-Type': 'application/json'  }, 
        });  
    }, 
}; 

// WRONG WAY

const x = Resource.SomeSecret.value;

export default {
    async fetch(request: Request, env: any, ctx: any) {
        return new Response(JSON.stringify({"config_val": x})), {
            headers: { 'Content-Type': 'application/json'  }, 
        });  
    }, 
}; 
```

zxt-tzx:
> BUG 1:
> 
> Linking with `sst.cloudflare.Queue` simply doesn't work and returns the following error(s).
> 
> ```js
> export const queue = new sst.cloudflare.Queue("MyQueue"); 
> 
> export const worker = new sst.cloudflare.Worker("MyWorker", {
>   handler: "packages/functions/src/some_worker_handler.ts",
>   url: true,      
>   link: [queue],          
> });      
> ```
> 
> ```shell
> ~  Deploy
> 
> |  Error       MyWorkerScript cloudflare:index:WorkerScript
> |  cloudflare:index/workerScript:WorkerScript resource 'MyWorker
> Script' has a problem: Missing required argument. The argument "
> queue_binding.0.binding" is required, but no definition was foun
> d.. Examine values at 'MyWorkerScript.queueBindings'.
> |  Error       MyWorkerScript cloudflare:index:WorkerScript
> |  cloudflare:index/workerScript:WorkerScript resource 'MyWorker
> Script' has a problem: Invalid or unknown key. Examine values at
>  'MyWorkerScript.queueBindings'.
> 
> ✕  Failed
>    MyWorkerScript cloudflare:index:WorkerScript
>    cloudflare:index/workerScript:WorkerScript resource 'MyWorker
> ```

did you figure this out? I suspect it's because the Cloudflare Queue construct has not been released officially

[...]

PR #5045, opened 2024-11-10
fix: align Queue binding property with Cloudflare's API

doichev-kostia:
According to the issue #3970 

Cloudflare's Worker bindings have inconsistencies in their API:
1. Binding name registration varies:
   - D1, KV, R2: uses 'name' property
   - Queue, Hyperdrive: uses 'binding' property

2. Resource identifier format varies:
   - KV, D1: requires resource ID
   - Queue, R2: requires resource name

Note: this pr only solves the binding, the queue subscription is something that I would only try doing after having some feedback from the SST team and their vision of the design

Reproduction sample (should be created locally, of course)
https://stackblitz.com/edit/stackblitz-starters-dxtpht?file=index.ts 

sst.config.ts
```ts
export default $config({
  app(input) {
    /* */
  },
  async run() {
    const queue = new sst.cloudflare.Queue('LogQueue');
    const worker = new sst.cloudflare.Worker('Worker', {
      url: true,
      link: [queue],
      handler: 'index.ts',
    });

    return {
      url: worker.url,
    };
  },
});
```

index.ts
```ts
export default {
  async fetch(request, env, ctx): Promise<Response> {
    let log = {  url: request.url,  method: request.method, headers: Object.fromEntries(request.headers) };
    await Resource.LogQueue.send(log);
    return new Response('Success!');
  },
  async queue(batch, env): Promise<void> {
    let messages = JSON.stringify(batch.messages);
    console.log(`consumed from our queue: ${messages}`);
  },
} satisfies ExportedHandler;
```

### DIFF
--- a/platform/src/components/cloudflare/binding.ts
+++ b/platform/src/components/cloudflare/binding.ts
@@ -17,6 +17,13 @@
  */
 
 import { Input } from "../input";
+import type {
+  WorkerScriptD1DatabaseBinding,
+  WorkerScriptKvNamespaceBinding,
+  WorkerScriptQueueBinding,
+  WorkerScriptR2BucketBinding,
+  WorkerScriptServiceBinding
+} from "@pulumi/cloudflare/types/output";
 
 export interface KvBinding {
   type: "kvNamespaceBindings";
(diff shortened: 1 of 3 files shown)


platform/src/components/cloudflare/binding.ts
```diff
@@ -78,3 +90,38 @@ export function binding<T extends Binding["type"]>(input: Binding & {}) {
     properties: input.properties as Extract<Binding, { type: T }>["properties"],
   };
 }
+
+/**
+ * @param bindingName - The name of the global variable for the binding in your Worker code.
+ * @param cfBinding
+ */
+export function toWorkerScriptBinding(bindingName: Input<string>, cfBinding: ReturnType<typeof binding>): WorkerScriptKvNamespaceBinding | WorkerScriptServiceBinding | WorkerScriptQueueBinding | WorkerScriptR2BucketBinding | WorkerScriptD1DatabaseBinding {
```
doichev-kostia, line 98: @fwang, I'm not sure what convention you have within the team. This could either be solved more elegantly (without type casting) by playing around with generics and relying more on types, or loosen the types a bit and add regression tests. I could try to solve it in both ways, but I would like some validation from your side

doichev-kostia:
Discord thread https://discord.com/channels/983865673656705025/983865673656705028/1306690249493184613

platform/src/components/cloudflare/binding.ts
doichev-kostia: I see that @thdxr was assigned, then I would like to forward the question to you.

PR #5052, opened 2024-11-12
Fix: realtime doesn't get correct endpoint when using a custom provider

nicholasgubbins:
When you create a realtime component using a custom provider (e.g. using a different region), the component is all correctly created, but. the output is incorrect / cannot be found. This is because the function iot.getEndpointOutput(), was missing the parent pulumi options. This fixes that


nicholasgubbins:
@thdxr any chance we can get this reviewed & released?

fwang:
Thanks @nicholasgubbins!

Will be in the next release.

PR #5053, opened 2024-11-12
bug: update path for drizzle t3 link (Docs only update)

aisflat439:
The current path of the t3 link [goes here](https://sst.dev/docs/start/aws/drizzle/docs/examples/#t3-stack-in-aws) which is a 404.

image

This just adds a `/` fixing the link


jayair:
Thanks!

PR #5058, opened 2024-11-12
docs: Add quotes around $default route

nguyendon:
The route variable needs to be a string otherwise you get `ReferenceError: $default is not defined`.


jayair:
Thanks!

PR #5081, opened 2024-11-18
Fix postgres.get passing opts.

izakfilmalter:
If postgres is created with a provider in a different region, it will fail to find `proxySecret`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ Pulumi.yaml
 .netlify
 .env
 tmp
+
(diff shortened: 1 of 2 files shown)


fwang:
Thanks @izakfilmalter!

I didn't have permission to ur branch. So i ended up copying ur changes over to the commit above.

fwang:
Released in v3.3.36

izakfilmalter:
All good!

PR #5090, opened 2024-11-20
Fix: remove duplicate example

moosashah:
SQS example seems to be duplicated.

The only difference being one shows the dlq code as well. 

Options:
1. Modify name of the dlq example to clarify that it's for dlq
2. Keep DLQ example only and remove base queue example. 
3. Leave as is 🥲


jayair:
Thanks! Can we keep the base example and move the comment/doc to it instead? We can then remove the dlq one.

vimtor:
closing in favour of #5819

PR #5107, opened 2024-11-23
Clarify Postgres `storage` docs

vimtor:
The current `storage` option for the `Postgres` component [sets the `maxAllocatedStorage`](https://github.com/vimtor/sst/blob/e5f4b6ce8c2c7ba44307091df7e5683e05e148fa/platform/src/components/aws/postgres.ts#L463-L463) not the [`allocatedStorage`](https://www.pulumi.com/registry/packages/aws/api-docs/rds/instance/#allocatedstorage_nodejs)

I thought the docs could be a bit more explicit about that.


jayair:
Good callout!

PR #5109, opened 2024-11-23
Add confirmation before running `sst dev` on a non-local stage

vimtor:
The other day, I accidentally ran `sst dev --stage production` instead of `sst tunnel --stage production`. I don't know if it was because I panicked and canceled it, but it messed with the Pulumi state to the point that I had to create a new AWS account to make my life easier.

I don't know if there is even a use case for running `sst dev` on a specific stage, but I figured it would make sense to have a confirmation before running it if the stage is not the personal one.

CleanShot 2024-11-23 at 09 37 15@2x

### DIFF
--- a/cmd/sst/mosaic.go
+++ b/cmd/sst/mosaic.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/charmbracelet/huh"
 	"log/slog"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


darioarjona24:
Hola soy nuevo aquí. Mi nombre es dario arjona

jayair:
Yeah this is on our roadmap. Will review.

jayair:
Actually our implementation will probably be a bit different. We'll show a warning if you are flipping between dev and deploy on the same stage.

vimtor:
By warning, do you mean a confirmation? Considering that the command starts deploying from the gecko, I would prefer a confirmation. For some side projects where I don't have autodeploy, I do run `dev` and `deploy` from my machine. I think the issue arises when you run `sst dev` with a stage parameter. Is there a use case for doing that?

Also, feel free to close this PR. It was just a quick thing in case it was helpful to you 😄

jayair:
Well we basically don't want you to accidentally switch between dev and deploy for the same stage. But yeah we'll take a look.

vimtor:
closing in favor of https://github.com/anomalyco/sst/pull/6559

PR #5110, opened 2024-11-23
fix(cli): ensure sst upgrade only replaces entry in package.json dependencies

### LINKED ISSUE #5098: `sst upgrade` replaces any package.json entry matching `"sst": ...`
This issue is fairly niche and likely doesn't need to be highly prioritized, but flagging nonetheless.

if you have a `package.json` that has an entry for `"sst": "..."` elsewhere in addition to `dependencies` (e.g. in `scripts`), running `sst upgrade` will replace that entry with the latest version of sst. Here's a repo with reproduction steps: https://github.com/mgramigna/sst-packagejson-repro

Before running `bun sst upgrade`
```json
{
	"name": "sst-packagejson-repro",
	"module": "index.ts",
	"type": "module",
	"scripts": {
		"sst": "sst --"
	},
	"dependencies": {
		"sst": "3.3.27"
	}
}
```

After running `bun sst upgrade`
```json
{
	"name": "sst-packagejson-repro",
	"module": "index.ts",
	"type": "module",
	"scripts": {
		"sst": "3.3.27"
	},
	"dependencies": {
		"sst": "3.3.27"
	}
}
```

It's because [this regex](https://github.com/sst/sst/blob/613c1b2c86075ffd31e10ef334fce8f59c1b2a0d/pkg/global/upgrade.go#L125) is used in a replaceAll [here](https://github.com/sst/sst/blob/613c1b2c86075ffd31e10ef334fce8f59c1b2a0d/pkg/global/upgrade.go#L141). I doubt this really matters in the long run if you're using a global install of `sst` cli or running it with bunx/npx/etc. as there's no need for a scripts entry in that case, or you can just call it something different if you really want.

mgramigna:
Fixes #5098 

Tested the following scenarios leave the `"sst"` entry in `scripts` untouched when running `sst upgrade` with a local cli binary after this change:

```json
	"scripts": {
		"sst": "echo \"hello\""
	},
	"dependencies": {
		"sst": "3.3.27"
	}
```

```json
	"scripts": {
		"sst": "echo \"hello\""
	},
	"dependencies": {
		"sst": "latest"
	}
```

```json
	"scripts": {
		"sst": "echo \"hello\""
	},
	"devDependencies": {
		"sst": "3.3.27"
	}
```

### DIFF
--- a/pkg/global/upgrade.go
+++ b/pkg/global/upgrade.go
@@ -122,7 +122,9 @@ func UpgradeNode(existingVersion string, nextVersion string) (map[string]string,
 	if err != nil {
 		return result, err
 	}
-	re := regexp.MustCompile(`"sst": "[^"]+"`)
+
+	// Ensure we only replace an existing "sst" entry in dependencies/devDependencies of package.json
+	re := regexp.MustCompile(`(("dependencies"|"devDependencies")\s*:\s*{[^}]*"sst"\s*:\s*)"[^"]*"`)
 	for _, file := range files {
 		if strings.HasPrefix(file, ".sst") {
 			continue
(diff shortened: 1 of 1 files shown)


klaemo:
Just encountered this as well, looking forward to this fix :)

PR #5125, opened 2024-11-25
update function to allow node 22.x

natac13:
and update the default to node 22.x

https://aws.amazon.com/about-aws/whats-new/2024/11/aws-lambda-support-nodejs-22/


Chenalejandro:
We could also update the nodejs version of the following two files: https://github.com/sst/sst/blob/dev/platform/src/components/aws/nextjs.ts https://github.com/sst/sst/blob/dev/platform/src/components/aws/ssr-site.ts

fwang:
Released in v3.3.42

Thanks @natac13!

PR #5126, opened 2024-11-25
Update lambda to allow Python 3.13 runtime

natac13:
https://aws.amazon.com/about-aws/whats-new/2024/11/aws-lambda-support-python-313/

### DIFF
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -269,6 +269,7 @@ export interface FunctionArgs {
     | "python3.10"
     | "python3.11"
     | "python3.12"
+    | "python3.13"
   >;
   /**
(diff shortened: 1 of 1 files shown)


jayair:
I think there's another big Python PR being worked on? Maybe mention this in there?

subssn21:
I just tested pushing Python 3.13 Lambdas using the new infrastructure and they work. It is safe to add this to the list of supported Python runtimes.

CorentinDoue:
can you please merge this? I think nothing is blocking

vimtor (approved):
thanks for your contribution @natac13 

tested this against both aws python examples and it seems to be working

we'll merge @subssn21's pr but in the meantime we can merge this one as well

Issue #5130, opened 2024-11-25
Excluding the encoding option when `fs.readFileSync` is called causes type error `Argument of type 'Buffer<ArrayBufferLike>' is not assignable to parameter of type 'BinaryLike'.`


the-simian:
"bun": "1.1.36"
"typescript": "^5.7.2"
"sst": "3.3.28",

Running `tsc --pretty` results in the following type error.

```sh

.sst/platform/src/components/aws/function.ts:1946:23 - error TS2345: Argument of type 'Buffer<ArrayBufferLike>' is not assignable to parameter of type 'BinaryLike'.
  Type 'Buffer<ArrayBufferLike>' is not assignable to type 'Uint8Array<ArrayBufferLike> | DataView<ArrayBufferLike>'.
    Type 'Buffer<ArrayBufferLike>' is not assignable to type 'Uint8Array<ArrayBufferLike>'.
      The types returned by 'slice(...).entries()' are incompatible between these types.
        Type 'IterableIterator<[number, number]>' is missing the following properties from type 'ArrayIterator<[number, number]>': map, filter, take, drop, and 9 more.

1946           hash.update(await fs.promises.readFile(zipPath));

.sst/platform/src/components/aws/ssr-site.ts:542:31 - error TS2345: Argument of type 'Buffer<ArrayBufferLike>' is not assignable to parameter of type 'BinaryLike'.
  Type 'Buffer<ArrayBufferLike>' is not assignable to type 'Uint8Array<ArrayBufferLike> | DataView<ArrayBufferLike>'.
    Type 'Buffer<ArrayBufferLike>' is not assignable to type 'Uint8Array<ArrayBufferLike>'.
      The types returned by 'slice(...).entries()' are incompatible between these types.
        Type 'IterableIterator<[number, number]>' is missing the following properties from type 'ArrayIterator<[number, number]>': map, filter, take, drop, and 9 more.

542                       .update(content)
                                  ~~~~~~~

.sst/platform/src/components/aws/ssr-site.ts:1030:21 - error TS2345: Argument of type 'Buffer<ArrayBufferLike>' is not assignable to parameter of type 'BinaryLike'.
  Type 'Buffer<ArrayBufferLike>' is not assignable to type 'Uint8Array<ArrayBufferLike> | DataView<ArrayBufferLike>'.
    Type 'Buffer<ArrayBufferLike>' is not assignable to type 'Uint8Array<ArrayBufferLike>'.
[...]

PR #5131, opened 2024-11-25
chore: add utf-8 encoding argument on all fs.readFile usages

the-simian:
This causes the the following typescript issue to not happen:

```
 Argument of type 'Buffer<ArrayBufferLike>' is not assignable to parameter of type 'BinaryLike'.
  Type 'Buffer<ArrayBufferLike>' is not assignable to type 'Uint8Array<ArrayBufferLike> | DataView<ArrayBufferLike>'.
    Type 'Buffer<ArrayBufferLike>' is not assignable to type 'Uint8Array<ArrayBufferLike>'.
      The types returned by 'slice(...).entries()' are incompatible between these types.
        Type 'IterableIterator<[number, number]>' is missing the following properties from type 'ArrayIterator<[number, number]>': map, filter, take, drop, and 9 more.
```

### DIFF
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -1943,7 +1943,7 @@ export class Function extends Component implements Link.Linkable {
 
           // Calculate hash of the zip file
           const hash = crypto.createHash("sha256");
-          hash.update(await fs.promises.readFile(zipPath));
+          hash.update(await fs.promises.readFile(zipPath, 'utf-8'));
           const hashValue = hash.digest("hex");
           const assetBucket = region.apply((region) =>
             bootstrap.forRegion(region).then((d) => d.asset),
--- a/platform/src/components/aws/ssr-site.ts
+++ b/platform/src/components/aws/ssr-site.ts
@@ -536,7 +536,7 @@ export function createServersAndDistribution(
                 ...(await Promise.all(
                   files.map(async (file) => {
                     const source = path.resolve(outputPath, copy.from, file);
-                    const content = await fs.promises.readFile(source);
+                    const content = await fs.promises.readFile(source, 'utf-8');
                     const hash = crypto
                       .createHash("sha256")
                       .update(content)
@@ -1028,7 +1028,7 @@ async function handler(event) {
                 }).forEach((filePath) =>
                   hash.update(
                     fs.readFileSync(
-                      path.resolve(outputPath, item.from, filePath),
+                      path.resolve(outputPath, item.from, filePath), 'utf-8'
                     ),
                   ),
                 );
--- a/platform/src/components/aws/static-site.ts
+++ b/platform/src/components/aws/static-site.ts
@@ -760,7 +760,7 @@ export class StaticSite extends Component implements Link.Linkable {
             ...(await Promise.all(
               files.map(async (file) => {
                 const source = path.resolve(outputPath, file);
-                const content = await fs.promises.readFile(source);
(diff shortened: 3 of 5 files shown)


the-simian:
resolves https://github.com/sst/sst/issues/5130

thdxr:
we'll likely accept this PR but can you explain what situation this typechecking error happens? if i run `tsc --noEmit` on our `platform` folder i can't recreate it

the-simian:
I'll explain what I can. This is a typechecking error that occurs in
```
     1  .sst/platform/src/components/aws/function.ts:1946
     2  .sst/platform/src/components/aws/ssr-site.ts:542
     2  .sst/platform/src/components/aws/static-site.ts:766
     1  .sst/platform/src/components/cloudflare/ssr-site.ts:138
     1  .sst/platform/src/components/cloudflare/static-site.ts:336
```
whenever the `fs.readFile` is uses (or `fs.promises.readFile`)    is used. That's the high level context that's explained in the PR's linked issue in more detail. 

Basically this error occurs 'out of the box' with the emitted SST code with the versions of `bun`, `@types/node`, `sst` and `typescript` I am using. I tried adding the latest `@types/node` as a troubleshooting step based on this [stack overflow suggestion](https://stackoverflow.com/questions/78790943/in-typescript-5-6-buffer-is-not-assignable-to-arraybufferview-or-uint8arr) but it didn't seem to matter. I've been relying on bun and its type libraries until the issues recently appeared. here's the relevant versions in my package.json:

```json
"@types/node": "^22.10.0",
"@types/bun": "^1.1.14",
"sst": "3.3.37",
"typescript": "^5.7.2"
```

This is in the context of a react-native project, so you'll see some expo-related stuff:

so to reproduce this on my end, its a pretty straigtforward process, I can

1. install my dependencies, including sst, which emits the .sst folder
`bun x expo install && sst install`
2.  check types
`tsc --pretty`

Then I see the errors in the associated issue, just as they appeared to me.

here is my compiler options in my tsconfig.json:
```json
{
  "extends": "expo/tsconfig.base",
  "compilerOptions": {
    "lib": ["ESNext", "DOM"],
    "target": "ESNext",
    "module": "ESNext",
    "moduleDetection": "force",
    "jsx": "react-native",
    "allowJs": true,
    "moduleResolution": "bundler",
    "allowImportingTsExtensions": true,
    "verbatimModuleSyntax": false,
    "noEmit": true,
    "strict": true,
    "skipLibCheck": true,
    "noFallthroughCasesInSwitch": true,
    "noUnusedLocals": false,
    "noUnusedParameters": false,
    "noPropertyAccessFromIndexSignature": false,
    "types": ["nativewind/types"],
    "paths": {
      "~/*": ["./*"]
    }
  }
  ```
  
This is the same configuration that's linting the sst-related things, such as my drizzle files, sst config and so-on. It was working just fine even back when SST was in an alpha state as 'ion', and only recently have I come into this issue after upgrading SST and typescript. I admit I've not been able to pin down exactly which version upgrade caused the issue in the first place. I experimented with downgrading typescript, sst, bun but I couldn't get back into a state where this wasn't happening. I strongly suspect its typescript versioning, but I don't know exactly what. The SO article specifies 5.6+, perhaps it is that.

Nontheless, I know that the fix is simple - including the UTF-8 encoding makes it so that typescript recognized the return type of the readFile function more-correctly and everything works as it should. Since that should be the default encoding anyways, its no harm to just indicate it where the function is used and therefore suppress this (probably typescript-layer-only) issue

Issue #5088, opened 2024-11-19
Python Function: UV can't build with SST dependency due to Git failure


madisonbullard:
I'm having trouble executing a Python lambda in sst dev. I'm getting an error from uv when I try to call the function: 
```
|  Build       packages/image-analysis/src/function.handler
|  +2562047h     × Invalid `--with` requirement
|  +2562047h     ╰─▶ Git operation failed
```

The error stems from trying to install SST as a dependency via `pyproject.toml`. If I remove the dependency, it builds without error.

I'm following the aws-python example from this repo, and including this in my `pyproject.toml`:
```toml
[tool.uv.sources]
sst = { git = "https://github.com/sst/sst.git", subdirectory = "sdk/python", branch = "dev" }
```

Seems related to [this](https://github.com/sst/sst/blob/bad21bf409eba1ed308aa2b89e0a8bb473b95e78/pkg/runtime/python/python.go#L321) in the python.go file, maybe?

jayair:
Hmm AFAIK we need that for referencing links in your functions.

theoneamendez:
+1 on fixing this issue. Would love for this to work, in the meantime I just copied and pasted https://github.com/sst/sst/blob/dev/sdk/python/src/sst/__init__.py into src/sst.py and then import as from sst import Resource to unblock myself since the code is not that long.

atwmarshall:
+1 on this issue - would love to see this resolved as caused quite a lot of pain and almost pushed our tech stack away from sst so may prevent other python users from adopting sst.

jayair:
This is an old issue, we've redone our python support since then. Open a new issue if you are still running into this?

Issue #5079, opened 2024-11-18
Python containers not building - `Dockerfile: no such file or directory`


nyacg:
Similar to a few other users I am unable to get the Python containers to build with the latest SST (v3.3.22) but it does work with (v3.2.55).

Seems to be caused by #4967

```
Dockerfile: no such file or directory
```

>  I cannot get the simple examples/aws-python-container to work given this MR. 
> 
> Error:
> container/.sst/artifacts/MyPythonFunction-src/Dockerfile: no such file or directory
> 
> I see in fact that the Dockerfile is at MyPythonFunction-src/src/Dockerfile not MyPythonFunction-src/Dockerfile.
> 
> Does work on v3.22.5
> 
> _Originally posted by @cosmic-ascendant in https://github.com/sst/sst/issues/4967#issuecomment-2439787147_
>

apetta:
Also facing this issue on the latest version, and I can confirm it works on `3.2.5`.  For mac M series users, don't forget to set architecture to `arm64` in `sst.config.ts` as the docker image is built locally.

Issue #4114, opened 2024-09-17
Python functions fail to build with mono repo structure / nested folders


cmanou:
It seems as though when you have a python function defined as follows

```
    new sst.aws.Function("MyPythonFunction", {
      handler: "projects/function-a/src/logic.handler",
      runtime: "python3.11",
      url: true,
    });
  ```
  
  You'll hit the following error during the build
  
```
|  Error: Failed to build function "projects/function-a/src/logic.handler": Error: ENOENT: no such file or directory, copyfile '/Users/chris/test-project/projects/function-a/pyproject.toml' -> '/Users/chris/test-project/.sst/artifacts/MyPythonFunction-src/Users/chris/test-project/projects/function-a/pyproject.toml'
|      at file:///Users/chris/test-project/.sst/platform/src/components/aws/function.ts:1378:23 {
|    promise: Promise { <rejected> [Circular *1] }
|  }
```

Guessing this needs to trim the target path to be of the same parent as the project file
https://github.com/sst/ion/blob/5b183e67ed7b9134cd05577b2d417fc7d482534b/platform/src/runtime/python.ts#L73C1-L76C7

**Comment by nyacg** [comment]
I also faced this issue today with v3.2.55 using python container runtime

**Comment by andycloke** [comment]
Same issue with python runtime (non container)

**Comment by andycloke** [comment]
> For zip deployments, one could do something ugly like this temporarily:
> 
>   async run() {
>     const functionA = new sst.aws.Function("FunctionA", {
>       python: {
>         container: false,
>       },
>       handler: "packages/functions/function_a/lambda.handler",
>       runtime: "python3.11",
>       url: true,
>       copyFiles: [{ from: "packages/functions/function_a/lambda.py", to: "packages/functions/function_a/lambda.py" }]
>     });
> 
> Which does allow allow the handler to run with its dependencies. But would break resource linking.

this worked for me to get the lambda function to run, but importing any packages failed

**Comment by aLjTap** [comment]
### Python Function:
```js
export const lambdaAI = new sst.aws.Function("LambdaAI", {
  handler: "src/module/api.handler",
  runtime: "python3.11",
  url: true,
});
```
### Project Tree(as in the simple example)

```
.
|--- module
|    |--- __init__.py
|    |--- api.py
|
|--- pyproject.toml
|--- uv.lock
|--- sst.pypi
```
### .toml:

```toml
[project]
name = "module"
version = "0.1.0"
description = "Lambda function handlers"
[...]

Issue #4062, opened 2024-09-24
Allow a path setting for Python Functions


subssn21:
Any way we can add a parameter to tell things to include code from other directories / set the python path differently, It looks like right now, it is all relative to the handler function and my v2 app (and I am sure others) have their library code in a different directory than their handlers. 
As an example

- Project Root
- - backend
- - - lib
- - - - models
- - - - utils
- - - services
- - - - api
- - - - - handler1
- - - - - handler2
- - - - dynamodbstream
- - - - - handler1
- - - - - handler2
- - frontend
- - sst.config.ts

Does that make sense?

_Originally posted by @subssn21 in https://github.com/sst/sst/issues/4669

subssn21:
Closed by pull request #5132

PR #5132, opened 2024-11-26
Python Support Improvements

walln:
Redesigned the python support to move towards leveraging uv's new build system when lambda builds become a first-class citizen. These changes make the builds much more reliable, faster, and configurable. The caveat being the requirement of using uv workspaces. This change is breaking, but necessary to support the sorts of folder structures that have been commonly requested in several issues and in the discord. The migration is incredibly simple, should only take a few minutes and is also more in line with the direction python packaging is going.

Also this change moves to using the RPC based runtime builder so there is no more duplication of function building!

Additionally, this PR fixes the sdk which was broken with the move to using an encrypted links file, ensured support for non-lambda environments, and improved the type hints.

All the examples have been updated with more thorough  code and folder structures, up to date readmes, and a few dependency updates. The new default dockerfile is simpler, less error prone and some race condition bugs should be solved.


walln:
fixes: https://github.com/sst/sst/issues/5088
should fix: https://github.com/sst/sst/issues/5079
fixes: https://github.com/sst/sst/issues/4002
does not fix this but will in a follow up pr when a new uv feature gets released: https://github.com/sst/sst/issues/4003
fixes: https://github.com/sst/sst/issues/4062 adding first class support for local dependencies
fixes: https://github.com/sst/sst/issues/4114 same as above

JacksonBowe:
Reading through the examples; this is a fantastic improvement to dev experience.

pkg/runtime/python/python.go
```diff
@@ -345,150 +158,288 @@ func (r *PythonRuntime) Run(ctx context.Context, input *runtime.RunInput) (runti
...
-		return fmt.Errorf("failed to create destination directories for %s: %v", dst, err)
+		slog.Error("failed to run uv sync", "error", err, "output", string(syncOutput))
+		return nil, fmt.Errorf("failed to run uv sync: %v", err)
 	}
+	slog.Error("uv sync output", "output", string(syncOutput))
 
-	// Create the destination file
-	destinationFile, err := os.Create(dst)
+	outputRequirementsFile := filepath.Join(input.Out(), "requirements.txt")
+	exportCmd := process.CommandContext(ctx, "uv", "export", "--all-packages", "--output-file="+outputRequirementsFile, "--no-emit-workspace")
```
doron98: Hi, thanks for all your work. Why do you export all the dependencies of all the packages of the workspace and not only those of the package where the handler belongs to ? Which can be done with the `--package <package_name>` option

walln: Because that does not build other local dependencies. For example if functions depends on core specifying functions would only build functions. This is temporary and I will update when UV supports the build process we need.

doron98: From what I understand from the code, maybe I am wrong. 
Local packages are included thanks to the part of the code that builds them and then extracts them to place them next to the handler. Here, in the export command we create the requirements file for all non-local dependencies (hence the no-emit-workspace option), so we can restrict them to the handler package only.

walln: Right but since the build does all local packages the install should also do all local packages - I'd love to tree shake packages as much as possible but I think implementing that with current apis will be very error prone. The change in direction with the build apis is to accommodate the wide range of configurations that have been requested. There is discussion on the uv side to support the kind of workflow you are suggesting and when it gets added I will update this so we minimize the bundles as much as possible.

apetta:
This would be a tremendous improvement to python support! waiting eagerly 😁

pkg/runtime/python/python.go
doron98: Yes I agree for the local packages but my concern is about external packages. I really don't see why we should include those that does not depends of our handler's package. I played a little bit with an uv workspace and test manually by typing each command following the steps in your code except the export restrict to the handler package and I obtain what I think we want: all local packages (coming from the build+extraction not the export) and all the external package that my handler depends on (coming from the export). I don't see what will break if we don't include external dependencies of the local packages the handler's package don't depends on.

cgilly2fast:
Would this PR allow for deploying python lambda functions with SST?

pkg/runtime/python/python.go
walln: It was my understanding that `uv export --package <name>` would not resolve transitive dependencies for local dependencies. I may be wrong and if that is the case then yes we can absolutely only install the minimum necessary external packages. I will play around with making that work later.

walln:
> Would this PR allow for deploying python lambda functions with SST?

Yes. SST currently has very basic python support for lambda functions this PR is an overhaul to the implementation that makes the experience much better.

pkg/runtime/python/python.go
doron98: It is something that have tested and yes transitive dependencies of locals dependencies are included even if we exclude them with the no-emit-workspace option.

walln: Thanks for figuring this out, it should work as desired now!

subssn21:
Can't wait to try this

cgilly2fast:
Just read through the code on this, very excited for it! Thank you @walln

SamuelLeapifai:
oooh, this looks like something I've been waiting for, thank you so much @walln 🙌

doron98:
@walln if there is a way with uv to export a requirements file without path dependencies and copy paste those path dependencies in a specific directory: 1) we will not need to use uv workspace and 2) we will able to put the right path dependencies in the bundle ?

walln:
> @walln if there is a way with uv to export a requirements file without path dependencies and copy paste those path dependencies in a specific directory: 1) we will not need to use uv workspace and 2) we will able to put the right path dependencies in the bundle ?

Yes it is possible to do that I believe. The problem is that if we use path dependencies and do not export them in the requirements file, then its up to us to build the graph resolver that can determine which path dependencies need to be included. This would be a lot simpler if uv could build all transitive dependencies for a single target package. Theres some open discussion for this but its not implemented yet. I am leaning towards being slightly opinionated that we require the use of uv workspaces in order to provide a really simple and powerful experience. I also want to keep the build system here simple so we can leverage improvements on the uv side.

cosmic-ascendant:
@walln Thank you for adding in the custom dockerfile example and preserving that functionality. Being able to do so was key for me to be able to enable streaming lambda responses, which requires wrapping the lambda in AWS's lambda-web-adapter and doing things a little differently. I do think the result is clean and general purpose though. (Working example here: https://github.com/cosmic-ascendant/sst/tree/lambda-web-adapter/examples/aws-python-lambda-web-adapter).

We may want to include a `requirements.txt` in the example for clarity.

+1 On using uv workspaces. I'm finding that for other projects, the more I migrate towards uv the more maintainable my code has been.

Do you know if live lambda dev works with a custom Dockerfile?

Edit:
No need for including a `requirements.txt`. I now see it is auto generated with `uv export` during the build process.

walln:
> @walln Thank you for adding in the custom dockerfile example and preserving that functionality. Being able to do so was key for me to be able to enable streaming lambda responses, which requires wrapping the lambda in AWS's lambda-web-adapter and doing things a little differently. I do think the result is clean and general purpose though. (Working example here: https://github.com/cosmic-ascendant/sst/tree/lambda-web-adapter/examples/aws-python-lambda-web-adapter).
> 
> 
> 
> We may want to include a `requirements.txt` in the example for clarity.
> 
> 
> 
> +1 On using uv workspaces. I'm finding that for other projects, the more I migrate towards uv the more maintainable my code has been.
> 
> 
> 
> Do you know if live lambda dev works with a custom Dockerfile?
> 
> 
Yeah you could use a requirements file if you want for the docker build but I didn't want to make the example too complicated. Regarding live dev yes! You can use custom dockerfiles as the docker build only applies during deploys so everything works the same as a standard python live dev even with custom dockerfiles. I'm glad to see web adapter works I was thinking about seeing how to support streaming in a future PR so I'll keep web adapter in mind for that.

walln:
Container mode builds now cache layers in ECR the same as cluster. I removed the python specific code paths in the function component (rpc builds are awsome btw). Updated the docs for how to use python functions and added a note about uv not being installed by default anymore.

sockthedev:
pretty excited for the sst python experience to get boosted. thanks for putting in the elbow work here x

cosmic-ascendant:
Built and ran this PR and the flow is much much cleaner. Thanks @walln you're a legend. 

While I'm not able to get live lambda working (even across sst:latest, could be issue on my side), I was able to easily get lambda streaming deployed. It required very minimal modification. 

Any thoughts on being able to include a `.dockerignore` in the artifacts?

One aspect that I initially struggled on was installing a user package which is located in the same directory as the lambda itself. Something like this:
```
├── function_a
│   ├── Dockerfile
│   ├── pyproject.toml
│   ├── src
│   │   └── function_a
│   │       ├── __init__.py
│   │       ├── lambda.py
│   │       └── user_lib
│   │       │   ├── __init__.py
│   │       │   └── ...
```

The `requirements.txt` from `uv export` that's generated for a given function includes all external dependencies, but when bundled, the function's `pyproject.toml` is omitted so I wasn't able to modify the `Dockerfile` to include something like `uv pip install .`. There might be some other solutions in using relative imports or adding to systempath that would work, though I wasn't initially successful.

In the end, I decided to embrace the structure in the example where I keep my lambda handler lightweight and delegate all user packages as external dependencies, as is done with `core` in the example. I.e.,
```
├── function_a
│   ├── Dockerfile
│   ├── pyproject.toml
│   ├── src
│   │   └── function_a
│   │       ├── __init__.py
│   │       └── lambda.py
├── user_lib
│   ├── pyproject.toml
│   ├── src
│   │   └── user_lib
│   │       ├── __init__.py
│   │       ├── l...
```
 This way it gets picked up in `uv export`'s `requirements.txt`. I now see this as a better paradigm anyway -- separating the infra-level lambda handler from the app-level user_lib. The standalone `user_lib` also makes for better testing and could even reside in a different repo if need be, being installed as a git source. That said, I'll probably need to think more about a good method for dependency injection for things like sst linkables or resources that are referenced in `user_lib`.

Thanks again.

walln:
> Built and ran this PR and the flow is much much cleaner. Thanks @walln you're a legend.
> 
> While I'm not able to get live lambda working (even across sst:latest, could be issue on my side), I was able to easily get lambda streaming deployed. It required very minimal modification.
> 
> Any thoughts on being able to include a `.dockerignore` in the artifacts?
> 
> One aspect that I initially struggled on was installing a user package which is located in the same directory as the lambda itself. Something like this:
> 
> ```
> ├── function_a
> │   ├── Dockerfile
> │   ├── pyproject.toml
> │   ├── src
> │   │   └── function_a
> │   │       ├── __init__.py
> │   │       ├── lambda.py
> │   │       └── user_lib
> │   │       │   ├── __init__.py
> │   │       │   └── ...
> ```
> 
> The `requirements.txt` from `uv export` that's generated for a given function includes all external dependencies, but when bundled, the function's `pyproject.toml` is omitted so I wasn't able to modify the `Dockerfile` to include something like `uv pip install .`. There might be some other solutions in using relative imports or adding to systempath that would work, though I wasn't initially successful.
> 
> In the end, I decided to embrace the structure in the example where I keep my lambda handler lightweight and delegate all user packages as external dependencies, as is done with `core` in the example. I.e.,
> 
> ```
> ├── function_a
> │   ├── Dockerfile
> │   ├── pyproject.toml
> │   ├── src
> │   │   └── function_a
> │   │       ├── __init__.py
> │   │       └── lambda.py
> ├── user_lib
> │   ├── pyproject.toml
> │   ├── src
> │   │   └── user_lib
> │   │       ├── __init__.py
> │   │       ├── l...
> ```
> 
> This way it gets picked up in `uv export`'s `requirements.txt`. I now see this as a better paradigm anyway -- separating the infra-level lambda handler from the app-level user_lib. The standalone `user_lib` also makes for better testing and could even reside in a different repo if need be, being installed as a git source. That said, I'll probably need to think more about a good method for dependency injection for things like sst linkables or resources that are referenced in `user_lib`.
> 
> Thanks again.

Im glad that it works for you - I think the issue with live dev is unrelated and has to do with the appsync backend for live lambdas now. Regarding the local lib I do think using the separate package is the way to go - but you might be able to do some things with the custom dockerfile and copyFiles to make it work the initial way.

platform/functions/docker/python.Dockerfile
doron98, line 1: - We should compile bytecode with the `ENV UV_COMPILE_BYTECODE=1` or use the `--compile-bytecode` flag  it would improve start up time, hence reduce cold start https://docs.astral.sh/uv/guides/integration/docker/#compiling-bytecode
- If we are using the `ENV UV_COMPILE_BYTECODE=1` we could also use `ENV UV_SYSTEM_PYTHON=1` and remove the system flag
- Because we are not using uv in the final image we can just mount it at each invocation instead of keeping it in the container https://docs.astral.sh/uv/guides/integration/docker/#using-uv-temporarily
- For uv caching I don't know if what they advise apply to our cases: but either we don't cache with `ENV UV_NO_CACHE=1` to reduce image size or we try to follow their advice.  https://docs.astral.sh/uv/guides/integration/docker/#caching
- I am deploying with a custom dockerFile with python3.12, sst install from github and everything works well without the installation of gcc, do we need to install it ?

pkg/runtime/python/python.go
```diff
@@ -345,150 +161,321 @@ func (r *PythonRuntime) Run(ctx context.Context, input *runtime.RunInput) (runti
...
-	dir := filepath.Dir(absPath)
-
-	var pythonFiles []string
+	// If making a zip build or a local build then we need to install the dependencies and adjust the handler path
+	if !input.IsContainer || input.Dev {
 
-	// Walk through the directory
-	err = filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		// 5. Install the dependencies as a target
+		installCmd := process.CommandContext(ctx, "uv", "pip", "install", "-r", outputRequirementsFile, "--target", input.Out())
```
doron98: It lacks the correct platform option when deploying zip build to aws

pkg/runtime/python/python.go
doron98, line 1: While playing to make sst works with python, I notice that for live lambda we could run this command if we are using workspaces:
```sh
uv export --frozen --package <package_name> --directory <root-workspace-dir> | uv pip install --target <live-lambda-artifact-dir> -r - 
```
Or if not using workspace (the difference is due to the way path dependencies are resolved, within a workspace they are relative to the workspace root and outside of it they are relative to the package that install them)
```sh
uv export --frozen --directory <lambda-handler-pyproject-dir> | uv pip install --target <live-lambda-artifact-dir> -r - 
```
With this we have the correct dependencies bundle local+external, also with the way editable install works in python, it is something that should be tested, we could only rebuild when the uv.lock and handler file changes. 

The only reason why this not works with regular zip deployement, it is because the source code for local dependency (except the handler package itself) is not bring to the target directory even with `--no-editable`
(it also not working because we also need to add the python platform option to the command). 
We have a dist-info folder and _{package_name}.pth file that include the **absolute path** of the local package down to the src directory. As I see from this pr, it seems open to some manual work. Here the only work that is left to do is copy paste everything from the path in the _{package_name}.pth file to the lambda artifact dir. Which is less work that build and do the extraction for zip build.

thdxr:
why is this conflict so massive lol

walln:
I have no idea lmao - it was like 3 lines of changes

pkg/runtime/python/python.go
walln: we could set --python-platform=linux but I think this will not play nicely because pip will have to build deps and if on a different platform the build may break

platform/functions/docker/python.Dockerfile
walln: GCC is needed depending on the python version to be able to install git based deps. For whatever reason the aws published lambda base images have completely different setups based on the version. Some lack shells some have gcc some do not. I have not narrowed down the base dockerfile to cover every possible case - but i am willing to take feedback if people narrow down what is what for each distribution.

Regarding the env flags I agree I that this can be optimized - In will play with it and see what makes sense.

pkg/runtime/python/python.go
doron98: Most packages deployed pre-built binaries for specific platform, so we don't need to build anything. Without the platform option, packages will install the binaries relative to the platform making the installation. So to make our zip deployement work we need to specify the right python platform https://docs.aws.amazon.com/lambda/latest/dg/python-package.html#python-package-native-libraries

If we need to build ourself which is pretty uncommon now, one should use autodeploy to deploy their stack

platform/functions/docker/python.Dockerfile
doron98: I see, at least we know that for 3.12 and higher it is not needed

sockthedev:
ah, have you been merging instead of rebasing?

cosmic-ascendant:
@walln Running this on my own app that deviates from your provided example, found that `/platform/dist/python-runtime/index.py` failed to import external dependency `requests`. Easy enough to add to my `pyproject.toml` but worth considering checking for this or adding in by default. 

The above is for `sst dev` and the downstream `lambdaric_python_bridge`. That said, live lambdas are largely working now. For my lambda streaming python example, it gets a little weird as the handler is designed to start a uvicorn server, which doesn't play perfectly with live lambda, but this may be a topic for a different thread.

walln:
> @walln Running this on my own app that deviates from your provided example, found that `/platform/dist/python-runtime/index.py` failed to import external dependency `requests`. Easy enough to add to my `pyproject.toml` but worth considering checking for this or adding in by default. 
> 
> 
> 
> The above is for `sst dev` and the downstream `lambdaric_python_bridge`. That said, live lambdas are largely working now. For my lambda streaming python example, it gets a little weird as the handler is designed to start a uvicorn server, which doesn't play perfectly with live lambda, but this may be a topic for a different thread.

Great point yeah I forgot to add that check back to the new build system. I'll add it in.

walln:
> ah, have you been merging instead of rebasing?

Yeah that's the issue - but still and insane diff lol

pkg/runtime/python/python.go
```diff
@@ -345,150 +162,328 @@ func (r *PythonRuntime) Run(ctx context.Context, input *runtime.RunInput) (runti
...
-	var pythonFiles []string
-
-	// Walk through the directory
-	err = filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		// 5. Install the dependencies as a target
+		args := []string{"pip", "install", "-r", outputRequirementsFile, "--target", input.Out()}
+		if !input.Dev {
+			// If we are not in dev mode then we need to install the dependencies for the target platform
+			// which is amazon linux
+			args = append(args, "--python-platform", "linux")
```
doron98: Thanks for the change but it should also depends on the cpu: 
- x86_64-unknown-linux-gnu when deploying to x86_64
- aarch64-unknown-linux-gnu when deploying to arm64

walln:
Everything should be synced with dev and working properly now - I didn't dig too deep into the upstream changes but I corrected the lambda runtime api (fix the change to the route url) same as the other runtimes and everything seems to work the same as before. I think this is ready to merge unless anyone finds an issue I missed.

platform/functions/python-runtime/index.py
doron98, line 1: With pydantic, I had an error here probably related to the fact that it uses type annotations during runtime. I managed to fix it by noticing that the line to load the module from our handler was different here compared to what is done in the awslambdaric source code. Why is it done here? 

```py
spec = importlib.util.spec_from_file_location(module_name, module_path + ".py")
module = importlib.util.module_from_spec(spec)
spec.loader.exec_module(module)
```
Instead of this in awslambdaric: https://github.com/aws/aws-lambda-python-runtime-interface-client/blob/849e874de01776cb386c18fb2c1731d43cd2b2f4/awslambdaric/bootstrap.py#L53

```py
m = importlib.import_module(modname.replace("/", "."))
```
I fixed my bug by changing those 3 lines by this line, but shorltly after, I stop using live lambda with python so I didn't investigate more

walln: great point - this is an artifact of the old build system. Using import from spec allows importing from any path within the file system regardless of its module scope. This was previously useful and no longer is so I have moved to import module same as lambdaric. This should improve the compatibility.

cosmic-ascendant:
I'm having some trouble getting `copyFiles` to work here. Tested against the examples/aws-python-container. Since only the python source files are brought into the function package, I was hoping solve my issues by using copyFiles to copy some assets so that I could then load them into my custom docker image.

walln:
> I'm having some trouble getting `copyFiles` to work here. Tested against the examples/aws-python-container. Since only the python source files are brought into the function package, I was hoping solve my issues by using copyFiles to copy some assets so that I could then load them into my custom docker image.

Any errors? Or is copyFiles silently breaking? Does the expected behavior work with other runtimes?

cosmic-ascendant:
No errors, no logs I can find, silently breaking. Tested on node and python (non-container) runtimes, both of which worked. In those cases, it appears that the files do not get copied into the .sst/artifacts/fn-src but ultimately do show up in the .zip deployments. If there is an alternative way to include non-python-source files into a docker context without copyFiles, I'd be interested. Thanks.

cosmic-ascendant:
My findings - copyFiles has different paths for zip vs container builds:
1. Zip builds use `createZipAsset()` in `function.ts` which handles copying through the Node.js archiver
2. Container builds flow down to use the Go runtime's copy mechanism, but it's currently blocked in `runtime.go`
```go
[155]	// copying fiels still happens in node
	if !input.Dev && false {
```
My suggested fix: Added a `copyPath` helper function to allow copying entire directories. Added the copy logic directly in the Python container build path (since containers are Python-specific). Keeps the fix isolated and avoids modifying core runtime behavior. 

`pkg/runtime/python/python.go` - helper function.
```go
func copyFile(src, dst string) error {
...
}

func copyPath(from, to string) error {
	info, err := os.Stat(from)
	if err != nil {
		return fmt.Errorf("failed to stat source path: %v", err)
	}

	if info.IsDir() {
		return filepath.Walk(from, func(path string, info os.FileInfo, err error) error {
			if err != nil {
				return err
			}

			relPath, err := filepath.Rel(from, path)
			if err != nil {
				return fmt.Errorf("failed to get relative path: %v", err)
			}

			targetPath := filepath.Join(to, relPath)

			if info.IsDir() {
				return os.MkdirAll(targetPath, 0755)
			}

			return copyFile(path, targetPath)
		})
	}

	return copyFile(from, to)
}
```

`pkg/runtime/python/python.go` - container build branch
```go
	} else {
		// 5. Check if there is a Dockerfile in the handler directory
		// 	If not then copy over the default one from the platform directory
		workspaceDir, err := r.getWorkspaceDirectory(input)
		if err != nil {
			return nil, fmt.Errorf("failed to get workspace directory: %v", err)
		}

		slog.Info("checking for Dockerfile in workspace directory", "dir", workspaceDir)
		_, err = os.Stat(filepath.Join(workspaceDir, "Dockerfile"))
		if err != nil {
			slog.Error("workspace directory does not contain Dockerfile", "dir", workspaceDir)
			// Check if the Dockerfile exists in the platform directory
			defaultDockerfilePath := filepath.Join(path.ResolvePlatformDir(input.CfgPath), "/dist/dockerfiles/python.Dockerfile")
			_, err = os.Stat(defaultDockerfilePath)
			if err != nil {
				slog.Error("failed to check for Dockerfile in platform directory", "error", err)
				return nil, fmt.Errorf("failed to check for Dockerfile in platform directory: %v", err)
			} else {
				slog.Info("dockerfile exists in platform directory", "dir", path.ResolvePlatformDir(input.CfgPath))
			}

			slog.Info("copying default Dockerfile from platform directory to output directory", "dir", path.ResolvePlatformDir(input.CfgPath))

			// Copy over the default Dockerfile from the platform directory
			copyFile(defaultDockerfilePath, filepath.Join(input.Out(), "Dockerfile"))
			slog.Info("copied default Dockerfile to output directory", "dir", input.Out())
		} else {
			slog.Info("Dockerfile already exists in workspace directory", "dir", workspaceDir)
			copyFile(filepath.Join(workspaceDir, "Dockerfile"), filepath.Join(input.Out(), "Dockerfile"))
		}

		// 6. Copy additional files to the build context
		if len(input.CopyFiles) > 0 {
			slog.Info("copying additional files to build context", "dir", input.Out())
			for _, item := range input.CopyFiles {
				from, err := filepath.Abs(item.From)
				if err != nil {
					return nil, fmt.Errorf("failed to resolve source path: %v", err)
				}

				// Match runtime.go behavior for default 'to' path
				dest := filepath.Join(input.Out(), item.To)
				if item.To == "" {
					dest = filepath.Join(input.Out(), item.From)
				}

				if err := os.MkdirAll(filepath.Dir(dest), 0755); err != nil {
					return nil, fmt.Errorf("failed to create destination directory: %v", err)
				}

				if err := copyPath(from, dest); err != nil {
					return nil, fmt.Errorf("failed to copy path %s to %s: %v", item.From, item.To, err)
				}
			}
			slog.Info("finished copying additional files", "dir", input.Out())
		}

		adjustedHandler, err := r.adjustHandlerPath(input)
		if err != nil {
			return nil, fmt.Errorf("failed to adjust handler path: %v", err)
		}

		errors := []string{}
		sourcemaps := []string{}

		return &runtime.BuildOutput{
			Handler:    adjustedHandler,
			Errors:     errors,
			Sourcemaps: sourcemaps,
		}, nil
	}
}
```

Tested using `/examples/aws-python-container/sst.config.ts`
```typescript

export default $config({
	app(input) {
		return {
			name: "aws-python-container",
			home: "aws",
			providers: {
				aws: true,
			},
		};
	},
	async run() {
		const linkableValue = new sst.Linkable("MyLinkableValue", {
			properties: {
				foo: "Hello World",
			},
		});

		const base = new sst.aws.Function("PythonFn", {
			python: {
				container: true,
			},
			handler: "./functions/src/functions/api.handler",
			runtime: "python3.11",
			url: true,
			link: [linkableValue],
			copyFiles: [
				{
					from: "things",
				},
			],
		});

		const custom = new sst.aws.Function("PythonFnCustom", {
			python: {
				container: true,
			},
			handler: "./custom_dockerfile/src/custom_dockerfile/api.handler",
			runtime: "python3.11",
			url: true,
			link: [linkableValue],
			copyFiles: [
				{
					from: "things",
					to: "xthings",
				},
			],
		});

		return {
			base: base.url,
			custom: custom.url,
		};
	},
});

```

walln:
Yeah, that makes sense -  in that case, this isn't a Python-specific bug but instead a bug with non-zip archive deployments. The copyFiles is just patching the created zip rather than adding them to the build artifact. I wonder if, instead, it might make sense to change how copyFiles works to include them in the build artifact directory before the build step. That way, all runtimes work the same regardless of whether a zip is uploaded. Any thoughts @thdxr ?

sockthedev:
howdy - i'd love to start testing this. is there a runbook for how to use this build?

--

_update:_ ignore me, i actually decided to read the file called readme in the end

ashishpandey001:
looking to evaluate this with certain libraries and tools, what would be the recommended approach to use this?

SamuelLeapifai:
Hi, @doron98 @thdxr can you please tell us what is the status of this ticket? From the comments it seems it's ready for merging (minus recent conflicts). There hasn't been new comments from contributors for a month. Are you still waiting for some additional changes? Walln said this fixes 4-5 issues, so it would be great to get our hands on this. If there's more testing required, I also volunteer to test this out, although I'm not python-savvy enough to know how to use an SST version built from a branch (is this documented)?

walln:
For people looking to test this out:

1. You can checkout this pr locally (the github cli makes this easy - `gh pr checkout 5132`)
2. Install bun deps `bun install`
3. Sync go build `go mod tidy`
4.  Build the platform directory `cd ./platform && bun run build`
5.  Navigate to any of the python examples such as `cd ./examples/aws-python`
6. Deploy or develop with `go run ../../cmd/sst <deploy | dev >`

In my testing things work pretty well (the mentioned issue with copy files I will raise in another pr/issue as it is out of scope for this one). Please let me know about any issues, I will try to get them resolved and put some more attention on this when I can. I will also raise getting this merged soon with the core team, I am getting sick of resolving the merge conflicts myself lol.

cosmic-ascendant:
Hi @thdxr @doron98,

I wanted to gently bump this PR as it addresses several important issues in the Python runtime. I've been using walln's branch for deployment for a number of weeks now, and everything has worked as expected. 

Given that this PR:
- Fixes at least 5 reported issues dating very far back
- Has been tested by several community members
- Currently has no merge conflicts
- Has had implementation details discussed and aligned

Would it be possible to get a review when you have the bandwidth? Love to start using these improvements in our production environments.

Thank you

mpiannucci:
Another vote of confidence from me as i have been using it, especially because the main branch doesnt even work for python right now so this can only make things better (or at least it seems)

walln:
I've yet to receive any bug reports from users on this PR, and since quite a few people are already using it, it seems to be in good shape. The only remaining considerations are packaging for the SDK and any documentation feedback from @jayair.

Regarding packaging, users currently install via Git, which works fine. However, managing versions by specifying Git commit revisions can be a bit cumbersome. We had discussions about this previously, given that the sst package is already reserved on PyPI. I'm happy to revisit potential solutions with the team, though this isn't a blocker for merging—just something to keep in mind and wanted to put the offer out there.

I'd love to get this merged and am happy to assist however needed. Let me know how I can help!

alexfernandez803:
Thanks @walln for this.

I have tested this on the actual PR and it works alright on the aws-python and aws-python container projects. Thanks for the custom docker container which is an added bonus. I was able also to test adding numpy package which worked perfectly.

Not related to the PR, maybe adding a layer can be added to this?

Screenshot 2025-01-29 at 16 56 28

alexfernandez803:
@jayair  @doron98  can we merge this one please?

jayair:
Yeah Dax was going to take a look and merge.

Issue #4662, opened 2024-05-05
Cloudflare dns: After deploy Proxy is set to disabled


webdeb:
Hello, I've deployed a cluster to aws, and created the service with `public.domain.dns: sst.cloudflare.dns()`

In cloudflare the Proxy status is disabled. I can enable it manually, but when I redeploy, the proxy is disabled again, I guess it's always a new entry if some cluster configurations are changed.

<img width="220" alt="Bildschirmfoto 2024-05-05 um 14 50 26" src="https://github.com/sst/ion/assets/14992140/d5d9ffad-aeae-4b94-8cb7-d1479b7de702">

So, after deploy/redeploy the service is unavailable.

PR #5134, opened 2024-11-26
Add `proxied` to Cloudflare DNS records

### LINKED ISSUE #4261: CloudFlare DNS provider does not enable proxying
The CloudFlare dns provider does not enable proxying, and there is no way to enable proxying programmatically via the config.

What I've tried:
```ts
export const next = new sst.aws.Nextjs("Next", {
  domain: {
    name: "connectalum.com",
    dns: sst.cloudflare.dns({
      transform: {
        record: {
          proxied: true,
          ttl: 1,
        }
      }
    })
  },
  path: "apps/nextjs",
  environment: { ... }
});
```
This fails with: `failed to create DNS record: Target _39b918f2edf59bc2ff944ca1a053462c.djqtsrsxkq.acm-validations.aws. is not allowed for a proxied record. (9003))`

[Discord conversation link](https://discord.com/channels/983865673656705025/983865673656705028/1275158435566063786)

vimtor:
This PRs does two things: expose Cloudflare's `proxied` option and fixes #4261 #4662.

Previously, if you enabled `proxied` using `transform`, you would encounter these errors:

CleanShot 2024-11-26 at 10 08 39@2x

This occurred because the same `transform` was used for creating the aliases and the certificate manager records. I've added a simple check to exclude ACM records from being proxied.


fwang:
Thanks @vimtor! Renamed the field to `proxy` and refactor the code a bit.

Released in v3.3.43.

PR #5135, opened 2024-11-26
example with cloudflare, libsql (Turso) and drizzle

doichev-kostia:
This is more of a question, than a proposal.

Is this the idiomatic way of interacting with databases using SST? 

The connection parameters should be injected not only into the Cloudflare workers but also into the drizzle config


jayair:
Yeah we do this in our Drizzle quick start: https://sst.dev/docs/start/aws/drizzle

doichev-kostia:
Oh, wow, this is awesome! Thanks

PR #5145, opened 2024-11-28
chore: bump opennext version

sommeeeer:
Bump `opennext` version to the latest which is `3.2.2`. This version is especially important if your gonna use next15.


jayair:
Thanks. Should be in the next release.

PR #5152, opened 2024-11-29
docs: add hyperlink to headings

sommeeeer:
I will open this in draft mode. It will make all headings (h1-h6) in docs into clickable links. Do we want the possibility to add custom id?

Now its using the default one that Astro injects by default as you can read about here in [astro docs](https://docs.astro.build/en/guides/markdown-content/#heading-ids-and-plugins)

We could use [this package](https://www.npmjs.com/package/remark-custom-heading-id) to make it possible, it would work like this:

`## My super heading {#custom-id}`

references: https://github.com/withastro/starlight/discussions/2050


jayair:
I think it's okay if we don't have custom ids. I'll try this out. Thanks!

Issue #4021, opened 2024-10-01
Sveltekit with Lambda edge gives error even tho i dont have environment variables set.


AdityaBanik:
MyWebCdnDistribution
sdk-v2/provider2.go:509: sdk.helper_schema: creating CloudFront Distribution: operation error CloudFront: CreateDistributionWithTags, https response error StatusCode: 400, RequestID: c02658b9-bf27-4142-80eb-dcfde2894aa1, InvalidLambdaFunctionAssociation: The function cannot have environment variables. Function: arn:aws:lambda:us-east-1:913524907086:function:cyclixfrontend-production-MyWebEdgeServerFunction:4: provider=aws@6.54.1
MyWebCdnDistribution
1 error occurred:
	* creating CloudFront Distribution: operation error CloudFront: CreateDistributionWithTags, https response error StatusCode: 400, RequestID: c02658b9-bf27-4142-80eb-dcfde2894aa1, InvalidLambdaFunctionAssociation: The function cannot have environment variables. Function: arn:aws:lambda:us-east-1:913524907086:function:cyclixfrontend-production-MyWebEdgeServerFunction:4

AdityaBanik:
> can you also share the related parts of your SST config? I have lambdas with env variables that are working fine with 3.5.51

I am using lambda edge true, with normal lambda its working fine.

cwagner22:
Any update? I can't deploy my react app to Edge due to this issue.

jayair:
@cwagner22 how are you deploying your React app to the edge right now?

cwagner22:
@jayair 
```
  async run() {
    new sst.aws.React('MyWeb', {
      edge: true,
    });
  },
  ```

jayair:
Oh that component hasn't been released yet. It's still a WIP.

cwagner22:
Same for the Remix component

jayair:
<img width="558" alt="Screenshot 2025-01-03 at 7 21 10 PM" src="https://github.com/user-attachments/assets/03dd8622-119a-4045-83ae-1b0a70781456" />

We haven't released this flag yet, which is why it's undocumented and marked as internal.

PR #5159, opened 2024-12-02
chore: make react router component able to deploy

Paliago:
This is not polished enough to be GA but I have fixed the component enough that it actually deploys my RR7 app. Here's my deployed example:

https://d29euu88a6pie5.cloudfront.net/


jackblanc (approved):
This made it work for me as well, upgrading from Remix. Is there any reason this change shouldn't be merged?

Paliago:
I am guessing that Jay wants to write some proper documentation around this as well as an example. Other than that someone should really look over if my removal of the polyfill is correct. I am not really familiar with SSR functions so this is mostly a quick fix to get it working. I've also not touched the edge function for deployments on Cloudflare.

ravibanks54:
A similar approach should work for SST v2, right? I can't migrate to Ion yet since I have a few `Job`s (which aren't supported in Ion yet)

cwagner22:
What about `edge-server.mjs`? I ask because I see the same import from `@react-router/node`.
But anyway with 
```
new sst.aws.Remix('MyWeb', {
      link: [bucket],
      edge: true,
    });
```   
I get `Resource linking is not currently supported when deploying to the edge.` so I have no idea how to deploy to Lambda@Edge

Paliago:
> What about `edge-server.mjs`? I ask because I see the same import from `@react-router/node`. But anyway with
> 
> ```
> new sst.aws.Remix('MyWeb', {
>       link: [bucket],
>       edge: true,
>     });
> ```
> 
> I get `Resource linking is not currently supported when deploying to the edge.` so I have no idea how to deploy to Lambda@Edge

Did you try changing this line to end with: `from "react-router";`?
https://github.com/sst/sst/blob/cb8323f0bfe0c0807781995f0f63c8cdbc442ae3/platform/functions/react-server/edge-server.mjs#L7

And then you should use React instead of Remix:
```typescript
 new sst.aws.React('MyWeb', {
       link: [bucket],
       edge: true,
     });
```

cwagner22:
With aws.React I still get `Resource linking is not currently supported when deploying to the edge.`
Plus [this issue](https://github.com/serverless/serverless/issues/1153)
I'll do without edge for now

cwagner22:
With aws.React I still get `Resource linking is not currently supported when deploying to the edge.`
Plus [this issue](https://github.com/sst/sst/issues/4021)
I'll do without edge for now

fwang:
Fixed in v3.9.26 via https://github.com/sst/sst/pull/5289

Thanks @Paliago!

PR #5169, opened 2024-12-03
Add missing config values to sst.config.ts in aws-accounts.mdx

tobobo:
I'm setting up a new SST project and noticed that following the instructions in aws-accounts.mdx results in this error when running `sst deploy`:

```
✕  You must specify a "home" provider in the project configuration file.
```

This adds additional config values from the sst.config.ts reference and includes them in the example. I'm open to feedback on a more helpful comment to include inside the `run` function.


jayair:
Thanks!

Issue #5178, opened 2024-12-06
build.esbuild config options ignored? in cf worker


qw-in:
Seems like since the rpc change for building

https://github.com/sst/sst/blob/c67aed308434ad54338efb3a9905e59e4ece1325/pkg/runtime/worker/worker.go#L90-L128

PR #5180, opened 2024-12-06
Cloudflare: fix `WorkerArgs.build.esbuild.platform` being ignored

qw-in:
Ref: #5178

Example of fixing documented `build.esbuild.platform` option. It's the only one I need to unblock my work, but can take a pass at more if this is a direction y'all are interested in


jayair:
Is that the right issue you referenced btw?

qw-in:
> Is that the right issue you referenced btw?

Nope, my bad! 😅

qw-in:
Likely out of date

PR #5213, opened 2024-12-15
feat: allow transformation of fck-nat tags and instance

oieduardorabelo:
i'm unable to customize the tags of the nat instances created by fck-nat

i believe is missing the `transform` fn on resource creation

```ts
    const vpc = new sst.aws.Vpc("vpc", {
      az: 2,
      nat: "ec2",
      transform: {
        vpc: {
          tags: {
            Name: `${RESOURCE_PREFIX}-vpc`,
          },
        },
        natInstance: {
          tags: {
            Name: `${RESOURCE_PREFIX}-fck-nat`,
          },
        },
      },
    });
```

CleanShot 2024-12-15 at 13 25 01

### DIFF
--- a/platform/src/components/aws/vpc.ts
+++ b/platform/src/components/aws/vpc.ts
@@ -922,22 +922,25 @@ export class Vpc extends Component implements Link.Linkable {
           ([zones, publicSubnets, keyPair, bastion]) =>
             zones.map((_, i) => {
               return new ec2.Instance(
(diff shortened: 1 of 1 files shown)


oieduardorabelo:
we can close this, it was rebased on @fwang commit:

https://github.com/sst/sst/commit/7f18a174aef2e7e9255969c1f624880c95d131f9

PR #5222, opened 2024-12-17
Delete irrelevant tip about FFmpeg in lambda in aws.email

muametgrooby:


jayair:
lol how did that get in there

PR #5224, opened 2024-12-17
[AWS Bus]: add lambda context to bus subscriber handler

### LINKED ISSUE #5223: [AWS Bus]: Add lambda `context` to subscriber callback
https://github.com/sst/sst/blob/ab2c709543982938fb37ffea05d95e56f0506ed4/sdk/js/src/aws/bus.ts#L14C2-L31C4

Can the lambda context be added as a 3rd param to the subscriber callback? I would like to use lambda power tools to make part of the subscriber idempotent and do not want to wrap the entire handler. 

I propose the following. And I will get a PR up as well shortly.
```ts
export function subscriber<Events extends event.Definition>(
  _events: Events | Events[],
  cb: (
    input: {
      [K in Events['type']]: Extract<Events, { type: K }>['$payload']
    }[Events['type']],
    raw: EventBridgeEvent<string, any>,
    context: Context,
  ) => Promise<void>,
): EventBridgeHandler<string, any, void> {
  return async function (event, context) {
    const payload = {
      type: event['detail-type'],
      properties: event.detail.properties,
      metadata: event.detail.metadata,
    }
    return cb(payload, event, context)
  }
}
```

natac13:
closes #5223

### DIFF
--- a/sdk/js/src/aws/bus.ts
+++ b/sdk/js/src/aws/bus.ts
@@ -1,7 +1,7 @@
 import { AwsOptions, client } from "../aws/client.js";
 import { Resource } from "../resource.js";
 import { event } from "../event/index.js";
-import { EventBridgeEvent, EventBridgeHandler } from "aws-lambda";
+import { EventBridgeEvent, EventBridgeHandler, type Context } from "aws-lambda";
 
 export module bus {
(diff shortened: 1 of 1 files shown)


dragonautdev:
@thdxr any chance this PR can be merged? It would be super handy to get lambda function context in the handler

vimtor:
duplicate of #5872

PR #5230, opened 2024-12-19
[AWS Bus]: include metadata in published event

### LINKED ISSUE #5225: [AWS Bus]: `metadata` from `publish` not getting passed down
When I have the following, the metadata does not make it to the event

```ts
import { z } from 'zod'
import { defineEvent } from '../event'
import { getErrorMessage } from '../utils/misc'
import { bus } from 'sst/aws/bus'
import { Resource } from 'sst'
import crypto from 'node:crypto'

const OrganizationCreatedEventSchema = z.object({
  organizationId: z.string(),
})
type OrganizationCreatedEvent = z.infer<typeof OrganizationCreatedEventSchema>

export const OrganizationEvents = {
  OrgCreated: defineEvent('org.created', OrganizationCreatedEventSchema),
}

export interface OrganizationEventsDispatcher {
  publishOrgCreatedEvent(payload: OrganizationCreatedEvent): Promise<void>
}

export function createOrganizationEventsDispatcher(): OrganizationEventsDispatcher {
  return {
    async publishOrgCreatedEvent(payload: OrganizationCreatedEvent) {
      try {
        await bus.publish(
          Resource.Bus,
          OrganizationEvents.OrgCreated,
          payload,
          {
            metadata: {
              id: crypto.randomUUID(),
              timestamp: new Date().toISOString(),
            },
          },
        )
      } catch (error) {
        const errMsg = getErrorMessage(error)
        console.error(
          `[OrganizationEventsDispatcher] Error publishing organization created event: ${errMsg}`,
        )
        throw new Error('Failed to publish organization created event')
      }
    },
  }
}
```

```ts
// subscriber
export const handler = bus.subscriber(
  [OrganizationEvents.OrgCreated],
  async (event) => {
    console.log(event.type, event.properties, event.metadata) // event.metadata is empty
    switch (event.type) {
      case 'org.created': {
        // ...... other code
        break

      }
    }
  },
)


```

I believe the fix is below

https://github.com/sst/sst/blob/c2657a114ad3498721bea7f8c9d9761c0966ddd6/sdk/js/src/event/index.ts#L45-L57
```diff
      async function create(properties: any, metadata?: any) {
        metadata = input.metadata
          ? typeof input.metadata === "function"
            ? input.metadata(type, properties)
            : input.metadata(metadata)
+        : metadata
-         : {};
      }
```

Only thing I am unsure of is the `input.metadata` which I see comes from the `defineEvent` call, and how it fits into all this. 
However, I did some testing locally and the above change would get the metadata from the `publish` method call to the lambda event subscriber.

If this is good let me know and I can make a quick PR. Thanks again

natac13:
closes #5225

### DIFF
--- a/sdk/js/src/event/index.ts
+++ b/sdk/js/src/event/index.ts
@@ -47,7 +47,7 @@ export module event {
           ? typeof input.metadata === "function"
             ? input.metadata(type, properties)
             : input.metadata(metadata)
-          : {};
+          : metadata;
         properties = validate(properties);
         return {
           type,
--- a/sdk/js/src/event/index.ts
+++ b/sdk/js/src/event/index.ts
@@ -47,7 +47,7 @@ export module event {
           ? typeof input.metadata === "function"
             ? input.metadata(type, properties)
             : input.metadata(metadata)
-          : metadata;
+          : metadata ?? {};
         properties = validate(properties);
         return {
           type,


sdk/js/src/event/index.ts
```diff
@@ -47,7 +47,7 @@ export module event {
           ? typeof input.metadata === "function"
             ? input.metadata(type, properties)
             : input.metadata(metadata)
-          : {};
+          : metadata;
```
vimtor: minor improvement:

```suggestion
          : metadata ?? {};
```

vimtor (changes requested):
thanks for the contribution @natac13 

when trying to reproduc the issue i've noticed that this only happens if you don't provide `metadata` to the `event.builder`

either way, your change works! i've requested a small change to be redundant

here is proof of it working:

```ts
const defineEvent = event.builder({
  validator: ZodValidator,
});

export const MyEvent = defineEvent(
  "app.myevent",
  z.object({
    foo: z.string(),
  }),
);

export async function handler() {
  await bus.publish(
    Resource.Bus,
    MyEvent,
    { foo: "hello" },
    {
      metadata: {
        example: 123,
      },
    },
  );
  return {
    statusCode: 200,
  };
}
```

Image

PR #5232, opened 2024-12-19
fix: add a check for spaces in component name

### LINKED ISSUE #5231: `Dev` crashes without proper error message with spaces in component name
## Description

Due to a typo, I came across the problem that sst does not handle spaces in every component names. This led to strange behaviour when using the `dev` command of sst. It would result in a successful deployment, but the pane with the misnamed component would shut down with an unexpected error. Trying to add logs didn't work. So in `dev` mode it was hard to pinpoint the source of the error. If you try to deploy it, it throws an error when creating `cachePolicies` (this was tested with the `Nextjs` construct). I have tested it with the `examples/aws-nextjs` example and for that bucket the problem does not occur (probably due to internal renaming).

## Reproduce

SST Version: `3.3.47`
OS: `macOS 15.1.1`
Chip: `Apple M2 Pro`

Just create a simple construct and add a space anywhere in the component name. 

image

Then run `npx sst dev` and the deploy will be successfully but the component pane will have an error which exited the process

image

image

When I run the `npx sst deploy` command I get this error indicating a misformed name for the cache policy. But this does not happen for every construct (the bucket for example).

image

## Solution proposal

I am not sure if you want to allow spaces in component names. I think not allowing spaces and throwing an proper error message if spaces are present in components would be a good solution. Therefore a simple check in `component.ts` should be enough to check the name for spaces and throw an error with a descriptive message if this is the case. ~I will make a PR suggestion soon.~ I opened a PR for that #5232

RiFrost:
This closes #5231.
This would throw an error if a component name contains a space. The message would include the incorrect name and point to the construct causing the error.

Screenshot 2024-12-19 at 3 21 20 PM


jayair:
Oh nice, let me get this reviewed.

fwang:
Released in v3.4.58

Thanks @RiFrost!

PR #5233, opened 2024-12-19
docs: example lambda retry with queues

natac13:
I thought it would be nice to have a retry example which uses similar logic to how sst v2 did retries for EventBus subscribers.

I have seen it asked a few times on discord and just thought it would be a nice challenge for me to try. I was able to get things working so I figured an PR to add an example might be helpful to others. 

Let me know if you want me to change anything.


jayair:
Awesome. Thanks the comment block, it'll get added to the docs as well.

PR #5239, opened 2024-12-20
add(sst init nextjs): disable eslint on reference line in sst.config.ts

sommeeeer:
When you create a new nextjs application with `npx create-next-app` it will by default add `eslint.config.mjs` to the project with the rule [triple-slash-reference](https://typescript-eslint.io/rules/triple-slash-reference/) turned on as error. 

This PR will check if the template == `nextjs` and if the directory contains a file `*eslint*`. If so it will add a disable eslint above the top reference line.

### DIFF
--- a/cmd/sst/init.go
+++ b/cmd/sst/init.go
@@ -228,6 +228,26 @@ func CmdInit(cli *cli.Cli) error {
 		spin.Stop()
 	}
 
+	if template == "nextjs" {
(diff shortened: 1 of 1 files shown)


jayair:
Nice! I'll test it out soon.

jayair:
Just tested it, looks good. Thanks!

PR #5242, opened 2024-12-20
feat: add `sst state list` command to list all deployed stages

jonlambert:
## Summary

This PR adds a command to the SST CLI which will list all deployed stages of an application.

```
sst state list
```

(* Command name updated per comment from @thdxr!)

## Approach

I would really appreciate any critique of the approach I've taken here – the main motivation for opening this PR was as an exercise to understand the SST codebase a bit better.

The stages are fetched directly from the state store for each provider. I've implemented and tested this for AWS, Cloudflare and local providers. 

I was unsure regarding the naming of this command too. Other potential options could bring a bit more clarity:

```
sst list stages
```

...or align with the existing secrets approach: 

```
sst stage list
```

...or go maximum brevity:

```
sst list
```

```
sst ls
```

~~

Happy to change to whatever the team feels is most appropriate providing the idea and approach is deemed sound.

Thank you in advance for the consideration!

## Example Output

(The account IDs have been changed for privacy!)

### AWS

```bash
Provider:   AWS
Region:     eu-west-2
Account:    492000000000
Stages:     dev
            jon
```

### Cloudflare

```bash
Provider:   Cloudflare
Account:    aDcX2U0d9xd4f130ea16fb723d14a00v
Stages:     dev
            jon
```

### Local

```bash
Provider:   Local
Path:       /Users/jon/Library/Application Support/sst
Stages:     dev
            jon
```

### `--simple`

The output should be the same for all providers when running `sst stages --simple`:

```bash
dev
jon
```


jayair:
Thanks for looking into the internals.

Does this pick up stages from the AWS account of the current credentials? Or does it try to fetch across accounts?

jonlambert:
> Does this pick up stages from the AWS account of the current credentials? Or does it try to fetch across accounts?

Yes, the former! It doesn't try to do anything different in terms of credentials. It'll behave the same as `sst secret list` for example.

jayair:
Hmm yeah the stages are a bit tricky because it might give you the impression that say prod has gone missing. But it might be because of the credentials.

We could overcome that by displaying the AWS account id that it's pulling the stages from. Let me give it some thought.

jonlambert:
Good point, @jayair. You're right that could be misleading. I did wonder about adding more information in the output, but I couldn't spot anything common that existed already between all the providers. 

I hope you don't mind, but I've taken a stab at collecting and outputting this information in addition to the list of stages. I've added this in [a single commit](https://github.com/sst/sst/pull/5242/commits/26eecf0c9893040023578ccf71a50f4cce5a738e) in case it's helpful to review in isolation. It could be split out or removed if you had another approach in mind. 🙂

I've added examples of the output for each provider to the PR description, too.

I did think this could be refactored in future to output the info in a different format via flags (JSON, etc) but didn't want to complicate anything further at this stage 😄 I'm not precious about the approach but I'd love any feedback if you have time.

(I'd also like to take this opportunity to thank you for creating SST! It's a real joy to use and I've not been this excited by a project in a long while 😄)

pkg/project/provider/local.go
```diff
@@ -105,3 +106,10 @@ func (a *LocalHome) listStages(app string) ([]string, error) {
 
 	return stages, nil
 }
+
+func (c *LocalHome) info() (util.KeyValuePairs[string], error) {
+	return util.KeyValuePairs[string]{
+		{Key: "Provider", Value: "Local"},
+		{Key: "Path", Value: global.ConfigDir()},
```
jonlambert, line 113: Is changing the local home path common enough to be useful here?

jayair:
Appreciate the effort. Let me ask @thdxr to take a look.

thdxr:
thanks for doing this! some ideas

- i think this should go under `sst state list` - that's where we plan on putting all state related ops
- should the added function be specifically `listStages` or something more generic like `list(prefix)` and that way it can be used to enumerate any path?

jonlambert:
Thanks for the comments @thdxr! 

> i think this should go under `sst state list` - that's where we plan on putting all state related ops

Agreed - that makes a lot of sense. I've changed that to `sst state list` 🙂

> should the added function be specifically listStages or something more generic like list(prefix) and that way it can be used to enumerate any path?

Hmm yeah I did wonder if this function could be a bit more generic (it definitely feels pretty specific) but I'm not too sure what other things would fit into the pattern: `/app/{app}/{stage}.json`. Most other enumerable things seem to live under an app and stage-specific directory. Would `app` become the prefix in this scenario?

jonlambert:
@thdxr I've rebased this branch and moved the `list` command inline with the other new state commands 🙂

jonlambert:
Hey @jayair + @thdxr - other than resolving conflicts, is there anything else I can do here to help progress this? It would be really useful for us to be able to list all deployed stages.

jayair:
I'll bump this

thdxr:
i'll merge once conflicts are fixed

jonlambert:
Thanks @jayair + @thdxr - conflicts are resolved!

Issue #5241, opened 2024-12-20
Bug: Router worker for `sst.cloudflare.StaticSite` fails with Error 1101; ESBuild `define` option not applied correctly in `sst.cloudflare.Worker`


john-royal:
If I deploy a static site using the `sst.cloudflare.StaticSite` component, all requests to the website result in a Cloudflare 1101 error. The router worker is throwing an error because the `SST_ASSET_MANIFEST` is undefined.

This appears to be because the ESBuild `define` option, used to provide the asset manifest to the router worker, is not actually applied in the build step for the `sst.cloudflare.Worker` component. I've identified a one-line fix for this, which I'll include in a PR.

sam-goodwin:
Also running into this. Is there a workaround while we wait for the PR to be merged?

john-royal:
Not the answer you're looking for, but my workaround has been to just use `sst.aws.StaticSite` instead. Since the Cloudflare StaticSite component isn't documented on the website (and the JSDoc comments in the code seem to indicate it's still a WIP), I don't think this is a priority for them at the moment.

However, it does seem like everything is working except for this one thing, so I do hope my 1-line PR gets merged!

If you're desperate, it might not be too hard to just copy the code for the component and the static site router worker, but then inject the asset manifest using environment variables instead of using `define` in ESBuild. Definitely not ideal, but doable.

sam-goodwin:
Thanks @john-royal, I also bailed on CloudFlare and went back to AWS. Agree that CloudFlare does not appear to be the priority. I was hoping to take CF for a spin, but given this bug and DurableObjects not being available, I think i'll have to wait a bit longer.

thdxr:
going to close this for now because we are still waiting on cloudflare updates before we dive back into cloudflare support

PR #5243, opened 2024-12-20
fix(sst.cloudflare.Worker): include "define" property in esbuild options

john-royal:
When building a Cloudflare Worker using the `sst.cloudflare.Worker` component, it appears that the `define` option in ESBuild is not being passed along correctly. This also causes the `sst.cloudflare.StaticSite` component to fail in production — see #5241 for details.

This one-line change seems to fix it — I've tested the Worker component with and without.

Edit: Confirmed that this fixes the StaticSite component as well!


sam-goodwin:
What's the hold up on merging this? Seems like a sensible fix and not being able to deploy a static site to CloudFlare is a painful blocker. @jayair @thdxr

PR #5245, opened 2024-12-21
[Example]: add go lambda with dynamo usage example.

natac13:
Adding a example of using a go lambda with dynamodb.

I tried to keep things as minimal as possible to get the basic idea across of how to connect to dynamo.

Let me know if you would like anything changed.


natac13:
Sorry I forgot comments in `sst.config.ts` Ill add them shortly I get the high level setup on the website docs.

jayair:
Thanks!

PR #5251, opened 2024-12-23
[Example]: Go APIGatewayV2 with Standard Library Http server

natac13:
add go api gateway v2 example with standard library http server


jayair:
Thanks!

PR #5256, opened 2024-12-25
Windows support

abeisleem:
Took a stab at this over the weekend—there’s still work to do, but I wanted to create this PR for visibility and confirm it’s aligned with expectations before proceeding (@jayair / @thdxr). I’ve developed and tested this on Windows and spun a Linux instance to confirm changes are not breaking (via a Docker container, which I haven’t committed but may be useful for future testing).

I got `go build ./cmd/sst && ./sst(.exe) dev` & `go run ../../cmd/sst dev` to work without errors against cloudflare-hono & aws-api examples up until it asks for credentials (used creds for Cloudflare but not AWS). Pending is further `dev` testing (tunnel, etc), supporting other commands and build/release pipeline for Windows.

Where do we go from here? 
1. You think this is hot trash and/or prefer to implement it in-house → I remove this PR from existence and we never talk about it
2. You think this is a good starting point but prefer to push it to the finish line in-house → I'll add more details to the PR, answer any questions, supports however I can
3. You're good with this, the idea of others contributing to this feature, and give me the green light to keep going → I'll add more details to answer open questions about the PR and keep contributing as per your feedback


thdxr:
wow this is great thanks for doing this

need to spend some time looking over it but on first glance looks great - maybe will land on 2 or 3

pinoniq:
Is there a way others can help on this? Would be nice to get this PR merged (as beta or whatever) and then work on porting over feature per feature in seperate issues to lighten the load?

PR #5257, opened 2024-12-25
Fix: copyFiles in Function used $dev instead of dev

doron98:
You accidentally used the $dev variable instead of dev variable in this commit 2caabf2 to prevent copy files in dev


jayair:
Oh we'll take a look.

fwang:
Nice catch. Thanks @doron98!

fwang:
Fixed in v3.4.36

PR #5273, opened 2025-01-01
Added index.handler path

robertseghedi:
In the provided example, it's about the src/index.handler, but in the primary definition, it's only index.handler, which makes the build fail.


jayair:
Good catch!

PR #5274, opened 2025-01-01
Aurora: Enable Data API via `arg` & expose arns for cluster and secret

### LINKED ISSUE #5275: Aurora: Expose secret arn and cluster arn
Would it be good to still expose the cluster arn and secret arn so that linked services can use the DataAPI connection.

I have a PR to do this I believe #5274

natac13:
useful for DataAPI usage and when other services need to connect to RDS

closes #5275


fwang:
Thanks @natac13!

PR #5276, opened 2025-01-01
fix: ensure pause after is only set with min ACU 0

zammitjames:
Fixes the following error:
```
  InvalidParameterValue: SecondsUntilAutoPause can only be specified when minimum capacity is 0.: provider=aws@6.66.2
```

The updated code checks whether the minimum capacity is 0.
- If `minimum_capacity === 0` then it uses the given `pauseAfter` duration or defaults to 5 minutes. 
- If `minimum_capacity > 0`, then `pauseAfter` isn't set.


fwang:
Thanks @zammitjames!

PR #5281, opened 2025-01-02
Add event/validator to export map

mwood23:
After upgrading to the latest a script of mine started returning:

```sh
node:internal/modules/esm/resolve:303
  return new ERR_PACKAGE_PATH_NOT_EXPORTED(
         ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './event/validator' is not defined by "exports" in /Users/marcuswood/moonshots/foo/packages/core/node_modules/sst/package.json
    at exportsNotFound (node:internal/modules/esm/resolve:303:10)
    at packageExportsResolve (node:internal/modules/esm/resolve:645:13)
    at resolveExports (node:internal/modules/cjs/loader:591:36)
    at Module._findPath (node:internal/modules/cjs/loader:668:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:1130:27)
```

I'm guessing it was this commit https://github.com/sst/sst/commit/6353b5886e2d25d62a57b10722905e67db48cc16?

How it's used in my code:
```ts
import { event } from 'sst/event';
import { ZodValidator } from 'sst/event/validator';
import { ActorContext } from './actor';

export const defineEvent = event.builder({
  validator: ZodValidator,
  metadata() {
    return {
      actor: ActorContext.use(),
    };
  },
});

```

```ts
import { bus } from 'sst/aws/bus';

export const handler = bus.subscriber(
  [Workspace.Event.Created],
  async (event) => {
// ...
```

### DIFF
--- a/pkg/project/provider/provider.go
+++ b/pkg/project/provider/provider.go
@@ -192,7 +192,7 @@ func PutSecrets(backend Home, app, stage string, data map[string]string) error {
 func PushPartialState(backend Home, updateID, app, stage string, data []byte) error {
 	slog.Info("pushing partial state", "updateID", updateID)
 	err := json.Unmarshal(data, &map[string]interface{}{})
-	if err != nil {
+	if err != nil || len(data) == 0 {
 		return fmt.Errorf("something has corrupted the state file - refusing to upload: %w", err)
 	}
 	return backend.putData("app", app, stage, bytes.NewReader(data))


thdxr:
hm we added a glob pattern at the bottom - it's not resolving that? we were actually planning on getting rid of all exports except that

PR #5286, opened 2025-01-03
[Cloudflare/KV] Add ability to reference existing Cloudflare Workers KV

Max101:
Add ability to reference an existing KV storage when creating it via SST


Max101:
@thdxr is there any chance this might be merged in ?

Max101:
@thdxr the same can be achieved with this, so maybe instead of this PR we can just add the below to the docs

```
const cloudflareKv = new sst.Linkable('cloudflareKv', {
      properties: {},
      include: [
        sst.cloudflare.binding({
          type: 'kvNamespaceBindings',
          properties: { namespaceId: `your-existing-namespace-id` },
        }),
      ],
    });

// ...
// link: [cloudflareKv]
```

vimtor (changes requested):
thanks for the contribution @Max101 

most components implement a static `get` method for this use case

for example the `Postgres` component:
https://sst.dev/docs/component/aws/postgres/#static-get

if you add one like this to the `cloudflare.Kv` component we can merge this PR

vimtor (approved):
<details>

Image

Image

</details>

PR #5289, opened 2025-01-03
React Router v7 Support

jackblanc:
## Description

This PR adds support for React Router v7 as an SSR site. It fixes a few critical issues in the current React component:
- Change imports from `@react-router/node` to `react-router`, as these exports are removed in v7+ [Changelog: Removed Adapter Re-exports](https://reactrouter.com/start/changelog#removed-adapter-re-exports)
- Removed `polyfill.mjs`, as React Router no longer exports the `installGlobals` polyfill, and instead requires >= Node 20 now [Changelog: Minimum Versions](https://reactrouter.com/start/changelog#minimum-versions)
- Remove `index.html` as `defaultRootObject`, because the server build does not produce an index.html file

## Testing 

Initially I added an example `aws-react` to this PR, but decided it should be added separately. Happy to add it back if desired. 

You can test in a React Router app with the following steps: 
- Create a starter react router app: `bunx create-react-router@latest aws-react`
- Install SST, add the `sst.aws.React` component to `sst.config.ts`
- Copy the files from this PR to the `.sst` folder in the repo (run `bun sst install` before copying, otherwise they'll be overwritten)

In addition, I'm running this code in production right now in my own app. 

## Credits & Future work

This PR builds on top of https://github.com/sst/sst/pull/5159.

I'm happy to make more changes if desired, but the current version doesn't work at all so I don't see an issue with merging and building from there.


andrewbellucci:
This looks great! What're your thoughts on changing the name of the component to ReactRouter instead of React? @jackblanc

jackblanc:
@andrewbellucci personally I think that makes more sense, I was initially confused what the "React" component was, only to learn it's just ReactRouter. I'll leave it up to @jayair @thdxr to decide if it should be renamed?

ironheart122:
Wow. Why is this PR not merged? :/

dreamorosi:
I might be wrong, but when a loader throws a redirect the response has no body, i.e. 

```
{
  status: 302,
  statusText: '',
  headers: Headers { Location: '/login' },
  body: null,
  bodyUsed: false,
  ok: false,
  redirected: false,
  type: 'default',
  url: ''
}
```

With the current implementation in `platform/functions/react-server/regional-server.mjs` (L118-126) the server hangs:

```ts
if (response.body) {
  const reader = response.body;
  const writer = awslambda.HttpResponseStream.from(
    responseStream,
    httpResponseMetadata,
  );
  await streamToNodeStream(reader.getReader(), responseStream);
  writer.end();
}
```

I had to modify the function similar to the one under `platform/functions/remix-server/regional-server.mjs` to account for this:

```ts
const writer = awslambda.HttpResponseStream.from(
  responseStream,
  httpResponseMetadata,
);
if (response.body) {
  await streamToNodeStream(response.body.getReader(), responseStream);
} else {
  writer.write(" ");
}
writer.end();
```

or it wouldn't work.

jackblanc:
You are correct @dreamorosi, I figured that out the hard way yesterday, sadly didn't think to check remix-server lol

Pushing that change up now

platform/src/components/aws/react.ts
```diff
@@ -141,10 +135,10 @@ export interface ReactArgs extends SsrSiteArgs {
    */
   invalidation?: SsrSiteArgs["invalidation"];
   /**
-   * Set [environment variables](https://remix.run/docs/en/main/guides/envvars) in your React app. These are made available:
+   * Set environment variables in your React app. These are made available:
```
jackblanc: There's no equivalent docs link here

Murderlon:
Not sure if this PR is the right place for it, but I would love it if errors in logs actually pointed to a file and line instead of some `bundle.mjs` gibberish. In Remix, this was done in `remix-serve` (and presumably the same in React Router). They have docs how to[ enable source maps if you don't use it](https://remix.run/docs/en/main/start/v2#source-map-support). I'm wondering if that's something that can be done here? It would make dealing with errors significantly more meaningful.

fwang:
Thanks @jackblanc!

Two things i'm working on:

- [x] 1. Use dynamic origins.
  **Context**:
    - CloudFront only allows for 25 behaviors by default. Currently the `React` component creates 1 behavior for the server function, and 1 behavior per top level asset file/folders. And the deploy fails if there are more than 24 top level asset file/folders.
    - Recently CloudFront announced [dynamic origins](https://aws.amazon.com/about-aws/whats-new/2024/11/amazon-cloudfront-origin-modifications-cloudfront-functions/). We are no longer bound by the 25 behavior limit.
- [x] 2. Rework edge mode.
  **Context**:
    - Currently the [deploying to edge](https://github.com/sst/v2/tree/master/packages/astro-sst#deployment-strategies) use Lambda@Edge. L@E has limitations, slow to deploy, cannot be removed (has to wait 20min).
    - SST v3 can deploy to multiple regions cleanly & natively. I want to move away from L@E, and have the component deploy to multiple regions. ie. `regions: ["us-east-1", "eu-west-1"]`, or `regions: "all"`.


I wanted to get them out before this RR7 change. I'll try to get them out next week.

It should be an easy merge into this PR.

churbis:
Does anyone have an sst config setup to build React-Router v7 Cloudflare workers separately before deploying as an SST cloudflare worker component as a stop-gap for getting React-Router v7 supported in SST? I am trying to use Wrangler as per their Cloudflare templates to generate an index.js and then deploy with SST but cannot figure out how to get it to package assets correctly.

platform/functions/react-server/regional-server.mjs
```diff
@@ -101,7 +35,7 @@ function convertApigRequestToNode(event) {
 }
 
 const createApigHandler = (build) => {
-  const requestHandler = createNodeRequestHandler(build, process.env.NODE_ENV);
+  const requestHandler = createRequestHandler(build, "production");
```
Nimbo1999: I'm wondering why have you removed this environment variable. Isn't it necessary at all?

ch-stuart:
@fwang any update on this PR or your dependent changes you mentioned above? We made the unfortunate mistake of updating to RR7 from Remix without first checking if SST had support for RR7 and I'm just looking to sort out the best way to proceed. Many thanks.

ironheart122:
> @fwang any update on this PR or your dependent changes you mentioned above? We made the unfortunate mistake of updating to RR7 from Remix without first checking if SST had support for RR7 and I'm just looking to sort out the best way to proceed. Many thanks.

Why not use this fork in the meantime? tbh, I would rather drink sour milk made than have to undo all the RR7 migration changes.

ironheart122:
https://github.com/ironheart122/react-router-7-x-sst-template

ch-stuart:
> https://github.com/ironheart122/react-router-7-x-sst-template

This worked for me, thanks!

platform/functions/react-server/regional-server.mjs
jackblanc: I think any time you are building via SST you would always want the server to operate in "production" mode, but I'm not super familiar with the nuances of each mode

Kylar13:
Just ran across this while trying to move my cloudflare rr7 app to using SST.

Any ideas on how to move this over to cloudflare?
The listed workarounds use an `aws.React` component that is not available there, but the `cloudflare.Remix` component uses imports from `@remix-run/cloudflare` which is missing from my project since I'm on RR7 as a framework.

barakcodes:
> Just ran across this while trying to move my cloudflare rr7 app to using SST.
> 
> Any ideas on how to move this over to cloudflare? The listed workarounds use an `aws.React` component that is not available there, but the `cloudflare.Remix` component uses imports from `@remix-run/cloudflare` which is missing from my project since I'm on RR7 as a framework.

I've been waiting for a while too, for now it seems better to stick to cloudflare, SST still has a lot of stuff they need to support for cloudflare to be a viable option.

fwang:
Released in v3.9.26

@jackblanc thanks for putting this together!!

Apart from this pr, v3.9.26 also has 2 improvements over the old `Remix` component:
- Dynamic routing at edge
- Reworked edge mode via the [`regions` prop](https://sst.dev/docs/component/aws/react#regions)

Details about these two improvements mentioned above https://github.com/sst/sst/pull/5289#issuecomment-2588226633

richardscarrott:
> @andrewbellucci personally I think that makes more sense, I was initially confused what the "React" component was, only to learn it's just ReactRouter. I'll leave it up to @jayair @thdxr to decide if it should be renamed?

Totally should be renamed — super confusing. Aside, great work here!

PR #5295, opened 2025-01-04
[docs] Include Cloudflare API instructions and subdomain mention

sergical:
* Cloudflare dashboard link to make it easier for people to find their api key
* Was no mention of subdomains here so was worried it wouldn't work until I tried it


jayair:
Thanks!

PR #5339, opened 2025-01-13
feat: prepend user banner to SST banner in worker runtime

siddhatiwari:
When using the worker runtime, any user-provided ESBuild are not being applied. This prevents users from adding their own banner code that needs to run before SST's require setup.

This change modifies the worker runtime to:
- Preserve user-provided banner options
- Prepend them before SST's default banner
- Maintain backward compatibility when no user banner is provided

Note: it's a simple change but currently untested


siddhatiwari:
Just tested with a custom banner, all working as expected

jayair:
Thanks for taking a crack at it.

user-0a:
+1, just wanted to nudge on this

PR #5374, opened 2025-01-22
docs(apigateway): Mention that VPC AZ's need to support VPC or deploy will FAIL

DriesCruyskens:
I lost quite a bit of time debugging the following error trying to use ApiGateway's `routePrivate()`:

```shell
creating API Gateway v2 VPC Link (nextjs-app-production-TestApiVpcLink): operation error ApiGatewayV2:
CreateVpcLink, https response error StatusCode: 400, RequestID: bccc43a2-256c-4274-80f8-1c925c2f06c1,
BadRequestException: Subnet is in Availability Zone 'euw3-az2' where service is not available: provider=aws@6.66.2
```

I learned that
- VPC link is not supported in all AZ's: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vpc-links.html
- You need to map supported AZ ID's to your AWS Account's AZ's using `aws ec2 describe-availability-zones`. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#az-ids

I wanted to at least mention this in the docs so SST users that come after me can be set on the right track. I made some proposed changes to the docs and the example. I wanted to be as thorough as can be but feel free to rewrite or eliminate of course.

### DIFF
--- a/examples/aws-cluster-vpclink/sst.config.ts
+++ b/examples/aws-cluster-vpclink/sst.config.ts
@@ -29,7 +29,14 @@ export default $config({
     };
   },
   async run() {
-    const vpc = new sst.aws.Vpc("MyVpc");
+    const vpc = new sst.aws.Vpc("MyVpc", {
+      // VPC links (created by `api.routePrivate()`) are not supported in all availability zones.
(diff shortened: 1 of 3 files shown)


jayair:
Appreciate it, I'll take a look.

PR #5378, opened 2025-01-23
fix(config): rename old `protected` field to `protect`

### LINKED ISSUE #5314: False type definitions in sst.config for `protect` property
This might be an issue on my side but when trying to add the new `protect` option to my `sst.config.ts` I only got autocompletion for the `protected` property:
image

This seems to be a naming mistake as the actual property is called `protect` and `protected` just does nothing.
Again this might be some weird issue on my side but I thought Id report it just in case.

unrevised6419:
Closes #5314


fwang:
Thanks @iamandrewluca!

PR #5379, opened 2025-01-23
Update iam credentials Docs

j-dexx:
Live now uses appsync, so needs the appsync permissions rather than the IOT ones


j-dexx:
Our app uses a specific lambda role we attach permissions to this works fine for me.

jayair:
Appreciate it!

PR #5381, opened 2025-01-24
Update Tanstack Start Component + Examples

mckamyk:
Prior to `v1.96.x`, (alpha) server functions were called
- `http://localhost:3000/_server?serverFnId=....&payload=....` - single route, everything in query params

Now with `v1.96.x` and later (beta) server function are called as
- `http://localhost:3000/_server/app_path_to_file--funcitonName?payload=...` - route per server function, payload in params

This is a problem because the `sst.aws.TanstackStart(...)` component was only building cloudfront rules for `/_server/` and not `/_server/*` pointing to lambda. *All server functions are broken after v1.96.x` This PR remedies that.

This PR also
- accommodates vinxi bug creating empty folders inside `.output/public/` for server-only routes, which caused SST to create static routes to the S3 bucket erroneously.
- updates `examples/aws-tanstack-start/` to latest version
- adds `examples/aws-tanstack-start-alpha/` which is pinned to `v1.95.x`, before the route change.

Deployed examples
- [alpha](https://d8vvn83ygpcps.cloudfront.net)
- [beta](https://d3jy97ub3kg60q.cloudfront.net/)

Mouseover any of the links with devtools open to see the server functions fly


platform/src/components/aws/tanstack-start.ts
```diff
@@ -485,13 +497,15 @@ export class TanstackStart extends Component implements Link.Linkable {
...
+              // These seem duplicate?
               {
                 cacheType: "server",
                 cfFunction: "serverCfFunction",
                 origin: "server",
               },
+              // I'm not even sure this is necessary. Default should send it to lambda.
               {
-                pattern: "_server/",
+                pattern: "_server/*",
```
mckamyk, line 508: Primary change that fixes the new ServerFunctions, although Vinxi leaves junk empty `.output/public/_server/` and `.output/public/api/` for some unknown reason, so...

platform/src/components/aws/tanstack-start.ts
```diff
@@ -444,6 +444,18 @@ export class TanstackStart extends Component implements Link.Linkable {
             .readdirSync(path.join(outputPath, assetsPath), {
               withFileTypes: true,
             })
+            .filter((item) => {
```
mckamyk, line 447: This simply filters out empty folders, which prevents errors on cloudfront behavior creation with duplicated path for `/_server/*` and preventing `/api/*` from pointing to an S3 bucket instead of the server lambda.

PR #5426, opened 2025-02-05
DRAFT: feat: extended api gateway v1 to have configurable cors headers

Pascal-Lohscheidt:
# Why
I would like to specify the headers returned by my gateway, this makes setting up a gateway much easier.

# What
I changed the `cors`  property to be a config object and added the "responseHeaders" field.

# Note
I know at the moment, this would be a breaking change for the API of that component, but this is rather a concept PR.
Is this according to how you want to design your API or should this be solved differently.

I thought before I put more work into this topic, I ask about the plans on how something like that should be done in SST.

(First time contributing here I hope I did this correctly)

### DIFF
--- a/platform/src/components/aws/apigatewayv1.ts
+++ b/platform/src/components/aws/apigatewayv1.ts
@@ -262,7 +262,20 @@ export interface ApiGatewayV1Args {
    * }
    * ```
    */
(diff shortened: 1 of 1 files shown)


jayair:
Thanks. Can you show me a before and after snippet of the API change?

PR #5430, opened 2025-02-05
Task: handle no dev command set in dev mode

### LINKED ISSUE #5383: Error when using Tasks: panic: runtime error: slice bounds out of range [1:0]
I get this error when running task in API handlers

Image

sst.config.ts
```ts
export const vpc = new sst.aws.Vpc("Vpc");

export const cluster = new sst.aws.Cluster("Cluster", { vpc });

// https://github.com/sst/sst/blob/dev/examples/aws-task/image/Dockerfile
export const task = cluster.addTask("SyncTask", {
  link: [...allSecrets],
});

```

Dockerfile
```ts



FROM node:20
    
WORKDIR /app/

COPY package.json /app
RUN npm install

RUN npm run build

ENTRYPOINT [ "node", "dist/index.js" ]
```

fwang:


fwang (commented):
sync

PR #5440, opened 2025-02-07
SNS topic reference method

OldSneerJaw:
Adds a static `get` method to the `SnsTopic` class so that, like other lower-level SST components, one can retrieve a reference to an SNS Topic from outside the app/stage. Complements the existing static `subscribe` and `subscribeQueue` methods that allow one to subscribe to an outside SNS Topic without having a reference to it. Inspired by the static `get` method in the `Bucket` component.


fwang:
Will be in the next release.

Thanks @OldSneerJaw!

PR #5460, opened 2025-02-13
fix: remix server lambda occasionally hangs

JannikWempe:
We had an issue that the Remix server Lambda occasionally hangs and runs into a timeout. Usually the p95 request time is <1s but sometimes it runs into a 30s timeout. It seemed like that is more often the case if there is more traffic.

Also, the requests seem to be quite slow. 

I have switched the implementation to `pipeline` which also handles things like backpressure and cleanup (and is more readable). Using `pipeline` also makes it easier to add compression which shrinks the response size significantly.

We currently use this implementation in production.

### DIFF
--- a/platform/functions/remix-server/regional-server.mjs
+++ b/platform/functions/remix-server/regional-server.mjs
@@ -51,7 +51,7 @@ const createApigHandler = (build) => {
         ...Object.fromEntries(response.headers.entries()),
         "Transfer-Encoding": "chunked",
       },
-      cookies: response.headers.getSetCookie(),
+      cookies: accumulateCookies(response.headers),
     };
 
     if (response.body) {
@@ -97,6 +97,21 @@ const createApigHandler = (build) => {
       writer.end();
     }
   });
+};
+
+const accumulateCookies = (headers) => {
+  // node >= 19.7.0 with no remix fetch polyfill
+  if (typeof headers.getSetCookie === "function") {
+    return headers.getSetCookie();
+  }
+  // node < 19.7.0 or with remix fetch polyfill
+  const cookies = [];
+  for (let [key, value] of headers.entries()) {
+    if (key === "set-cookie") {
+      cookies.push(value);
+    }
+  }
+  return cookies;
 };
 
 export const handler = createApigHandler(remixServerBuild);
--- a/platform/functions/remix-server/regional-server.mjs
+++ b/platform/functions/remix-server/regional-server.mjs
@@ -2,72 +2,9 @@
 // build and performs the Remix server rendering.
 
 import { createRequestHandler as createNodeRequestHandler } from "@remix-run/node";
-
-/**
- * Common binary MIME types
- */
-const binaryTypes = [
-  "application/octet-stream",
-  // Docs
-  "application/epub+zip",
-  "application/msword",
-  "application/pdf",
-  "application/rtf",
-  "application/vnd.amazon.ebook",
-  "application/vnd.ms-excel",
-  "application/vnd.ms-powerpoint",
-  "application/vnd.openxmlformats-officedocument.presentationml.presentation",
-  "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
-  "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
-  // Fonts
-  "font/otf",
-  "font/woff",
-  "font/woff2",
-  // Images
-  "image/bmp",
-  "image/gif",
-  "image/jpeg",
-  "image/png",
-  "image/tiff",
-  "image/vnd.microsoft.icon",
-  "image/webp",
-  // Audio
-  "audio/3gpp",
-  "audio/aac",
-  "audio/basic",
-  "audio/mpeg",
-  "audio/ogg",
-  "audio/wavaudio/webm",
-  "audio/x-aiff",
-  "audio/x-midi",
-  "audio/x-wav",
-  // Video
-  "video/3gpp",
-  "video/mp2t",
-  "video/mpeg",
-  "video/ogg",
-  "video/quicktime",
-  "video/webm",
-  "video/x-msvideo",
-  // Archives
-  "application/java-archive",
-  "application/vnd.apple.installer+xml",
-  "application/x-7z-compressed",
-  "application/x-apple-diskimage",
-  "application/x-bzip",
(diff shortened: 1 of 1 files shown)


platform/functions/remix-server/regional-server.mjs
```diff
@@ -107,51 +44,59 @@ const createApigHandler = (build) => {
...
-  }
-  // node < 19.7.0 or with remix fetch polyfill
-  const cookies = [];
-  for (let [key, value] of headers.entries()) {
-    if (key === "set-cookie") {
-      cookies.push(value);
-    }
-  }
-  return cookies;
-};
```
lukehrucker: This logic is here so that this lambda handler will work regardless of the node runtime and usage of remix's fetch polyfill. Removing it feels unrelated to your change?

JannikWempe: You are right, this is unrelated. It was inspired by the React Router v7 PR which also removes it: https://github.com/sst/sst/pull/5289/files#diff-d9ecacd8b64692e8372eaaa10480ce00d2aaea944f12ba54196a8063570c7af3

But I'll add it back in since this is not the point of this PR.

platform/functions/remix-server/regional-server.mjs
```diff
@@ -107,51 +44,59 @@ const createApigHandler = (build) => {
     context.callbackWaitsForEmptyEventLoop = false;
     const request = convertApigRequestToNode(event);
     const response = await requestHandler(request);
```
JannikWempe, line 46: This is the line of code where lots of requests get stuck at some point... Not sure if changes in the handler can fix this.

JannikWempe:
⚠️ UPDATE: The issues came back. Every once in a while (~all few hours) there are lots of requests that get stuck at `requestHandler(request)`.

JannikWempe:
I have closed this PR since the issues returned :-/

I have figured out that this is the line of code that occasionally hangs: `const response = await requestHandler(request);` But I have no clue why...

And it seemed the compression disables streaming (or at least the `Suspense` fallback isn't shown anymore). Also, I came to the realization that probably CloudFront should handle compression. But I am not sure if they support that with streamed content. It seemed like they require the `Content-Length` header for it to work.

I have found some other implementations of the Lambda handler for Remix ([like this](https://github.com/tappleby/remix-aws-streaming-example/blob/5c4fe9ac98cde8373ce61dad92a05c4f279c5593/lambda/request-handler.ts)) but it seemed they suffer the the same issue...

Sidenote: why is SST not using `writeReadableStreamToWritable` from `@remix-run/node`? 🤔

And would it be a good idea to use an `AbortSignal` for the `Request` that aborts if the stream ends?

I am not even sure anymore if the issue is SST / Lambda handler related, BUT this is the current implementation that we use that doesn't seem to suffer from hanging requests for some reason 🤷🏼‍♂️

```ts
// This is a custom Lambda URL handler which imports the Remix server
// build and performs the Remix server rendering.

import { createRequestHandler as createNodeRequestHandler } from "@remix-run/node";

function convertApigRequestToNode(event) {
  if (event.headers["x-forwarded-host"]) {
    event.headers.host = event.headers["x-forwarded-host"];
  }

  const search = event.rawQueryString.length ? `?${event.rawQueryString}` : "";
  const url = new URL(event.rawPath + search, `https://${event.headers.host}`);
  const isFormData = event.headers["content-type"]?.includes(
    "multipart/form-data",
  );

  // Build headers
  const headers = new Headers();
  for (let [header, value] of Object.entries(event.headers)) {
    if (value) {
      headers.append(header, value);
    }
  }

  return new Request(url.href, {
    method: event.requestContext.http.method,
    headers,
    body:
      event.body && event.isBase64Encoded
        ? isFormData
          ? Buffer.from(event.body, "base64")
          : Buffer.from(event.body, "base64").toString()
        : event.body,
  });
}

const streamToNodeStream = async (reader, writer) => {
  try {
    let readResult = await reader.read();
    while (!readResult.done) {
      // Add error handling for write operation
      if (!writer.write(readResult.value)) {
        // Handle backpressure - wait for drain event
        await new Promise((resolve) => writer.once("drain", resolve));
      }
      readResult = await reader.read();
    }
  } catch (error) {
    console.error("Streaming error:", error);
    // Ensure we close the stream even on error
    writer.end();
    throw error;
  }
  writer.end();
};

const createApigHandler = (build) => {
  const requestHandler = createNodeRequestHandler(build, process.env.NODE_ENV);

  return awslambda.streamifyResponse(async (event, responseStream, context) => {
    try {
      // Calculate remaining time for the Lambda execution
      const timeoutMs = context.getRemainingTimeInMillis();
      const timeoutPromise = new Promise((_, reject) => {
        setTimeout(
          () => {
            reject(new Error("Lambda timeout"));
          },
          Math.max(timeoutMs - 500, 0),
        ); // Leave 500ms buffer
      });

      context.callbackWaitsForEmptyEventLoop = false;
      const request = convertApigRequestToNode(event);

      // Race between the request handling and timeout
      const response = await Promise.race([
        requestHandler(request),
        timeoutPromise,
      ]);

      const httpResponseMetadata = {
        statusCode: response.status,
        headers: {
          ...Object.fromEntries(response.headers.entries()),
          "Transfer-Encoding": "chunked",
        },
        cookies: accumulateCookies(response.headers),
      };

      const writer = awslambda.HttpResponseStream.from(
        responseStream,
        httpResponseMetadata,
      );

      if (response.body) {
        await streamToNodeStream(response.body.getReader(), responseStream);
      } else {
        writer.write(" ");
      }
      writer.end();
    } catch (error) {
      console.error("Handler error:", error);
      // Send error response to client
      const writer = awslambda.HttpResponseStream.from(responseStream, {
        statusCode: 504,
        headers: {
          "Content-Type": "application/json",
        },
      });
      writer.write(JSON.stringify({ error: "Gateway Timeout" }));
      writer.end();
    }
  });
};

const accumulateCookies = (headers) => {
  // node >= 19.7.0 with no remix fetch polyfill
  if (typeof headers.getSetCookie === "function") {
    return headers.getSetCookie();
  }
  // node < 19.7.0 or with remix fetch polyfill
  const cookies = [];
  for (let [key, value] of headers.entries()) {
    if (key === "set-cookie") {
      cookies.push(value);
    }
  }
  return cookies;
};

export const handler = createApigHandler(remixServerBuild);

```

PR #5461, opened 2025-02-13
Queue reference method

OldSneerJaw:
Adds a static `get` method to the `Queue` class so that, like other lower-level SST components, one can retrieve a reference to an SQS queue from outside the app/stage. Complements the existing static `subscribe` method that allows one to subscribe to an outside SQS queue without having a reference to it. Inspired by the static `get` method in the `SnsTopic` component.

### DIFF
--- a/platform/src/components/aws/queue.ts
+++ b/platform/src/components/aws/queue.ts
@@ -294,7 +294,7 @@ export interface QueueSubscriberArgs {
 
 interface QueueRef {
   ref: true;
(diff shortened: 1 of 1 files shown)


OldSneerJaw:
@fwang: FYI, this is nearly identical to the final version of the introduction of `SnsTopic.get` in  #5440.

fwang:
Will be in the next release.

Thanks @OldSneerJaw!

Issue #4944, opened 2024-10-22
Redis host is null when cluster mode is disabled


zacwood9:
I'm trying to use the `Redis` component to setup a Redis instance for a [Sidekiq](https://sidekiq.org/) job queue. [Redis cluster mode is not appropriate for Sidekiq](https://github.com/sidekiq/sidekiq/wiki/Using-Redis#architecture) so I disabled it:

```typescript
const redis = new sst.aws.Redis('Redis', {
  vpc,
  nodes: 1,
  transform: {
    cluster: (args) => {
      args.clusterMode = 'disabled';
      args.automaticFailoverEnabled = false;
    },
  },
});
```

However, `redis.host` is now `null`. 

When cluster mode is disabled, the host can be found in the [`primaryEndpointAddress` field](https://www.pulumi.com/registry/packages/aws/api-docs/elasticache/replicationgroup/#primaryendpointaddress_nodejs). The SST component should use the appropriate value for `host` depending on if `clusterMode` is enabled.

zacwood9:
fixed in https://github.com/sst/sst/pull/5478#issuecomment-2811685800

PR #5478, opened 2025-02-19
Returning primaryEndpointAddress if Redis/Valkey cluster mode is disabled

luisdalmolin:
When you set cluster mode to "disabled" in Elasticache, the configuration endpoint becomes empty the the "Primary endpoint" should be used instead.

```ts
const redis = new sst.aws.Redis('SST-Laravel-Cache', {
  vpc,
  transform: {
    cluster: {
      automaticFailoverEnabled: false,
      clusterMode: 'disabled',
    },
  }
});
```

Screenshot 2025-02-19 at 08 37 29

### DIFF
--- a/platform/src/components/aws/redis.ts
+++ b/platform/src/components/aws/redis.ts
@@ -458,7 +458,11 @@ Listening on "${dev.host}:${dev.port}"...`,
   public get host() {
     return this.dev
       ? this.dev.host
-      : this.cluster!.configurationEndpointAddress;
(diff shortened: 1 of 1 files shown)


zacwood9:
Thank you! This would fix #4944.

fwang:
Added `cluster: false` option in v3.13.13, and `host` returns primary endpoint when set to false.

fwang:
Thanks @luisdalmolin!

PR #5512, opened 2025-03-01
Better monorepo support for Python Functions

apetta:
Introduce enhancements for handling monorepo paths for python function deployments.

The current implementation does not support monorepos using the `apps/*` and `packages/*` pattern (e.g. [turborepo](https://turbo.build/repo/docs/crafting-your-repository/structuring-a-repository)). I ran into challenges integrating Python Lambda functions into my monorepo that otherwise works well with SST.

A new `monorepoPath` field has been introduced to the Function config:

```typescript title="sst.config.ts"
new sst.aws.Function("MyPythonFunction", {
  runtime: "python3.11",
  python: {
    monorepoPath: "apps/python-backend/src",
  },
  handler: "apps/python-backend/src/functions/api.handler",
  url: true,
});
```

The monorepoPath is stripped from the handler prior to deployment to get it working as expected.

This has been tested locally and is working as expected.

++ Add a new AWS Python & Node.js monorepo example to demonstrate the feature.


apetta:
@walln  would appreciate your thoughts on this if any :)

walln:
I'm not sure I understand why this is necessary/what issue you were having. SST already supports python monorepos. In the examples there are several packages split apart but leveraged within the deployed functions. We use UV workspaces to do this. Is the current implementation not working for you?

apetta:
> I'm not sure I understand why this is necessary/what issue you were having. SST already supports python monorepos. In the examples there are several packages split apart but leveraged within the deployed functions. We use UV workspaces to do this. Is the current implementation not working for you?

Unfortunately, the current implementation isn't working for me using a [turborepo](https://turbo.build/repo/docs/crafting-your-repository/structuring-a-repository) directory structure, with other node apps in it already. The current implementation seems to require packages to be 1 level deep, and not nested. In my example, the packages are at `./apps/python-backend`, not at `./python-backend` relative to root (compared to the aws-python-container example).

`uv` builds this nested structure fine with the following config in the root `pyproject.toml`:

```toml
members = ["apps/python-backend"]
```

However, when SST tries to run the function, it faces an issue with the path as `uv` does not retain the nested directory structure but the SST handler path contains it: `apps/python-backend/src/functions/api.py` instead of just `functions/api.py` as per uv build output.

walln:
Hmm - I will try to reproduce. This seems like a bug rather than a missing feature. The current implementation is supposed to support what you want. I think we should fix without changing the API, especially since monorepo support was a huge part of the design intent.

apetta:
> Hmm - I will try to reproduce. This seems like a bug rather than a missing feature. The current implementation is supposed to support what you want. I think we should fix without changing the API, especially since monorepo support was a huge part of the design intent.

Cool! Thanks. Yes, that'd be ideal. I'm not super familiar with this repo's inner workings & the go language, so I appreciate it!

DanielAbdelNour:
I'm encountering a similar issue where handler paths get messed up when individual functions are grouped into a single parent directory within the root sst folder.

From the current examples, it seems like each Python function's directory needs to be directly within the root sst monorepo directory, which leads to a bloated folder structure—unless I've misunderstood the examples (which is quite possible).

aplotor:
@walln is it recommended to have one `pyproject.toml` per python lambda, or just a root one for all python lambdas that have different deps? i saw https://github.com/sst/sst/tree/dev/examples/aws-python says "builds currently do not tree shake so lots of workspaces can make larger builds than necessary" which seems to indicate the former should be done?

coming from the node lambda world, i use one root `package.json` for all the lambdas and they are correctly treeshaken. this also seems to be the way that the docs recommend. is there any reason it's not the same for python lambdas?

thanks !

apetta:
@walln Wondering if you managed to find some time to recreate this? :)

apetta:
@jayair Do you have any recommendations on including this new config field in the least disruptive manner?  It would remove the roadblock I'm facing with integrating Python support into a monorepo built with Turborepo. 

Happy to make any changes required for this; at least whilst we explore a better solution directly with `uv` - although I'm not sure how given the nesting of this monorepo structure (thus the handler path) and how `uv` does its bundling.

apetta:
Closing in favour of #6061

Issue #5534, opened 2025-03-06
Next.js construct - Introduce support for passing a custom schedule expression for the cron job responsible for warming


bojidaryovchev:
I just gotta say not having restriction on the PR description of a public repository is just lovely, mad props yo

I think there is some issue with my lambdas not being warm even tho I have it set up and I can see my logs for the warmer looking all good..

I open up my website from my laptop on my wifi and it takes 10 seconds to load (cold start apparently), then I open it on my phone (cellular network) and it loads within a second (warm), or the vice-versa scenario, all the same, I have defined `warm: 10` so I have 10 lambdas warmed concurrently and one speculation was that perhaps when I try to access it even tho theoretically I should have 10 lambdas warmed up and ready I could be having a cold one spawned for me and not hitting them somehow, but what I just described with laptop vs phone kind of counteracts this line of thinking..

so my only good speculation would be that decreasing the current default rate to something like 2 or 3 minutes could hypothetically solve my problem

I would appreciate your thoughts on this, thank you in advance

bojidaryovchev:
![Image](https://github.com/user-attachments/assets/09581ffe-450a-4da1-9a76-a4aed3919044)

![Image](https://github.com/user-attachments/assets/e6f789aa-7967-439c-9411-95f27798f8e2)

![Image](https://github.com/user-attachments/assets/77e03601-5475-4ddb-acab-c2289b79c513)

![Image](https://github.com/user-attachments/assets/20684d1c-977a-4ad5-a756-b7ecdbedcd6a)

ok so apparently 1 minute is like just right to ensure no cold starts, I have been trying the same flow I've described above and it looks all good

jayair:
Are you saying hitting it in under 1 min gives you a warm start?

bojidaryovchev:
@jayair exactly, this is what we can see from the logs, apparently the lambda container dies off in like 1m 30s to 10m or something like that on a random basis, but 1 minute seems to be enough to make sure it doesnt

as you can see I have a pull request to give us flexibility to define a custom scheduling rule and handle the issue gracefully

sommeeeer:
how big is your default server function? 10s+ cold starts seems very long. what ORM are you using? perhaps you could use `<Suspense>` and `aws-lambda-streaming` or do the data fetching client-side.

you probably want your index page to be as fast as possible. most preferably a `HIT` from Cloudfront.

bojidaryovchev:
[...]

PR #5535, opened 2025-03-06
feat: support for passing custom schedule expression to nextjs construct (#5534)

bojidaryovchev:


jayair:
Tell me more about the use case for this?

bojidaryovchev:
I mean it is all described thoroughly in the issue it is related to #5534

jayair:
Ah I missed that. Is this still a pressing issue for you?

bojidaryovchev:
@jayair this is something needed by anyone who is looking to deploy a Next.js app using SST and currently it is hardcoded with a value that does not ensure warmth and based on my discoveries presented in #5534 it would seem that ideally it should be 1 minute instead of 5, at least for my case that value seems golden, and with this PR we would be able to customize it freely

platform/src/components/aws/ssr-site.ts
```diff
@@ -112,6 +112,31 @@ export interface SsrSiteArgs extends BaseSsrSiteArgs {
...
+   *   warmExpression: `rate(5 minutes)`
+   * }
+   *
+   * @example
+   * // Use a cron expression to trigger warm-up requests at 12:00 UTC every day
+   * {
+   *   warmExpression: `cron(0 12 * * ? *)`
+   * }
+   */
+  warmExpression?: Input<`rate(${string})` | `cron(${string})`>;
```
dsonet, line 139: I'd suggest that `warmSchedule` would be a better fit here.

vimtor:
i'm thinking that maybe lowering the warm schedule to something a bit lower like 3-4 minutes would be the simplest solution

what do you think @bojidaryovchev? what schedule do you intend to use?

PR #5537, opened 2025-03-06
Disable exporting of uv dev dependencies

cvanes:
Right now, when building a Python lambda all dev dependencies will be included in the zip/container as dev dependencies are exported by default. This change disables that behaviour.


jayair:
@walln sorry to throw this your way.

Lmk if it looks okay to merge.

walln:
Yeah this looks good

jayair:
Thanks!

Issue #5550, opened 2025-03-12
Not good support to change bucket policies


pbalenzuela:
sst does not provide a useful way to change bucket policies. Currently, the only way to do it is by using transform. There are a few problems with this approach though. Here a  code snippet showing the approach

```ts
type PolicyDocument = Pick<
  aws.iam.GetPolicyDocumentArgs,
  'statements' | 'policyId'
>;
function toAimPolicyDocument(
  policyDocument: PolicyDocument
): Promise<aws.iam.PolicyDocument> {
  return aws.iam.getPolicyDocument(policyDocument).then((result) => {
    return {
      Statement: result.statements!.map((statement) => ({
        Effect: statement.effect,
        Action: statement.actions,
        Resource: statement.resources,
        Principal: statement.principals,
      })) as aws.iam.PolicyStatement[],
      Version: '2012-10-17',
      Id: result.id,
    };
  });
}
//create the bucket with a typed policy
new sst.aws.Bucket(bucketName, {
  transform: {
    policy: { policy: toAimPolicyDocument(policyDocument) }
  },
});
```

assigning a text based policy to the transform works too.

## What's the problem then?

- sst creates some default policies even when you don't want to
- It is not possible to create a policies targeting the newly created bucket because the bucket and the policy are created at the same time at `sst.aws.Bucket`'s constructor
- policies can replaced by text based policies but this works only for `string` policies, not `Input<string>`. This also means that they can not target other resources created by sst

## What's the solution?

Currently I had to change `.sst/platform/src/components/aws/bucket.ts` to make it work in the intended way. There are many in-line functions defined at `sst.aws.Bucket`'s constructor that prevents me from extending `sst.aws.Bucket` so I could take advantage of redefining some methods, `createBucketPolicy` in particular, which is actually an inline function as I said before.

## What's the expected behavior?

sst currently defines

```ts
access?: Input<"public" | "cloudfront">;
```

but it would be better to have it defined like this:

```ts
access?: Input<"public" | "cloudfront" | "denyUnsecureTransport">;
```

Making `denyUnsecureTransport` an explicit option instead of a forcefully implied one.

Additionally, a new `BucketArgs` property to provide additional policy statements

```ts
getAdditionalPolicyStatements?: (
  bucket: aws.s3.BucketV2,
) => GetPolicyDocumentStatement[];

export type GetPolicyDocumentStatement = Omit<
  NonNullable<aws.iam.GetPolicyDocumentArgs["statements"]>[number],
[...]

PR #5551, opened 2025-03-12
Allow changing bucket policies

pbalenzuela:
This PR targets resolving 
- https://github.com/sst/sst/issues/5550


platform/src/components/aws/bucket.ts
```diff
@@ -620,34 +649,43 @@ export class Bucket extends Component implements Link.Linkable {
...
-            {
-              test: "Bool",
-              variable: "aws:SecureTransport",
-              values: ["false"],
-            },
-          ],
-        });
+        if (
+          ["public", "cloudfront", "denyUnsecureTransport"].includes(
+            access ?? "",
```
dsonet, line 672: there is an extra `,` should be removed. And make this if condition one line would be much easier to read.

PR #5560, opened 2025-03-15
Add HTTP header routing support to ALB

### LINKED ISSUE #4179: AWS Cluster: Allow setting additional Rules on ALB listeners
I would like to put a CDN in front of the ALB and as per the AWS docs I want to add a conditional HTTP Header that the CDN will send so the ALB only access request from it.

Therefore I would like to be able to do the following:

```ts

const service = cluster.addService('MyService', {
  ...,
  public: {
   ports: [{listen: '80/http', rules: [
    {
      conditions: [
        {
          httpHeader: { values: ['randomstring'], httpHeaderName: 'X-ALB-Header' }
        }
      ],
      priority: 1,
      actions: [
        {
          targetGroupArn: targetGroup.arn, // this would hopefully come from the internals of sst's Service component
          type: 'forward
        }
      ]
    }
  ]]
  },
  transform: {
    listener: {
      defaultActions: [
        {
          fixedResponse: {
            contentType: 'text/plain',
            messageBody: 'Forbidden',
            statusCode: '403',
          },
          type: 'fixed-response',
        },
      ],
    },
  }
})
```

Or if we could expose the `targetGroup`s and `listener`s `arns` in an array that has the same order as `public.ports` that would also work

andrephilipsson:
Closes: https://github.com/sst/sst/issues/4179


fwang:
Thanks @andrephilipsson! Will be out in the next release.

Issue #3886, opened 2024-09-09
NextjsSite.buildDefaultServerCachePolicyProps() is no longer available?


iDVB:
We are transition this repo to https://github.com/sst/v2 and are closing this issue for now. If it is still relevant please reopen it there - sorry for the trouble

PR #5573, opened 2025-03-18
feat: expose default cache policy prop builder

jonlambert:
This PR implements the `buildDefaultServerCachePolicyProps` function that existed in previous versions of SST. It exposes the default properties used to build a cache policy for `Nextjs` + other `SsrSite` components for situations where users are hitting the 20 quota limit.

Like #3886, we've just copied the default properties from the SST source code into our `sst.config.ts`, but I thought I'd open this PR in case it's useful for others. This means if the default policy config changes moving forward, users can inherit the changes for free.

Changes in this PR: 

1. Logic to build out the server cache policy props has been extracted to an exposed function called `buildDefaultServerCachePolicyProps()` in both `ssr-site.ts` + `ssr-site-new.ts`
2. Logic to build out the Next.js server cache policy consumes this new function and passes in the default allowed headers, then exposes this as a static function on the `Nextjs` component. I've only included this for Next.js as this is the only sub-component to have specific allowed headers.

Please feel free to amend / discard this PR as needed. I followed the old legacy name for the function, but happy to amend or shift around if there's a more modern or better name and location for this 🙂

### DIFF
--- a/platform/src/components/aws/nextjs.ts
+++ b/platform/src/components/aws/nextjs.ts
@@ -8,6 +8,7 @@ import { Function } from "./function.js";
 import {
   Plan,
   SsrSiteArgs,
+  buildDefaultServerCachePolicyProps,
   createBucket,
   createDevServer,
   createServersAndDistribution,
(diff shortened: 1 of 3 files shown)


jayair:
Oh man, we should've messaged before this. 3.10 was a major change to handle that limit.

I haven't looked at your PR in depth. But wanna take a look at the latest release and see if it still applies?

Issue #4684, opened 2024-05-01
OAC should be enabled when using lambda streaming


ravenscar:
When using a Nextjs site if you enable streaming then the lambdas get publicly accessible URLs for use as origins for the CloudFront distribution.

If you have WAF enabled in your CF you may want to prevent the lambda URLs from being invoked directly via the public url, e.g. not via CF. AWS now supports OAC for lambdas and I think it would be secure to have this enabled by default in this case.

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-lambda.html

jrwpatterson:
Any update on this?

jrwpatterson:
> Aaaaalright, I think I finally got this mostly working. Here's my code.
> 
> UPDATE: Include ImageFunction thanks to @srobroek .
> 
> ```
>     const oac = new aws.cloudfront.OriginAccessControl('MyOAC', {
>       name: 'My OAC',
>       description: 'NextJS OAC for lambdas',
>       originAccessControlOriginType: 'lambda',
>       signingBehavior: 'always',
>       signingProtocol: 'sigv4',
>     });
> 
>     const appName = 'MySite';
>     const app = new sst.aws.Nextjs(appName, {
>       transform: {
>         server: {
>           url: {
>             authorization: 'iam',
>           },
>         },
>         imageOptimization: {
>           url: {
>             authorization: 'iam',
>           },
>         },
>         cdn(args) {
>           args.origins = [...args.origins].map((origin) => {
>             if (origin.originId !== 's3') {
>               return { ...origin, originAccessControlId: oac.id };
>             }
>             return origin;
>           });
>         },
>       },
>     });
> 
>     new aws.lambda.Permission('OACPermissionServer', {
>       action: 'lambda:InvokeFunctionUrl',
>       principal: 'cloudfront.amazonaws.com',
>       function: app.nodes.server?.nodes.function,
>       functionUrlAuthType: 'AWS_IAM',
>       sourceArn: app.nodes.cdn?.nodes.distribution.arn,
>     });
> 
>     // Hacky workaround because the nodes output contains server lambda but not imageOptimizer.
>     const imageOptimizationFunctionName = `${$app.name}-${$app.stage}-${appName}ImageOptimizerFunction`;
>     new aws.lambda.Permission('OACPermissionImage', {
>       action: 'lambda:InvokeFunctionUrl',
>       principal: 'cloudfront.amazonaws.com',
[...]

PR #5581, opened 2025-03-19
feat: add options to restrict ssr lambda functions to cloudfront-only access using OAC

MrHertal:
# Lambda Protection for SSR Components

This PR adds a comprehensive `protection` option to SSR components (Nextjs, TanStackStart, etc.) to control Lambda function URL access through CloudFront using AWS Origin Access Control.

## Key Motivation

**Enterprise Security Compliance**: In many companies, deploying public Lambda URLs is prohibited by cybersecurity policies. Security tools like Prisma Cloud detect public Lambda URLs as security violations, preventing deployment of SSR applications.

**With this PR, you can now use SST to deploy Next.js and other SSR frameworks in enterprise environments with strict security requirements!**

See the AWS documentation: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-lambda.html

## API Design

The `protection` option provides three security modes:

### 1. `"none"` (Default)
- Lambda URLs are publicly accessible
- Explicit public access permissions are created to ensure consistent behavior
- Works with external webhooks and callbacks out of the box

### 2. `"oac"` 
- Lambda URLs protected by CloudFront Origin Access Control
- Requires manual `x-amz-content-sha256` header for POST requests
- Use when you control all POST requests to your application

### 3. `"oac-with-edge-signing"`
- Full protection with automatic header signing via Lambda@Edge
- Works with external webhooks and callbacks seamlessly
- Higher cost and latency but works out of the box for all scenarios

## Advanced Configuration

For the `"oac-with-edge-signing"` mode, you can:

**Use auto-created Lambda@Edge function:**
```typescript
new sst.aws.Nextjs("MyApp", {
  protection: "oac-with-edge-signing"
});
```

**Customize the auto-created function:**
```typescript
new sst.aws.Nextjs("MyApp", {
  protection: {
    mode: "oac-with-edge-signing",
    edgeFunction: {
      memory: "256 MB",
      timeout: "10 seconds"
    }
  }
});
```

**Use existing Lambda@Edge function:**
```typescript
new sst.aws.Nextjs("MyApp", {
  protection: {
    mode: "oac-with-edge-signing",
    edgeFunction: {
      arn: "arn:aws:lambda:us-east-1:123456789012:function:my-function:1"
    }
  }
});
```

### Lambda@Edge Integration
- **Node.js 22.x runtime** for optimal performance
- **Automatic SHA256 header computation** for POST/PUT/PATCH requests
- **Smart request filtering** to minimize processing overhead
- **Proper IAM roles** with both `lambda.amazonaws.com` and `edgelambda.amazonaws.com` trust policies
- **us-east-1 deployment** requirement handled automatically

### Consistent Type System
- **Memory/timeout parameters** use the same types as SST Function (`Size`, `DurationSeconds`)
- **ARN validation** uses SST's consistent `FunctionArn` type and `parseLambdaEdgeArn()` helper
- **Error messages** follow SST's `VisibleError` patterns

### Multi-Region Support
- **All protection modes** work seamlessly with multi-region deployments
- **Permissions are created** for every server function in every region
- **Lambda@Edge functions** automatically handle global distribution

### Robust Permission Management
- **Explicit permissions** are always created to ensure consistent behavior
- **Mode switching** works reliably (no state transition issues)
- **Public access permissions** include the required `lambda:FunctionUrlAuthType=NONE` condition
- **CloudFront permissions** use proper `sourceArn` conditions

## Example Usage

```typescript
// Basic protection
const app = new sst.aws.Nextjs("MyApp", {
  protection: "oac"
});

// Full protection with Lambda@Edge
const front = new sst.aws.TanStackStart("Front", {
  protection: "oac-with-edge-signing"
});

// Multi-region with protection
const global = new sst.aws.Nextjs("GlobalApp", {
  regions: ["us-east-1", "eu-west-1", "ap-southeast-1"],
  protection: "oac-with-edge-signing"
});
```

## Benefits

✅ **Enterprise-ready security** - Meets strict cybersecurity compliance requirements  
✅ **Flexible protection levels** - Choose the right balance of security vs complexity  
✅ **External webhook support** - Lambda@Edge mode works with any external service  
✅ **Multi-region ready** - All modes work seamlessly across regions  
✅ **Type-safe configuration** - Consistent with SST's type system  
✅ **Reliable state management** - No issues when switching between modes

This enhancement makes SST's SSR components suitable for enterprise environments while maintaining the developer experience SST is known for.


jayair:
Hmm that file was a WIP as a part of the v3.10 release. Would appreciate if you made these changes to just the updated ssr-site.ts

MrHertal:
@jayair I have updated the PR

MrHertal:
@jayair @thdxr I found a better way to achieve this by using the distribution key value store

fwang:
Appreciate the PR @MrHertal!

Could u try out POST and PUT requests? According to this thread, might need to calculate and set the content hash header - https://github.com/sst/sst/issues/4684#issuecomment-2766606465

berenddeboer:
@MrHertal you can't use OAC, only supports GET, not POST/PUT etc etc. Mostly useless. Copy the implementation from sst v2, and use my PR https://github.com/sst/v2/pull/50 that fixes an import bug in the current implementation.

@fwang OAC is not good enough, as it requires clients to compute hashes so it's not transparent. Use the sst v2 method with my PR: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-lambda.html

fwang:
@berenddeboer this comment suggests calculating the hash in CF function, does that work?

https://github.com/sst/sst/issues/4684

berenddeboer:
> @berenddeboer this comment suggests calculating the hash in CF function, does that work?
> 
> #4684

ah thanks, got that now. Not sure I have a good setup to try this right now, but first question would be: any memory limits for CF functions? Probably ok in most cases, but how big can the body be?

MrHertal:
@fwang @berenddeboer Thanks for your reply 🙂

You cannot access request body in CloudFront function, so the `x-amz-content-sha256` header cannot be added automatically.

It seems doable in a Lambda@Edge, but that would introduce a lot of changes in the architecture.

The solution I propose is to offer an option to use OAC and protect lambda URL, by default it does not change anything, lambda URL will still be public. But if you enable it, it will setup the OAC and protect the URL behind CloudFront. You must then add the `x-amz-content-sha256` header to each POST and PUT requests sent from your application.

The implementation depends on your application framework. Here is an example of a `fetch` function that does the job:

```typescript
async fetchThroughOAC(
  input: string | Request | URL,
  init: RequestInit | undefined,
) {
  const body = init?.body?.toString() || "";
  const digestHex = await digestMessage(body);

  const headers = new Headers(init?.headers);
  headers.append("x-amz-content-sha256", digestHex);

  return fetch(input, { ...init, headers });
}

// See https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest
async function digestMessage(message: string) {
  const msgUint8 = new TextEncoder().encode(message);
  const hashBuffer = await crypto.subtle.digest("SHA-256", msgUint8);
  const hashArray = Array.from(new Uint8Array(hashBuffer));
  const hashHex = hashArray
    .map((b) => b.toString(16).padStart(2, "0"))
    .join("");

  return hashHex;
}
```

I can add an example for `TanStackStart` component since I'm currently working on it. Users will need to use a similar process for NextJS or other framework.

I understand that it is a bit cumbersome, but it's a good first step for users who don't want to deploy public lambda URL.

I work in a company where cloud resources are monitored by tools like Prisma, deploying unprotected lambda URL is a red flag.

With that option, I can at least enable the protection, knowing that for POST and PUT requests, I will need to add the payload hash value of the request body in the x-amz-content-sha256 header.

berenddeboer:
@MrHertal lambda@edge is another service which looks free on the surface, but starts popping up in the bills at scale. Possibly would work, but something to take into account.

berenddeboer:
@MrHertal this thing:

>You must then add the x-amz-content-sha256 header to each POST and PUT requests sent from your application.

That's not great, as it would require changes to applications. Imagine having to add this to all your graphql requests... Very very painful.

MrHertal:
@berenddeboer 

`protectedUrl` is optional, false by default. If for some reason adding the header is too much work and does not worth the benefit of having a protected URL, then people won't use it.

I think that having an option is better than no option 🙂

vincentvanderweele:
We recently upgraded to SST v3 especially because of the AOC support. Our security team flags public lambda urls as a critical security issue due to potential resource exhaustion.

We would like to upgrade to Nextjs 15 but for that we need to upgrade to the latest SST version, which no longer supports AOC.

So we would really like to have this support. It makes sense to keep it optional because of the additional header required for POST and PUT requests.

cgcompassion:
Question, does this solution also secure the Image Optimization lambda?

cgcompassion:
Hi, is there a way to use this, even before it merges? Like a patch? We are stuck on SST 3.7 until this merges, because of the changes to how NextJS runs on Cloudfront... Until we have this alternate way of making lambdas private, we can't update.... Most NextJS users will not be making POST endpoints... we would use API Gateway for that... NextJS is mostly focused on GET. Since this PR is providing an option, not changing the default, it should be fairly low risk, right? I could help test it if possible?

jayair:
I'll bump it up internally.

MrHertal:
Thanks @jayair 

I've added an example showing how to use TanStackStart with a protected Lambda URL.

The example includes a simple global middleware that handles the `x-amz-content-sha256` header requirement. It's actually pretty straightforward to implement!

Since many companies (including mine) don't allow public Lambda URLs for security reasons, I thought this example might be helpful. Let me know if you'd like to explore this further 😊

MrHertal:
@cgcompassion 

What I currently do is overriding `router.ts` and `ssr-site.ts` in the sst install folder:

```
sst install
cp this-pr-router.ts .sst/platform/src/components/aws/router.ts
cp this-pr-ssr-site.ts .sst/platform/src/components/aws/ssr-site.ts
```

alin-plamadeala:
any updates on this?

It seems like this issue also affects the ImageOptimizer lambda, so it would be nice to cover it as well

cgcompassion:
> I'll bump it up internally.

@jayair Sorry to bug, but any update here? Or guidance on what we can do to help?

edbramwell:
@MrHertal - adding the lambda@edge function wasn't significant given that it's applied on the Origin Request not the Viewer Request (so works in conjunction with the new Router component rather than against it; it's just a single additional resource (excl. perms/role etc).

Obviously with a small latency and cost penalty, but with no requirement to change application code (to add the `x-amz...` I wonder whether this is a neater solution?

edbramwell:
our AWS SA came back with the suggestion of putting API Gateway between the cloud front distribution and the lambda function - which would remove the `x-amz-content-sha256` header requirement. Food for thought.

berenddeboer:
@edbramwell the reason people don't like is it costs extra money. But let us know if this works for you!

MrHertal:
We've been using this option in production for 6 months with a Tanstack start app. It works great, even when we deploy our app in multiple regions with the regions option.

Adding the AWS header is totally manageable with things like a wrapper around fetch or a middleware.

Again, it's an option. If you don't need it, leave it undefined. If you need it, you have the possibility to use it. It's better than nothing.

I don't see any reason why this is not merged.

Ilnore:
My two cents: I'm really in favour of having this optional, but not default ability setup for OAC. I also am working in a context where we need OAC for securing the Lambda behind CloudFront.

The only place where this matters is POST / PUT requests. Do we have any indication how many users of OpenNext actually define any custom PUT / POST NextJS route handlers?

Of the arguments against Origin Access Control with simple flag and leaving responsibility of signature calculation to consumers, noted in the comments of this PR, I'd like to visit some:

> Question, does this solution also secure the Image Optimization lambda?

Image optimization shouldn't need PUT/POST, so it shouldn't suffer from OAC?


> our AWS SA came back with the suggestion of putting API Gateway between the cloud front distribution and the lambda function - which would remove the `x-amz-content-sha256` header requirement. Food for thought.

If there would be plans on adding further API Gateway resources in the mix, why not instead just have a separate API system outside NextJS? In my experience, many systems already do; especially larger systems where the website may not be the only consumer of API-like systems, but there may be a mobile app or other things in addition.


> @MrHertal this thing:
> 
> > You must then add the x-amz-content-sha256 header to each POST and PUT requests sent from your application.
> 
> That's not great, as it would require changes to applications. Imagine having to add this to all your graphql requests... Very very painful.

Do people generally use NextJS as GraphQL backend? I guess [@as-integrations/next](https://www.npmjs.com/package/@as-integrations/next) has 50k weekly downloads, so some apparently do.

But even those who do, may have convenient options to calculate the hashes on requesting side. Maybe it's not as painful as expected? For example `@apollo/client` has the [ApolloLink concept](https://www.apollographql.com/docs/deploy-preview/3d21a6ec9260ff42f380e9d8/react/api/link/introduction) for injecting middleware. `graphql-request` also has [middleware concept](https://github.com/graffle-js/graffle/blob/graphql-request/examples/other-middleware.ts).

MrHertal:
@Ilnore Hi,

> Image optimization shouldn't need PUT/POST, so it shouldn't suffer from OAC?

I can run a test with the NextJS component soon.

> If there would be plans on adding further API Gateway resources in the mix, why not instead just have a separate API system outside NextJS?

You cannot put an API Gateway between the CloudFront distribution and the Lambda because it won't work with Lambda URL and streaming response.

> Do people generally use NextJS as GraphQL backend?

Using GraphQL should not be a problem, there must be a way to use a single wrapper or middleware where you can add the header to all requests.

MrHertal:
This is now also working with Nextjs image optimizer function.
I have updated the PR description.

**With this PR, you can now use SST to deploy Next.js and other SSR frameworks in enterprise environments with strict security requirements!**

@jayair @fwang 

No more excuse to not merge this! 😄

edbramwell:
@MrHertal - unless I'm missing something, you can't use wrapper or middleware for `form_post` responses in a SAML flow. Given the "no public lambdas" probably has high overlap with SSO usage, I think it makes sense to have a solution on the infrastructure side, rather than requiring application code fixes?

MrHertal:
@edbramwell You're right, for POST requests from external sources, this option won't work without adding Lambda@Edge.

In my specific use case, I don't need external POST handling, so I prefer to avoid the additional infrastructure overhead. However, I recognize that many other scenarios would require this functionality - particularly when exposing APIs externally or implementing SAML authentication flows as you mentioned.

A potential solution could be making this option more flexible than a simple boolean. We could allow developers to choose their approach based on their specific requirements:

```
const app = new sst.aws.Nextjs("MyApp", {
  originAccessControl: "default", // Standard implementation or "edge" to handle all cases with Lambda@Edge
});
```

This would give developers the flexibility to opt into the Lambda@Edge overhead only when their use case requires it.

MrHertal:
@edbramwell @berenddeboer 

I've implemented a more complete approach. The original option has been evolved into a comprehensive `protection` system with three modes:

```typescript
// Simple protection (my use case)
new sst.aws.Nextjs("MyApp", {
  protection: "oac"  // Manual header signing required
});

// Full protection (external webhooks/SAML)
new sst.aws.Nextjs("MyApp", {
  protection: "oac-with-edge-signing"  // Automatic Lambda@Edge signing
});

// No protection (default)
new sst.aws.Nextjs("MyApp", {
  protection: "none"  // Public access
});
```

**Key improvements:**
- ✅ **Flexible security levels** - Choose the right balance for your use case
- ✅ **Lambda@Edge auto-creation** - Handles the infrastructure complexity automatically
- ✅ **External webhook support** - Works with SAML, payment providers, etc.
- ✅ **Multi-region ready** - All modes work across regions
- ✅ **Enterprise compliance** - Meets strict security requirements

The Lambda@Edge mode automatically computes SHA256 headers for POST requests, making it work seamlessly with external services while keeping the infrastructure overhead optional for those who don't need it.

vimtor:
planning on testing this by next week!
stay tuned @MrHertal

platform/src/components/aws/ssr-site.ts
```diff
@@ -1030,6 +1233,153 @@ async function handler(event) {
...
+        // Create the Lambda@Edge function using SST transform pattern
+        const edgeFunction = new aws.lambda.Function(
+          ...transform(
+            undefined,
+            `${name}EdgeFunction`,
+            {
+              runtime: "nodejs22.x",
+              handler: "index.handler",
+              role: edgeRole.arn,
+              code: new pulumi.asset.AssetArchive({
```
vimtor: could we move this function code to `platform/functions` instead of inline? just to keep it consistent with the rest of the codebase.

you'll find more examples in there

platform/src/components/aws/ssr-site.ts
```diff
@@ -1030,6 +1233,153 @@ async function handler(event) {
...
+        // Decode base64 to get the actual body content
+        bodyString = Buffer.from(request.body.data, 'base64').toString('utf8');
+      } else {
+        // If not base64 encoded, use as-is
+        bodyString = request.body.data;
+      }
+    }
+    
+    // Compute SHA256 hash of the body
+    const hash = crypto.createHash('sha256').update(bodyString, 'utf8').digest('hex');
```
vimtor: the below code fixed the file upload issue 

still, lambda@edge has a 1mb payload cap, so big uploads return the failed to verify hash error

maybe we can return a different error or put a heads-up in the docs

```suggestion
    // Get the request body as raw bytes (never convert to UTF-8 string)
    const bodyBuffer = request.body?.data
      ? (request.body.encoding === 'base64'
          ? Buffer.from(request.body.data, 'base64')
          : Buffer.from(request.body.data))
      : Buffer.alloc(0);

    // Compute SHA256 hash of the raw bytes
    const hash = crypto.createHash('sha256').update(bodyBuffer).digest('hex');
```

platform/src/components/aws/ssr-site.ts
```diff
@@ -121,6 +125,95 @@ export interface SsrSiteArgs extends BaseSsrSiteArgs {
...
+   *   lambdaProtection: {
+   *     mode: "oac-with-edge-signing",
+   *     edgeFunction: {
+   *       arn: "arn:aws:lambda:us-east-1:123456789012:function:my-signing-function:1"
+   *     }
+   *   }
+   * }
+   * ```
+   */
+  lambdaProtection?: Input<
```
vimtor: it’d be cool if we had this in the `Router` component too, so you’ve got protection when using it with an api for example.  

not something to change for this pr, but maybe it makes sense to move this type to `Router`

platform/src/components/aws/ssr-site.ts
```diff
@@ -121,6 +125,95 @@ export interface SsrSiteArgs extends BaseSsrSiteArgs {
...
+    | "none"
+    | "oac"
+    | "oac-with-edge-signing"
+    | {
+        mode: "oac-with-edge-signing";
```
vimtor, line 202: when deleting a stage that uses `"oac-with-edge-signing"` it takes ~5-10 minutes for the replicated functions to be deleted, so the edge function deletion takes a while

maybe we can add a small `::note` in the docs

vimtor (changes requested):
thanks for the contribution @MrHertal 

the only issue i found is file uploads with `"oac-with-edge-signing"`:

Image

this happens because the body bytes are being corrupted by converting to UTF-8 before hashing. which means that any request with non-UTF-8 bytes will fail, not just file uploads. i've suggested a fixed that seems to work

i've left other minor suggestions but everything else worked like a charm

platform/src/components/aws/ssr-site.ts
MrHertal: Good idea! I'll keep this PR focused on SSR sites, but I can open a follow-up issue to add lambdaProtection support to Router. At that point we can refactor the type to a shared location.

platform/src/components/aws/ssr-site.ts
MrHertal: Indeed, much better. I applied the change here: https://github.com/sst/sst/pull/5581/commits/61f63e2e3a574112eaaf50c7c00dfecf3d194c64

platform/src/components/aws/ssr-site.ts
MrHertal: Nice catch, I had not tested the file upload, shame on me.
I've added your suggested code, a note about the file size limit and also a better error type: https://github.com/sst/sst/pull/5581/commits/0791c643df9f7a887cb0a36fc40ca46a9d814bfa

platform/src/components/aws/ssr-site.ts
MrHertal: Note added: https://github.com/sst/sst/pull/5581/commits/e67507371df35b87725458fa88d87ee7f31bfff7

MrHertal:
Thank you @vimtor for your relevant comments.
It's great to see how things are moving with you and the new support team.

Could you test my last changes? I could not find a way to properly build the lambda@edge.
Before I was copying components right in a project with SST to check if they worked.

platform/src/components/aws/ssr-site.ts
vimtor: don't worry. your pr is one of the few that worked first try haha

platform/src/components/aws/ssr-site.ts
vimtor: good! let me know if you want me to create the issue

platform/functions/oac-edge-signer/index.ts
```diff
@@ -0,0 +1,86 @@
...
+  Records: Array<{
+    cf: {
+      request: CloudFrontRequest;
+    };
+  }>;
+}
+
+export async function handler(
+  event: CloudFrontEvent,
+): Promise<CloudFrontRequest | CloudFrontResponse> {
```
vimtor:
```suggestion
import { CloudFrontRequestHandler } from "aws-lambda";
import crypto from "node:crypto";

export const handler: CloudFrontRequestHandler = async (event) => {
```

platform/src/components/aws/ssr-site.ts
```diff
@@ -121,6 +125,103 @@ export interface SsrSiteArgs extends BaseSsrSiteArgs {
...
+   *
+   * :::note
+   * When removing a stage that uses `"oac-with-edge-signing"`, deletion may take 5-10 minutes while AWS removes the Lambda@Edge replicated functions from all edge locations.
+   * :::
+   *
+   * @example
+   * ```js
+   * // No protection (default)
+   * {
+   *   lambdaProtection: "none"
```
vimtor: i was thinking about how we'd name this attribute once we move it to the `Router` component

as far as i know you can protect lambdas and buckets for the most part. would we have a `bucketProtection`? not sure if that makes sense

this got me thinking maybe we should move this to the router and require people to pass the router instance to the `SsrSite` if they wanna use this feature

what do you think @MrHertal?

vimtor (commented):
i've tested the code and it works!

Image

just left a couple of suggestions before merging

platform/src/components/aws/ssr-site.ts
MrHertal: Thanks for the feedback @vimtor 

**Regarding buckets**: Looking at the Router code, bucket routes already have OAC protection enabled by default (via originAccessControlId). The specific security violation I'm addressing is public Lambda Function URLs - tools like Prisma Cloud flag these as security risks, but S3 buckets behind CloudFront with OAC are fine.

**Regarding Router support**: Agreed this would be useful there too! The Router's url routes currently use customOriginConfig without OAC. In a future PR, we could add something like:

```
router.route("/api/*", api.url, {
  protection: "oac" // or "oac-with-edge-signing"
})
```

For this PR, I'd prefer to keep the scope limited to SSR sites since:
  1. SSR sites are the most common case for Lambda function URLs
  2. The protection is tightly coupled with how SSR sites create their Lambda functions (setting authorization: "iam" on the function URL)
  3. It works today without requiring users to change how they create their sites

Once this lands, we can discuss the best API for Router in a follow-up. The type definitions and Lambda@Edge function could be shared. What do you think?

vimtor: @MrHertal i love this api you suggested for the router. you're right, let's focus first on ssr sites and then we do the router.

the only thing missing then (i promise) is the `lambdaProtection` arg name

given that the router will have `protection` i wonder if it would be a better, more elegant name (in the sst philosophy style) for the arg. what do you think?

MrHertal: I agree, let me rename it

MrHertal: done! https://github.com/sst/sst/pull/5581/commits/d11b8f258cfe4f63cfffb586e7c534b3aa5afa14

vimtor (approved):
approved! great work @MrHertal 

i will try to get this merged this week

<details>

<summary>i've tested the Next.js and Nuxt examples</summary>

#### `protection: "none" | undefined`

Image

Image

Image

#### `protection: "oac"`

Image

Image

Image

#### `protection: "oac-with-edge-signing"`

Image

Image

Image

</details>

edbramwell:
@MrHertal great work here & @vimtor too - looking forward to testing this (and then removing my monkey-patch)

masterFlippy:
Anyone else running in to issues with DELETE request?     "message": "The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details."

vimtor:
i remember something about the DELETE requests when i was looking into this

@masterFlippy can you open an issue? ideally add a sample of your sst.config.ts

@MrHertal any ideas?

masterFlippy:
> i remember something about the DELETE requests when i was looking into this
> 
> @masterFlippy can you open an issue? ideally add a sample of your sst.config.ts
> 
> @MrHertal any ideas?

Yeah sure! I'll do it during the day!

https://github.com/anomalyco/sst/issues/6427

PR #5582, opened 2025-03-19
fix: /component/aws/react.ts: typo Astro → React

LukasMurdock:


jayair:
Thanks!

We might've fixed this elsewhere.

PR #5583, opened 2025-03-20
Fix `sst stage list` showing deleted stages

### LINKED ISSUE #5566: [SST-2253] Duplicates in `state list` command
Thanks @jonlambert for the recent `state list` command [https://github.com/sst/sst/pull/5242#event-16716412671](https://github.com/sst/sst/pull/5242#event-16716412671)

Any idea why I'm seeing a duplicate local stage?

image

But in the Console I don't see it?

image

<sub>From [SyncLinear.com](https://synclinear.com) | [SST-2253](https://linear.app/sst-dev/issue/SST-2253/duplicates-in-state-list-command)</sub>

jonlambert:
When a stage is removed using `sst remove`, the resources are stripped from the state file, but the file remains in the backend. This PR reads the state file for each stage to check whether it contains any resources before including in the list.

This has the potential to become pretty unwieldy for users that create a lot of stages (PR stages, for example), but I don't think there's any other way of getting this info. I guess the main question is: should the state file be removed altogether when a user removes a stage? If yes, this PR isn't needed 🙂

Fixes #5566, #5567. Tested for local, aws and cloudflare backends.

### DIFF
--- a/pkg/project/provider/aws.go
+++ b/pkg/project/provider/aws.go
@@ -23,6 +23,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/aws/smithy-go"
 	"github.com/sst/sst/v3/internal/util"
+	"github.com/sst/sst/v3/pkg/state"
 
 	ecrTypes "github.com/aws/aws-sdk-go-v2/service/ecr/types"
 	s3types "github.com/aws/aws-sdk-go-v2/service/s3/types"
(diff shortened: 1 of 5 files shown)


qiqi112211 (approved):

jayair:
Thanks for digging into this

thdxr:
yeah we likely should remove the state file when all resources are removed. i didn't do that initially because i didn't want a bug that deleted it by accident but we're good now. would take a PR for that

PR #5603, opened 2025-03-24
Nextjs: set NODE_ENV to production as default in nextjs component

sommeeeer:
We should set `NODE_ENV` as `production` as default to avoid some issues in Next. One thing that can happen is that Next/React is trying to load modules that is meant for development purposes if `NODE_ENV` is undefined. in OpenNext we set the env variable aswell [here](https://github.com/opennextjs/opennextjs-aws/blob/42da2d96b8d49225d03b401520e13b82f7fa3711/packages/open-next/src/adapters/server-adapter.ts#L10). Sometimes that is not enough cause the imports happen before that line is invoked.

We need to remember to do this on every splitted function aswell when the new routing model supports it. i added a TODO so we remember it. 

OpenNext will probably patch this Next code in the future, using our new codepatcher during the build to avoid loading these env.

- https://github.com/opennextjs/opennextjs-aws/issues/624

### DIFF
--- a/platform/src/components/aws/nextjs.ts
+++ b/platform/src/components/aws/nextjs.ts
@@ -517,9 +517,8 @@ export class Nextjs extends SsrSite {
     { bucket }: { bucket: Bucket },
   ) {
     const parent = this;
-
-    const ret = all([outputPath, args?.imageOptimization]).apply(
-      ([outputPath, imageOptimization]) => {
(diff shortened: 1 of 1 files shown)


vimtor:
the related issue got fixed, is this still a problem @sommeeeer?

sommeeeer:
> the related issue got fixed, is this still a problem @sommeeeer?

This one can be closed! Its fixed by OpenNext now.

PR #5614, opened 2025-03-26
Fix compatibility with pulumi dynamic resources

user-0a:
I was running into some issues using pulumi dynamic resources with sst giving:

```
In <parent> component, the physical name of <child resource> (pulumi-nodejs:dynamic/<name>:<Name>) is not prefixed
```

The sst `Component` class currently checks for the prefix `"sst:"` on Resources / Components. This works for internal sst resources, but Pulumi Dynamic Resources typically have a prefix of `pulumi-nodejs:dynamic`. See here:

https://github.com/pulumi/pulumi/blob/1106c93d2d2bf353dca141bdf51e317d5e6c7d45/sdk/nodejs/dynamic/index.ts#L305

There was a check for `"pulumi-nodejs:dynamic:Resource"` already, but it only works for dynamic resources without a specified `module` or `type`.

This PR updates the checks in the file to check for both `sst:` and `pulumi-nodejs:dynamic` which fixes the issue. It also updates the `$transform`-related logic so that it works for dynamic resources as well.

### DIFF
--- a/platform/src/components/component.ts
+++ b/platform/src/components/component.ts
@@ -92,7 +92,7 @@ export class Component extends ComponentResource {
           // note: We are setting the default names here instead of inline when creating
           //       the resource is b/c the physical name is inferred from the logical name.
           //       And it's convenient to access the logical name here.
-          if (args.type.startsWith("sst:")) return;
+          if (args.type.startsWith("sst:") || args.type.startsWith("pulumi-nodejs:dynamic")) return;
           if (
             [
               // resources manually named
@@ -103,7 +103,6 @@ export class Component extends ComponentResource {
               "aws:servicediscovery/privateDnsNamespace:PrivateDnsNamespace",
               "aws:servicediscovery/service:Service",
               // resources not prefixed
-              "pulumi-nodejs:dynamic:Resource",
               "random:index/randomId:RandomId",
               "random:index/randomPassword:RandomPassword",
               "command:local:Command",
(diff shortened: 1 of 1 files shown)


user-0a:
Hey guys, small bump on this PR 🙏

@fwang

fwang:
Is the dynamic resource a child of an SST component? Is the relationship necessary?

user-0a:
Yeah something like this:

1. We have a dynamic resource that represents the lower-level resource.
2. This dynamic resource has other dependent/related resources like iam permissioning, devcommand, etc.
3. So we have a component that wraps the dynamic resource and also instantiates the other related resources.
4. In order for this component to support $transform, we need to use sst's `Component` class
5. So we get the error

There are definitely workarounds. Like not using a component, avoiding $transform on the component, etc. But it would be nice if it "just worked".

fwang:
> 4. In order for this component to support $transform, we need to use sst's Component class

Do you mean using`$transform` on `pulumi-nodejs:dynamic` resource type; or on your component?

`$transform(pulumi-nodejs:dynamic, ...)` should work even if the parent component is not an SST component.

user-0a:
We were using $transform on the parent component. The parent component owned the dynamic resource.

So to breakdown the changes:

1. We needed to use sst's `Component` (to support `$transform`). This causes the physical name prefix check to runs. This is the reason for updating that code to allow for components prefixed with `pulumi-nodejs:dynamic` (in addition to `sst:`).

2. Then I changed the code in `$transform` similarly, as I noticed it also enforced that the component has the prefix `sst:`. This part is not required but made sense given that the same requirement was relaxed on the resource side (in change 1). It could be the case that we simply want to enforce only components prefixed with `sst:` here though.

user-0a:
@fwang small nudge on this, happy to adjust if needed

PR #5629, opened 2025-03-29
fix: allow either `name` or `nameId` for `ApiGatewayV1DomainArgs`

ghostfuel:
I ran into this TS issue today when I began learning how to use SST.

I've simply mirrored these changes from `ApiGatewayV2`'s use of its domain args type and implementation!

Allows for the intended use case without TS errors;

```typescript
export const api = new sst.aws.ApiGatewayV1("Api", {
  domain: {
    nameId: "example.com",
  }
});
```


ghostfuel:
Apologies for not keeping this in-sync for ages, I've synced the forked branch and it appears it is still a required fix

PR #5632, opened 2025-03-30
docs: Typo in output filename

mikkopiu:
Should be `outputs.json` instead of `output.json` ([implementation writing the file](pkg/project/run.go)).


jayair:
Thanks!

Issue #5128, opened 2024-11-25
Feature request: Expose the load balancer Target Groups in `sst.aws.Service`


niss36:
Hi, would it be possible to return the target group(s) created by SST in `Service.nodes`?

My use case is monitoring, I need to refer to the load balancer and target group to create CloudWatch alarms.

Current state:

```ts
const cluster = new sst.aws.Cluster(...);
const service = cluster.addService(...);

new aws.cloudwatch.MetricAlarm('Health', {
  // ...
  namespace: 'AWS/ApplicationELB',
  metricName: 'UnHealthyHostCount',
  dimensions: {
    LoadBalancer: service.nodes.loadBalancer.arnSuffix,
    TargetGroup: '...', // hardcoded 😢
  },
});
```

Desired state:
```ts
const cluster = new sst.aws.Cluster(...);
const service = cluster.addService(...);

new aws.cloudwatch.MetricAlarm('Health', {
  // ...
  namespace: 'AWS/ApplicationELB',
  metricName: 'UnHealthyHostCount',
  dimensions: {
    LoadBalancer: service.nodes.loadBalancer.arnSuffix,
    TargetGroup: service.nodes.targetGroups.foo.arnSuffix,
  },
});
```

Thanks and let me know if you need any more context from me!

Trevato:
+1. trying to setup a NLB with `listen: "25565/tcp_udp"` (yeah baby, minecraft). I get:

```bash
~/projects/mc ❯ bun sst deploy --stage production
SST 3.10.16  ready!

➜  App:        mc
   Stage:      production

~  Deploy

|  Created     default_6_66_2 pulumi:providers:aws
|  Created     MCCluster sst:aws:Cluster
|  Created     MCVPC sst:aws:Vpc
|  Created     MCSservice sst:aws:Service → MCSserviceTaskRole aws:iam:Role
|  Created     MCSservice sst:aws:Service → MCSserviceExecutionRole aws:iam:Role
|  Created     MCCluster sst:aws:Cluster → MCClusterCluster aws:ecs:Cluster (10.7s)
|  Created     MCSserviceDev sst:sst:DevCommand
|  Created     MCSservice sst:aws:Service → MCSserviceLogGroupMCSservice aws:cloudwatch:LogGroup
|  Created     MCSservice sst:aws:Service → MCSserviceTask aws:ecs:TaskDefinition
[...]

PR #5633, opened 2025-03-30
fix: shorten load balancer target name prefix when using tcp_udp

Trevato:
load balancer target name prefixes must be 6 characters or less. when listening on a port using `tcp_udp`, deployment fails:

```
✕  Failed

MCSserviceTargetMCSserviceTCP_UDP25565 aws:lb:TargetGroup
aws:lb/targetGroup:TargetGroup resource 'MCSserviceTargetMCSserviceTCP_UDP25565' has a problem: "name_prefix" cannot be longer than 6 characters. Examine values at 'MCSserviceTargetMCSserviceTCP_UDP25565.namePrefix'.
```

more details in my report in #5128.

this gets it working for me. probably not a permanent fix but figured I'd share what I did to fix it.


jayair:
What happened with this PR? did it get out of sync?

Trevato:
yeah, @jayair not sure what happened. I can open a new one if you want and I can add a Minecraft on Fargate example. you need a workaround since ECS doesn't support port mapping both TCP and UDP. see: https://github.com/aws/containers-roadmap/issues/850

the PR was supposed to be just the single line change in `platform/src/components/aws/service.ts`.

jayair:
Yeah let's open a new one instead. Thanks!

vimtor:
fixed in [v3.18.9](https://github.com/anomalyco/sst/releases/tag/v3.18.9)

PR #5638, opened 2025-03-31
Support `provided.al2` runtime

berenddeboer:
This runtime is still supported by AWS and requirred by some lambda layers

In particular the bref.sh lambda layers require this runtime. This allows SST to use this runtime and enable older code to benefit from SST v3 functionality.


vimtor:
aws will drop `provided.al2` support in ~5 months, do you still need it @berenddeboer?

PR #5641, opened 2025-03-31
Fix: APIGateway with domain causes stage removal to fail

### LINKED ISSUE #5465: Simple API Gateway V2 with domain name causes removal to fail
Since upgrading to v3.7.1, all of my removals fail if I don't explicitly remove the domain name associated with the API Gateway instance.

This is example code:
```
const api = new sst.aws.ApiGatewayV2("Api", {
  domain: "some.domain.name",
  accessLog: {
    retention: "1 month",
  },
});

api.route(
  "POST /graphql",
  {
    handler: "packages/api/src/events/api.handler",
    logging: { retention: "1 month" },
    vpc: {
      securityGroups: [serviceSecurityGroupId],
      privateSubnets: vpc.privateSubnets,
    },
    link: [db],
  },
  { auth: { iam: true } },
);
```

It hangs when trying to delete the certificate because it's still in use.

brendankerrison:
Making the certificate an explicit dependency of the domain allows the correct order of dependencies to be deleted when removing the stage. 

Fixes #5465

### DIFF
--- a/platform/src/components/aws/apigatewayv2.ts
+++ b/platform/src/components/aws/apigatewayv2.ts
@@ -702,7 +702,12 @@ export class ApiGatewayV2 extends Component implements Link.Linkable {
     const logGroup = createLogGroup();
     const stage = createStage();
 
-    const certificateArn = createSsl();
+    const certificate = createSsl();
+    const certificateArn = all([domain, certificate]).apply(([domain, certificate]) => {
+      if (domain?.cert) return output(domain.cert);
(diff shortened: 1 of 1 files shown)


fwang:
Thanks @brendankerrison!

Fixed in a similar way in https://github.com/sst/sst/commit/8b0fdeda98039f57ff0401c627a26eb8c248c9d8

brendankerrison:
Thanks @fwang!

PR #5647, opened 2025-04-01
docs: updates for Cloudflare API tokens and AWS quota increase

AtAFork:
- clarification to explain that using the Cloudflare adapter requires a different way other than SST Secrets to add those environment variables (rephrasing of issue from user "grameens" in the discord)

- notes to help folks expedite increased AWS Lambda quota for SST Console


jayair:
Thanks!

PR #5677, opened 2025-04-09
Docs: add note for required preset for TanStackStart

lew:
```
 |  Error       
TanStackStart's app.config.ts must be configured to use the "aws-lambda" preset. It is currently set to "node-server".
```

I added a note to the docs with a required setting for TanStackStart, otherwise the example app won't build out of the box.


jayair:
Thanks!

PR #5689, opened 2025-04-11
Docs: Add Automatic Migrations to Drizzle with RDS Example

aidansunbury:
Originally written by @thdxr in [discord](https://discord.com/channels/983865673656705025/1322158305669550122/1328765378029027399), this snippet really helped me and is a common use case that was not in the docs! I added it to the docs and the example.


jayair:
Thanks for the PR! I might make a separate example out of this.

Issue #5673, opened 2025-04-08
sst.aws.StaticSite sets wrong cache headers on HTML files


Prinzhorn:
First time SST user here. My first deploy ever is a hello world vite + svelte app that just uses:

```ts
new sst.aws.StaticSite("MyApp", {
  build: {
    command: "npm run build",
    output: "dist",
  }
});
```

Docs say by default HTML files don't have a caching in the S3 headers:

https://github.com/sst/sst/blob/67940cbd59cde30523f1a0ddbad2875b1849bf1a/platform/src/components/aws/static-site.ts#L258-L274

but code is:

https://github.com/sst/sst/blob/67940cbd59cde30523f1a0ddbad2875b1849bf1a/platform/src/components/aws/static-site.ts#L886-L895

I must assume the second `**` overrides the previous rule for `*.html` because S3 console says:

![Image](https://github.com/user-attachments/assets/cef074af-ef88-4a1e-939c-19cc0ad49da7)

I manually added the alleged default config to my config and it's working properly now.

Prinzhorn:
Alright, either I'm super confused or the docs are very unclear. I changed the config to this:

```js
{
  assets: {
    fileOptions: [
      {
        files: ["**/*.css", "**/*.js"],
        cacheControl: "max-age=31536000,public,immutable",
      },
      {
        files: "**/*.html",
        cacheControl: "max-age=0,no-cache,no-store,must-revalidate",
      },
    ],
  }
}
```

but now it wasn't uploading the svg file. Docs regarding `fileOptions` say:

> _Specify the Content-Type and Cache-Control headers_ for specific files. This allows you to override the default behavior for specific files using glob patterns.

and

> A glob pattern or array of glob patterns of files _to apply these options to_.

no word that this _also_ specifies the files that are included in the deployment. I had to make this change:

```diff
{
  assets: {
    fileOptions: [
      {
-        files: ["**/*.css", "**/*.js"],
+        files: ["**/*.css", "**/*.js", "**/*.svg"],
        cacheControl: "max-age=31536000,public,immutable",
      },
      {
        files: "**/*.html",
        cacheControl: "max-age=0,no-cache,no-store,must-revalidate",
      },
    ],
  }
}
```

but shouldn't it deploy everything in `output`?

alexanderbh:
The issue is from here:
https://github.com/sst/sst/commit/f26a3f99fe4cb5184ec743ddef2fc97f316bb6fb

[...]

PR #5690, opened 2025-04-11
Fix .html files cache headers

alexanderbh:
Fixes:
https://github.com/sst/sst/issues/5673

The issue:
`.html` files have public caching headers set with 1 year lifetime. The commit here put the `fileOptions` in the wrong order:
https://github.com/sst/sst/commit/f26a3f99fe4cb5184ec743ddef2fc97f316bb6fb

This is severe. I expect more and more people having broken apps in production as they upgrade. This needs to be merged in asap @fwang


fwang:
Nice catch @alexanderbh! Releasing v3.13.10

PR #5696, opened 2025-04-13
Hyperdrive Example Fix

claughinghouse:
This updates the Hyperdrive Example to function properly. 

@fwang, @jayair, and @thdxr would you be able to take a look why this will deploy and work successfully only when first `sst deploy` is run with everything from `new set.aws.Service` to the end (line 45-131 and the return statements), then uncommenting on the second deploy it works. 

Hyperdrive fails to complete because the `cloudflared` container is not online and passing traffic.

I think the issue is the AWS Service taking a while to schedule a deployment and bring the container online. Is there a way to await another resource to be completely healthy?


claughinghouse:
Changed back to a draft since Cloudflare updated their provider to 6.x. I have a working deployment resolving issues noted above that I will clean up and add to this PR.

claughinghouse:
@fwang this is 99% ready with the new Cloudflare 6.1.1 provider but I would appreciate a look at the `dependsOn` section since deploying as-is causes the HyperdriveConfig to fail since DNS/Service is not ready. Stopping and restarting SST resolves the issue. 

This requires the [sst/provider](https://github.com/sst/provider) to be updated to support v6 fully.

cc: @BitPhinix

vimtor:
closing in favor of https://github.com/anomalyco/sst/pull/6774

PR #5697, opened 2025-04-13
Add example of how to enable an API key on a route

yellowtailfan:
The docs have an example of how to create an API key, but the API key only takes effect when it is explicitly enabled on the API routes. This proposal adds a description of this behaviour and an example.


jayair:
Thanks! I'll take a look.

fwang:
Just added `apiKey` to route in v3.14.15, ie.
```ts
api.route("GET /", "src/get.handler", {
  apiKey: true,
});
```

yellowtailfan:
Thanks @fwang for adding the `apiKey` option, that's much simpler!

Issue #5496, opened 2025-02-25
sst dev is not running


manularaai:
Using SST 3.9.10
On any function implemented, I get `sst dev is not running`.
The error might seem intermitent but gets more consistent when the function build time increases, such as when using a plugins file to compile decorators (my use case).

The error is caused because in live and dev mode, lambda's are deployed using .sst/platform/dist/bridge/bootstrap
This file comes from the source file .sst/platform/functions/bridge/bridge.go

In this file there is a line with this code:
`timeout := time.Second * 8`

This timeout of 8 seconds is being used when connecting to the sst dev process. If there is a latency of more than this timeout for any reason (in my case because of function building),  then you get the error `sst dev is not running`.

The workaround for this is to increase this timeout, lets say to 120 seconds. You will have to compile the boostrap file from source and then replace it in your project. When you run sst again, it deploys the new binary and now it doesn't timeout so fast.

The final workaround I propose is to use the env var `SST_FUNCTION_TIMEOUT` which is already being read, but not used, and use it to set the desired timeout to override the default timeout.

Here is the function I use to resolve the timeout:
```
// parseTimeout returns a time.Duration based on SST_FUNCTION_TIMEOUT.
// If the variable is not set or parsing fails, it returns 120 seconds.
func parseTimeout() time.Duration {
	// Default timeout is 120 seconds.
	defaultTimeout := time.Second * 120

	// If the environment variable is not set, return the default.
	if SST_FUNCTION_TIMEOUT == "" {
		return defaultTimeout
	}

	// Attempt to parse as a duration string (e.g., "150s" or "2m").
	if d, err := time.ParseDuration(SST_FUNCTION_TIMEOUT); err == nil {
		return d
	}

	// If parsing as a duration fails, try parsing it as an integer (seconds).
	if seconds, err := strconv.Atoi(SST_FUNCTION_TIMEOUT); err == nil {
		return time.Second * time.Duration(seconds)
	}

	// If all parsing fails, return the default timeout.
	return defaultTimeout
}
```

Then call the function to assing the timeout:
```
timeout := parseTimeout()
fmt.Printf("Timeout is set to: %v\n", timeout)
```

jayair:
Thanks for the detailed report.

Let me check internally. We could use a PR for this.

PR #5718, opened 2025-04-21
Configurable worker timeout

bill-kerr:
Addresses #5496 using the code suggested there, but I updated the default to Dax's recent increased default of 16s. I'm running into this issue quite regularly.

The changes in this PR allow you to configure the timeout with the `SST_FUNCTION_TIMEOUT` environment variable. It also adds the reason for the error to the returned error message.

### DIFF
--- a/platform/functions/bridge/bridge.go
+++ b/platform/functions/bridge/bridge.go
@@ -178,7 +178,7 @@ func run() error {
 }
 
 // parseTimeout returns a time.Duration based on SST_FUNCTION_TIMEOUT.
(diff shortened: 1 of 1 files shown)


thdxr:
you're still running into issues on latest sst? we fixed a bunch of problems that was causing this timeout to get hit

bill-kerr:
> you're still running into issues on latest sst? we fixed a bunch of problems that was causing this timeout to get hit

@thdxr We had some trouble with the routing changes introduced in `3.10.0` so we're still on `3.9.26`.

PR #5723, opened 2025-04-22
fix: pick correct bun on alpine arm64

### LINKED ISSUE #5708: SST breaks on alpine due to downloading glib version of bun
SST doesn't start on my CI node:alpine docker images; after enabling logging it shows:

```
time=2025-04-17T18:11:03.603Z level=INFO msg="bun selected" filename=bun-linux-aarch64.zip
time=2025-04-17T18:11:03.603Z level=INFO msg="bun downloading" url=https://github.com/oven-sh/bun/releases//download/bun-v1.2.1/bun-linux-aarch64.zip
```

That isn't the musl version of bun, which shows when it dies trying to do a `bun install`.

Looking at the 'which version of bun' code, it seems like the exceptions and the default cases might've switched?

[`sst/pkg/global/bun.go`](https://github.com/sst/sst/blob/166a612af2c9b421041cdbcfcf88afe8f62d1d02/pkg/global/bun.go#L68)
```go
switch {
	case goos == "darwin" && arch == "arm64":
		filename = "bun-darwin-aarch64.zip"
	case goos == "darwin" && arch == "amd64":
		if cpuid.CPU.Has(cpuid.AVX2) {
			filename = "bun-darwin-x64.zip"
		}
		filename = "bun-darwin-x64-baseline.zip"
	case goos == "linux" && arch == "arm64":
		if isMusl {
			filename = "bun-linux-aarch64-musl.zip"
		}
		filename = "bun-linux-aarch64.zip"
	case goos == "linux" && arch == "amd64":
		if isMusl {
			if cpuid.CPU.Has(cpuid.AVX2) {
				filename = "bun-linux-x64-musl.zip"
			}
			filename = "bun-linux-x64-musl-baseline.zip"
		}
		if cpuid.CPU.Has(cpuid.AVX2) {
			filename = "bun-linux-x64.zip"
		}
		filename = "bun-linux-x64-baseline.zip"
	case goos == "windows" && arch == "amd64":
		if cpuid.CPU.Has(cpuid.AVX2) {
			filename = "bun-windows-x64.zip"
		}
		filename = "bun-windows-x64-baseline.zip"
	default:
	}
```

kazdegroot:
This makes bun work on arm64 alpine, and probably a few other environments too.

This turned out to be two issues
 1. ldd --version has exit status 1, at least in my alpine container, which prevented the musl check from being executed
 2. as noted in the issue, all the exceptions are handled before the defaults

With these fixes the correct version of bun gets installed in my container, and starting sst works. Have not tested other environments.

Closes #5708

### DIFF
--- a/pkg/global/bun.go
+++ b/pkg/global/bun.go
@@ -57,9 +57,8 @@ func InstallBun(ctx context.Context) error {
 			isMusl = true
 		} else {
 			cmd := exec.Command("ldd", "--version")
-			if output, err := cmd.CombinedOutput(); err == nil {
-				isMusl = strings.Contains(strings.ToLower(string(output)), "musl")
-			}
+			output, _ := cmd.CombinedOutput()
+			isMusl = strings.Contains(strings.ToLower(string(output)), "musl")
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


vimtor (approved):

pkg/global/bun.go
```diff
@@ -57,9 +57,8 @@ func InstallBun(ctx context.Context) error {
 			isMusl = true
 		} else {
 			cmd := exec.Command("ldd", "--version")
-			if output, err := cmd.CombinedOutput(); err == nil {
-				isMusl = strings.Contains(strings.ToLower(string(output)), "musl")
-			}
+			output, _ := cmd.CombinedOutput()
+			isMusl = strings.Contains(strings.ToLower(string(output)), "musl")
```
vimtor, line 61: maybe this was changed by mistake?

```suggestion
			if output, err := cmd.CombinedOutput(); err == nil {
				isMusl = strings.Contains(strings.ToLower(string(output)), "musl")
			}
```

vimtor (changes requested):

pkg/global/bun.go
kazdegroot: As I mentioned in my original post, this was partially failing cause ldd —version didn’t return 0, at least on the alpine I tested with with, resulting in isMusl not being set, even if the stdout did in fact include musl. 

There might be a more idiomatic go way to write that, but checking for err == null was part of the problem in my test env.

vimtor:
do you think you could test this one @jamesgibbons92?

jamesgibbons92:
@vimtor 
This looks good to me. 
The error handling change is fine, we don't care about any error, only if musl is in the output. 
And the bug with the defaults being applied after the condition means the correct bun version is now installed. I verified this on the following systems:

aarch64 (glibc - ubuntu)
aarch64_glibc
aarch64 (musl - alpine)
aarch64_musl
amd64 (glbc - nixos)
amd64_glibc
amd64 (musl - void)
amd64_musl

vimtor (approved):

PR #5732, opened 2025-04-25
bad attempt at the function retrys with event invoke config...

natac13:
Sorry I could not seem to figure it out Frank. The permissions I think need to be done by creating another sst component I think I just was not completely sure but want to try this.

I look forward to see how you did it to learn

### DIFF
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -51,6 +51,8 @@ import {
 } from "./router.js";
 import { KvRoutesUpdate } from "./providers/kv-routes-update.js";
 import { KvKeys } from "./providers/kv-keys.js";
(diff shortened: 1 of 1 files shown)


vimtor:
@natac13 can you give me a bit more context about this?

natac13:
> @natac13 can you give me a bit more context about this?

Yep. I believe when I was attempting this I was trying to configure the retry amount for when a lambda is used as a destination for a dead letter queue if I remember correctly. There are defaults set by AWS that I was just trying to see if this would allow those to be changed. 

This is my best guess at what this was for. Happy to help if you need more.

vimtor:
@natac13 is it okay if i close the pr? not sure we (or you) need this?

natac13:
For sure close away

PR #5758, opened 2025-05-02
Feat: adding d1 static get

### LINKED ISSUE #5756: Add static get for Cloudflare D1
# Share Cloudflare D1 Databases Across SST Stages

## Problem
Currently, `Cloudflare D1 databases` from `sst` are not equipped with static get. Since it is a database, I think it should.

## Proposed Solution
Add a static `get` method to the D1 component that allows referencing existing D1 databases. 

### Example Usage
```typescript
// In your sst.config.ts
const d1 = $app.stage === "giorgio"
  ? sst.cloudflare.D1.get(
      "MyDatabase",
      "accountId",
      "databaseId"
    )
  : new sst.cloudflare.D1("MyDatabase");

// Link to a worker
new sst.cloudflare.Worker("Worker", {
  link: [d1],
  url: true,
  handler: "index.ts",
});
```

In nabla, we are actively working on a solution (as it is a necessity) and will provide a PR here.

giorgionocera:
This PR adds a static `get` method to the D1 Cloudflare database component, allowing users to reference existing D1 databases across stages. This is particularly useful for development workflows where you want to share the same database across different stages.

   Changes:
   - Added static `get` method to D1 component
   - Added related documentation

   Example usage:
   ```ts
   const d1 = $app.stage === "dev"
     ? new sst.cloudflare.D1("MyDatabase")
     : sst.cloudflare.D1.get(
         "MyDatabase",
         "accountId",
         "databaseId"
       );
   ```

## Technical Notes
- Uses Pulumi's `get` method for existing resources
- Handles the ID format differences between new and referenced databases

To retrieve the database ID from the Pulumi output, a workaround was necessary due to the merging of the database ID with the account ID in the `id` string when statically importing a database.

A potential solution to this issue is to incorporate the `uuid` field in the newer version of the Pulumi Cloudflare provider. This field **should** (to be checked) consistently match the database ID and **should** never be merged with the account ID.

## Test performed
- Test database sharing between stages
- Verify worker bindings work correctly
- Test backward compatibility

It fixes #5756.


platform/src/components/cloudflare/d1.ts
```diff
@@ -128,6 +155,60 @@ export class D1 extends Component implements Link.Linkable {
...
+   *
+   * ```ts title="sst.config.ts"
+   * return {
+   *   d1
+   * };
+   * ```
+   */
+  public static get(
+    name: string,
+    accountId: string,
```
vimtor, line 197: thanks for the contribution @giorgionocera 

maybe we should use DEFAULT_ACCOUNT_ID like we do when creating the database?

this way maybe we can avoid the trick you did on the getSSTLink

vimtor:
closed in favour of https://github.com/anomalyco/sst/pull/6420

(you have maintainer push disabled on your branch)

PR #5776, opened 2025-05-07
fix: load balancer namePrefix can only be 6 characters

Trevato:
remove the `_` from the namePrefix to abide by the AWS limit. also added minecraft on AWS example.

this PR is fixing my first one from earlier #5633


nicholasgubbins:
would love this to get merged - trying to set up a sip server and having this issue right now
cc @fwang

platform/src/components/aws/service.ts
```diff
@@ -1969,14 +1969,14 @@ export class Service extends Component implements Link.Linkable {
...
-                  // ie. "-1234567" is automatically added.
+                  // TargetGroup names allow for 32 chars, but an 6 letter suffix
+                  // ie. "-12345" is automatically added.
                   // - If we don't specify "name" or "namePrefix", we need to ensure
                   //   the component name is less than 24 chars. Hard to guarantee.
                   // - If we specify "name", we need to ensure the $app-$stage-$name
                   //   if less than 32 chars. Hard to guarantee.
                   // - Hence we will use "namePrefix".
-                  namePrefix: forwardProtocol,
+                  namePrefix: forwardProtocol.replace("_", ""),
```
vimtor, line 1974: if i'm not mistaken, we are doing this `.replace` for the TCP_UDP protocol case

could we add a comment so this becomes clearer?

examples/aws-minecraft/sst.config.ts
```diff
@@ -0,0 +1,47 @@
+/// <reference path="./.sst/platform/config.d.ts" />
+
+export default $config({
```
vimtor: could we move the Minecraft example into a separate PR?

vimtor (changes requested):
thanks for your contribution @Trevato

the fix works, i've left a couple of minor comments

PR #5791, opened 2025-05-11
Support shared Router in Auth and OpenControl components

jackblanc:
This PR modifies the `Auth` and `OpenControl` components to allow reusing an existing `Router`. 

Currently, both of these component apply the following transformation function to the function arguments:
```
        (args) => {
          args.url = {
            cors: false,
          };
        },
```

This will overwrite any `url` options passed in by the caller, for example: 
```
const router = new sst.aws.Router("Router", {
  domain: {
    name: "example.com",
    aliases: [`*.example.com`],
    dns: sst.cloudflare.dns(),
  },
})

// works as expected
export const ApiFunction = new sst.aws.Function("Api", {
  handler: "./packages/functions/src/api/function.handler",
  url: {
    router: {
      instance: router,
      domain: "auth.example.com",
    },
  },
});

// router options are ignored
export const AuthFunction = new sst.aws.Auth("Auth", {
  issuer: {
    handler: "./packages/functions/src/authorizer.handler",
    url: {
      router: {
        instance: router,
        domain: "auth.example.com",
      },
    },
  },
});
```

For the `Auth` component, I kept the transformation function but modified it so that other fields of `args.url` are combined with `cors: false`. For `OpenControl`, I moved the `cors: false` to functionBuilder's defaultArgs field, which feels more appropriate unless cors must always be false for `OpenControl`. In addition, I modified the behavior of Auth's url getter so that it only removes the last character of `this._issuer.url` if a trailing slash is present. 

I have tested this out in my own application by copying these files into my `.sst` directory.

### DIFF
--- a/platform/src/components/aws/auth.ts
+++ b/platform/src/components/aws/auth.ts
@@ -293,6 +293,7 @@ export class Auth extends Component implements Link.Linkable {
         },
         (args) => {
           args.url = {
+            ...(typeof args.url === "object" ? args.url : {}),
             cors: false,
           };
         },
--- a/platform/src/components/aws/auth.ts
+++ b/platform/src/components/aws/auth.ts
@@ -321,11 +321,15 @@ export class Auth extends Component implements Link.Linkable {
   /**
    * The URL of the Auth component.
    *
-   * If the `domain` is set, this is the URL with the custom domain.
+   * If the `domain` is set, this is the URL of the Router created for the custom domain.
+   * If the `issuer` function is linked to a custom domain, this is the URL of the issuer.
    * Otherwise, it's the auto-generated function URL for the issuer.
    */
   public get url() {
-    return this._router?.url ?? this._issuer.url.apply((v) => v.slice(0, -1));
+    return (
+      this._router?.url ??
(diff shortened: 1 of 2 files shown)


GMoon777:
Yes please!

devlsh:
would also really appreciate this - sharing a single `Router` throughout my infra would make life much easier

derekbking:
Would love to see this PR merged as well. Running into the same issue.

dragonautdev:
@fwang I'd also love to see this PR merged, as it should also help fixing the problem if Auth is hosted in a sub-path.

mschinis:
@fwang We have the same issue, and this PR would fix it for us. Can we please get this merged?

vikramasharma:
bump 🙏

aidanhchau:
Please can we get this PR merged 🙏🙏🙏

aidanhchau (approved):

devlsh:
would make life easier for sure - though from what i've seen/read, the guys are figuring out making things like sst more contributor focussed and having maintainers over going through them specifically, doesn't seem like that'll happen soonish

jayair:
Pinging internally

walln:
would really appreciate if this could be merged

PR #5794, opened 2025-05-12
Fix: Race condition when deleting ECS

### LINKED ISSUE #5678: Removing a deployed service fails because of cloud map namespace service dependency
When trying to remove a stack using `sst remove`, which has created a Service similar to the one below: 

```
const cluster = new sst.aws.Cluster('testCluster', {
  vpc: new sst.aws.Vpc('testVpc', {}),
});

export const service = new sst.aws.Service('testService', {
  cluster: cluster,
  capacity: 'spot',
  image: {
    context: './path',
    dockerfile: './path',
  },
  loadBalancer: {
    rules: [{ listen: '80/http', forward: '8080/http' }],
  },
});
```

We run into the following error:

`operation error ServiceDiscovery: DeleteNamespace, https response error StatusCode: 400, ResourceInUse: Namespace has associated services; delete the services before deleting the namespace`

It seems like it is trying to remove the cloud map namespace before it has removed the service that is linked to the namespace. We can get around this issue by manually removing the linked service through the AWS console, but then we get the an error saying it can no longer list the linked service:

`listing Service Discovery Instances: operation error ServiceDiscovery: ListInstances, https response error StatusCode: 400, ServiceNotFound`

This essentially puts the stage in a broken state, to recover we would need to remove the app state from S3.

RiFrost:
This fixes #5678

This adds a dependency to the `CloudmapService` pointing to the `CloudmapNamespace` and also a dependency to the `ECS Service` pointing to the `CloudmapService`. This way the race condition when removing the resources is handled correctly by pulumi.

Instead of using the id of the `PrivateDnsNamespace` in the `Service` construct I directly passed the Namespace to use it. As far as I can tell it doesn't break anything else.

I tested it by deploying and removing the following configuration:

```
export default $config({
  app(input) {
    return {
      name: "aws-service-discovery",
      home: "aws",
    };
  },
  async run() {
    const vpc = new sst.aws.Vpc("MyVpc", { nat: "ec2" });

    const cluster = new sst.aws.Cluster("MyCluster", { vpc });
    const service = new sst.aws.Service("MyService", { cluster });
  },
});
```


fwang:
Thanks @RiFrost!

Reverted some changes to leave the only missing dependency.

^ Other dependencies were already implicitly set, ie.
```ts
const ecsService = new ecs.Service("foo", {
  serviceRegistries: {
    registryArn: cloudmapService.arn, // ecsService automatically depends on cloudmapService
    ...
  },
}
```

bjornlll:
Sharing here in case related. When I try to sst remove my stages containing: VPC (nat = ec2, bastion), Service exposed through API Gateway, I get the following error relatively consistently: 

```
ArkiVPC sst:aws:Vpc → ArkiVPCPublicSubnet1 aws:ec2:Subnet
sdk-v2/provider2.go:515: sdk.helper_schema: deleting EC2 Subnet (subnet-008218cb1d22478a4): operation error EC2: DeleteSubnet, https response error StatusCode: 400, RequestID: 3560e326-4bb9-407b-83c8-5ad3260ca771, api error DependencyViolation: The subnet 'subnet-008218cb1d22478a4' has dependencies and cannot be deleted.: provider=aws@6.66.2
```

In AWS console, the resource blocking deletion is the VPC link created for exposing the service through the API Gateway.

PR #5819, opened 2025-05-21
Docs: Improvements to SQS examples

aidansunbury:
This PR cleans up a few components of the docs and enhances the basic SQS example to also showcase how to handle partial batch responses.

1. In `examples.mdx` both the main queue example and the dead-letter queue example were called "Subscribe to queues". This differentiates the naming to make it clear that it is in fact two different examples
2. SQS event handler type definitions existed for some of the examples but not all. This PR adds them in the places where they were missing
3. The main "Subscribe to queue" example was not particularly useful for demonstrating how to actually process the messages once they are in the read by the lambda handler. This pr adds basic scaffolding of looping through each record on the SQS event and demonstrates how to use partial batch responses.
4. `aws-bucket-queue-subscriber` did not have a package.json

### DIFF
--- a/examples/aws-bucket-queue-subscriber/package.json
+++ b/examples/aws-bucket-queue-subscriber/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "aws-dead-letter-queue",
+  "name": "aws-bucket-queue-subscriber",
   "version": "1.0.0",
   "description": "",
   "type": "module",
@@ -15,8 +15,5 @@
   "devDependencies": {
     "@types/aws-lambda": "^8.10.149",
     "sst": "3.0.37"
-  },
-  "dependencies": {
-    "@aws-sdk/client-sqs": "^3.515.0"
   }
 }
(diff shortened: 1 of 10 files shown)


jayair:
Thanks for this!

Just need one change here, the `examples.mdx` is generated from the `examples/` dir. So this part that you added: https://github.com/sst/sst/pull/5819/files#diff-a69230b69d1ba0091de5f6520443f5a5398f10b3e4dc7f2827111609ea63d336

Should be moved here: https://github.com/sst/sst/blob/dev/examples/aws-dead-letter-queue/sst.config.ts#L6

Lmk if that makes sense.

aidansunbury:
I added to the additional info to the `sst.config.ts` files, but I'm not sure that it will build correctly / was formatted correctly as I kept getting errors trying to build it locally.

Screenshot 2025-05-26 at 11 01 28 PM

jayair:
I'll take a look. Thanks!

vimtor (approved):
confirmed that the docs error is not coming from this PR #6263

CleanShot 2025-12-06 at 18 58 03@2x

thanks for your contribution @aidansunbury 

we should be able to merge this soon!

Issue #5657, opened 2025-04-03
Tunnel not working after upgrade to 3.12.3


andrewcoelho:
I updated from 3.9.4 to 3.12.3 which may or may not be related, but since then I can't get the tunnel to work.

```
// npx sst tunnel --stage production --print-logs

time=2025-04-03T09:46:47.644-04:00 level=INFO msg="checking for pulumi" path="/Users/andrew/Library/Application Support/sst/bin/pulumi"
time=2025-04-03T09:46:48.541-04:00 level=INFO msg="checking for bun" path="/Users/andrew/Library/Application Support/sst/bin/bun"
time=2025-04-03T09:46:49.155-04:00 level=INFO msg="starting tunnel"
time=2025-04-03T09:46:49.155-04:00 level=INFO msg="initializing project" version=3.12.3
time=2025-04-03T09:46:49.156-04:00 level=INFO msg="esbuild building" out=/Users/andrew/Projects/xxx/.sst/platform/sst.config.1743688009156.mjs
time=2025-04-03T09:46:49.166-04:00 level=INFO msg="esbuild built" outfile=/Users/andrew/Projects/xxx/.sst/platform/sst.config.1743688009156.mjs
time=2025-04-03T09:46:49.167-04:00 level=INFO msg="evaluating config"
time=2025-04-03T09:46:49.202-04:00 level=INFO msg="config evaluated"
time=2025-04-03T09:46:49.204-04:00 level=INFO msg="checking platform"
time=2025-04-03T09:46:49.205-04:00 level=INFO msg="checking provider" name=command version=1.0.2 compare=1.0.2
time=2025-04-03T09:46:49.205-04:00 level=INFO msg="loading home"
time=2025-04-03T09:46:50.007-04:00 level=INFO msg="aws credentials found" region=us-east-2 profile=xxx-prod
time=2025-04-03T09:46:50.007-04:00 level=INFO msg="fetching bootstrap"
time=2025-04-03T09:46:50.251-04:00 level=INFO msg="found existing bootstrap" data="{\"version\":5,\"asset\":\"sst-asset-xxx\",\"assetEcrRegistryId\":\"xxx\",\"assetEcrUrl\":\"xxx.dkr.ecr.us-east-2.amazonaws.com/sst-asset\",\"state\":\"sst-state-xxx\",\"appsyncHttp\":\"\",\"appsyncRealtime\":\"\"}"
[...]

PR #5820, opened 2025-05-21
Export elastic ip for tunnel when present.

gkpln3:
Solves #5657


mikecfisher:
Whats the status on this?

jescalan:
I run into problems with the tunnel every time I update the version of SST and it's very painful to resolve. I am really hoping this might fix it? 🙏

lputnam2000:
@jescalan Did this fix it for you?

lputnam2000 (approved):

jackwatters45:
Can confirm this fixed the issue for me. @lputnam2000 @fwang

brandonin:
Do we know when this will be merged?

pauljasperdev:
~~can confirm this~~

https://github.com/sst/sst/issues/5657#issuecomment-3183422968
> > I've been able to fix the issue, see [#5820](https://github.com/sst/sst/pull/5820).
> 
> ~works for me also.~
> 
> ~I rebased [@gkpln3](https://github.com/gkpln3)'s branch with the current and followed build steps for CLI. I use the executable with current SDK `3.17.10`~
> 
> EDIT: does NOT fix it for me. Apprently, I just had a lucky deploy that time.

joaquimserafim:
@fwang @lputnam2000 I'm a big fan of SST, by the way —it's made my life so much easier overall!  will this PR resolve the issue? Every time I'm using SST with RDS, I need to revert to an old version to connect through the tunnel — it's cumbersome and not the best practice. do you have more info on this tunnel issue and when it might be fixed?

jamesgibbons92:
@gkpln3 Small change needed to this PR before we can merge, the change you proposed breaks the tunnel when using managed nat. In this case, the IP has to be the bastion ec2 instance public ip address not the elastic ips - which are still created, as they are associated to the nat gateways.
Something like this should work I think:
```
              ip: natInstances.apply((instances) =>
                instances.length ? elasticIps[0]?.publicIp : bastion.publicIp,
              ),
```
If you're able to make this change I will push to get this merged

JohnCashmore:
> @gkpln3 Small change needed to this PR before we can merge, the change you proposed breaks the tunnel when using managed nat. In this case, the IP has to be the bastion ec2 instance public ip address not the elastic ips - which are still created, as they are associated to the nat gateways. Something like this should work I think:
> 
> ```
>               ip: natInstances.apply((instances) =>
>                 instances.length ? elasticIps[0]?.publicIp : bastion.publicIp,
>               ),
> ```
> 
> If you're able to make this change I will push to get this merged

@jamesgibbons92  I've opened a new pull request https://github.com/sst/sst/pull/6273 with this change and you're suggestion, they work well, sooner this goes in the better end up manually modifying the file every time i need to do a db migration etc.

Closed this now since the original author has made the change

gkpln3:
@jackwatters45 Done.

jamesgibbons92:
LGTM, tested the 3 combinations of ec2 nat, managed nat and no nat with the bastion and tunnel is working correctly:

2025-12-10-203845_hyprshot
2025-12-10-203705_hyprshot
2025-12-10-203229_hyprshot

jamesgibbons92 (approved):

gkpln3:
Lets merge this?

platform/src/components/aws/vpc.ts
```diff
@@ -680,14 +680,17 @@ export class Vpc extends Component implements Link.Linkable {
...
           self._publicSubnets,
         ]).apply(
-          ([bastion, privateKeyValue, privateSubnets, publicSubnets]) => {
+          ([bastion, elasticIps, privateKeyValue, privateSubnets, publicSubnets]) => {
             if (!bastion) return;
             return {
-              ip: bastion.publicIp,
+              ip: natInstances.apply((instances) =>
+                instances.length ? elasticIps[0]?.publicIp : bastion.publicIp,
+              ),
```
danielsharvey, line 693: 1. Should this be `self.instances` and included inthe `all()`? I think the static get case is affected.
2. I do not think this handles the case where NAT Eip's are specified (https://sst.dev/docs/component/aws/vpc/#nat-ip)
3. I suggest we add a comment to clarify the managed NAT case (https://github.com/anomalyco/sst/pull/5820#issuecomment-3593630357) - this is not clear from the code. Actually, more than that - why are the Elastic IPs created in the managed NAT case? And further, this will break the static get case.

GuilhermeCunha:
Heey, when will this new version be published on npm ?

How can i install it to test in the meantime ?

pauljasperdev:
> Heey, when will this new version be published on npm ?
> 
> How can i install it to test in the meantime ?

https://github.com/anomalyco/sst/issues/6215#issuecomment-3724410753

PR #5831, opened 2025-05-24
feat: add link and environment support to aws.Auth

derekbking:
## Summary
Adds link and environment support to sst.aws.Auth. For transparency, I haven’t worked in the SST codebase before, and this code was generated by Claude. I do believe there’s a real need to be able to pass links to the Auth component. Unless I'm overlooking a simple way to achieve this.

I haven’t tested these changes yet, but I plan to do so soon. I’m not currently set up to run SST locally and need to step away for now. This seems like a straightforward change, but since I’m not very familiar with SST, I’m sure there are some small decisions here that I might not have gotten quite right. If someone more familiar with the library is able to implement this properly, I’d really appreciate it.

## Use Case
Setting logo url based on my static site deployed to a CDN. I don't want to worry about hosting my logo somewhere when I already have the static site on a CDN.

**Before** (manual hardcoding / linking not allowed):
```ts
export const auth = new sst.aws.Auth("Auth", {
  issuer: "packages/auth/src/index.handler",
});
```

**After** (dynamic linking):
```ts
export const auth = new sst.aws.Auth("Auth", {
  issuer: "packages/auth/src/index.handler",
  link: [frontend] // Link frontend to access its URL
});
```

**In issuer function:**
```ts
const CUSTOM_THEME: Theme = {
  logo: `${Resource.Frontend.url}/logos/logo.svg`, // Dynamic URL
};
```

This allows me the reference the Static Site component which deploys my logo on a CDN already inside the theme config for the Auth component. I believe users could easily link buckets hosting their assets as well although a CDN seems preferable.

## Changes
- Added `link` parameter to `AuthArgs` interface with proper TypeScript documentation
- Added `environment` parameter to `AuthArgs` interface for custom environment variables  
- Updated issuer function creation to handle linked resources and custom environment
- Added comprehensive JSDoc examples showing asset hosting use case

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/platform/src/components/aws/auth.ts
+++ b/platform/src/components/aws/auth.ts
@@ -2,6 +2,8 @@ import {
   ComponentResourceOptions,
   jsonStringify,
   Output,
+  output,
+  all,
 } from "@pulumi/pulumi";
 import { Component } from "../component";
 import { Link } from "../link";
@@ -10,6 +12,7 @@ import { functionBuilder } from "./helpers/function-builder";
 import { env } from "../linkable";
 import { Auth as AuthV1 } from "./auth-v1";
 import { Input } from "../input";
+import { Permission } from "./permission";
 
 export interface AuthArgs {
   /**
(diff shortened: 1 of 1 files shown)


derekbking:
Somehow I overlooked this part of the docs https://sst.dev/docs/component/aws/auth#issuer. You can easily provide links like so:

{
  issuer: {
    handler: "src/auth.handler",
    link: [table]
  }
}

PR #5854, opened 2025-06-01
docs: update prisma docs with missing datasource

AaronFinn95:
Currently with the prisma documentation [here](https://sst.dev/docs/start/aws/prisma/) on the [Generate a migration](https://sst.dev/docs/start/aws/prisma/#generate-a-migration) step, it will fail with the following error:

```
aws-prisma npx sst shell --target Prisma -- npx prisma migrate dev --name init
Environment variables loaded from .env
Prisma schema loaded from schema.prisma

Error: A datasource block is missing in the Prisma schema file.
```

This is because as the error says:

> A datasource block is missing in the Prisma schema file

So this PR is to add the datasource in the [4. Create a schema](https://sst.dev/docs/start/aws/prisma/#4-create-a-schema) step


AaronFinn95:
Missed a step in the guide, removed unnecessary addition to docs

PR #5878, opened 2025-06-08
fix: aws.solidstart server actions

### LINKED ISSUE #5875: Solidstart server actions are broken
When using https://sst.dev/docs/component/aws/solid-start/
When you try to use any server function when navigating between routes (so it is client rendered) you get 403/405 status code . 
The issue seems to be the `.output/public/_server` folder which overrides the solid server function route.

comparing the solidstart implementation to the tanstack implementation:
https://github.com/sst/sst/blob/c4ed2ab092b241e35f347cd5d97adc9d0ef561fd/platform/src/components/aws/tan-stack-start.ts#L387
https://github.com/sst/sst/blob/c4ed2ab092b241e35f347cd5d97adc9d0ef561fd/platform/src/components/aws/solid-start.ts#L440

I'm not sure if this is considered a bug in sst or solidstart aws serverless streaming build output.
If it's a sst bug, I can submit a PR to remove the folder like tanstack does.

yinonburgansky:
Delete the `.output/public/_server` folder to prevent a conflicting route similar to the aws.tanstack implementation.
Fixes #5875

### DIFF
--- a/platform/src/components/aws/solid-start.ts
+++ b/platform/src/components/aws/solid-start.ts
@@ -438,6 +438,14 @@ export class SolidStart extends SsrSite {
       );
       const basepath = appConfig.match(/baseURL: ['"](.*)['"]/)?.[1];
 
+      // Remove the .output/public/_server directory from the assets
+      // b/c all `_server` requests should go to the server function. If this folder is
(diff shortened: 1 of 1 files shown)


jayair:
Thanks

PR #5880, opened 2025-06-10
fix: skip send to partial channel on diff

### LINKED ISSUE #5856: `sst diff` hanging
I'm running into an issue where `sst diff` is hanging on our production environment, this only seems to happen on this stage and it only seems to happen when I am adding some Lambda functions. 

The last line in in sst.log shows
`time=2025-06-02T10:15:38.912+01:00 level=INFO msg="pulumi exited" service=project.run err=<nil>`

Usually I'd see `eof and exited` and `breaking out of tail loop` for a successful diff. But these log events never occur. 

I've isolated the issue to this in `run.go`
```

        partial := make(chan int, 1000)
        ......
    	go func() {
    		if input.Command == "diff" {
    			return
    		}
    		for {
    			select {
    			case <-partialContext.Done():
    				partialDone <- nil
    				return
    			case <-partial:
    				workdir.PushPartial(update.ID)
    			case <-time.After(time.Second * 5):
    				workdir.PushPartial(update.ID)
    				continue
    			}
    		}
    	}()
        .......
	if event.ResOutputsEvent != nil || event.CancelEvent != nil || event.SummaryEvent != nil {
		partial <- 1
	}
```
this partial channel has a capacity of 1000, the buffer is filling up - but in diff mode, it's not being read from - so when it fills up the it blocks

jamesgibbons92:
Resolves https://github.com/sst/sst/issues/5856

partial channel was filling up during `sst diff` but this channel is not consumed in this mode due to this go routine:
```
	go func() {
		if input.Command == "diff" {
			return
		}
		for {
			select {
			case <-partialContext.Done():
				partialDone <- nil
				return
			case <-partial:
				workdir.PushPartial(update.ID)
			case <-time.After(time.Second * 5):
				workdir.PushPartial(update.ID)
				continue
			}
		}
	}()
```

 so once the buffer hit 1000 it started blocking and causes the command to hang indefinitely.

### DIFF
--- a/pkg/project/run.go
+++ b/pkg/project/run.go
@@ -505,7 +505,7 @@ loop:
 			}
 		}
 
-		if event.ResOutputsEvent != nil || event.CancelEvent != nil || event.SummaryEvent != nil {
+		if input.Command != "diff" && (event.ResOutputsEvent != nil || event.CancelEvent != nil || event.SummaryEvent != nil) {
 			partial <- 1
 		}
 


lkhari:
Any update on getting this merged @thdxr ? I have a few large projects that strugles with diff now.

PR #5894, opened 2025-06-13
Help with repackage

joegearlesss:
https://github.com/sst/sst/pull/5825


joegearlesss:
need to create [tests](https://github.com/sst/sst/pull/5893) before continue

joegearlesss:
will continue in few days, need fix agent tool first

joegearlesss:
golang is a dad way

PR #5896, opened 2025-06-14
Enhance TanStackStart error message for AWS-Lambda preset configuration

niklaswallerstedt:
- Added detailed error messages for missing AWS-Lambda preset in both Vite and Vinxi projects.
- Provided instructions for creating/updating `nitro.config.ts` and `app.config.ts` files to ensure proper configuration.

PR #5897, opened 2025-06-14
TanStack Start Vite example

niklaswallerstedt:
- adds `start-basic` template from TanStack Start based on vite
- fix naming in `aws-tanstack-start-alpha`

Ref:

- https://github.com/sst/sst/pull/5896
- https://github.com/TanStack/router/issues/4404#issuecomment-2971434717


jayair:
Thanks!

PR #5908, opened 2025-06-18
fix dynamodb subscription on existing tables

santiperone:


santiperone:
This fixes an issue where this.isStreamEnabled won't be set when referencing existing tables .

https://github.com/sst/sst/blob/8a82acf81b5ab681899b2168385b8c49aa443bc1/platform/src/components/aws/dynamo.ts#L621-L626

I guess another alternative would be to check against nodes.table.streamArn or nodes.table.streamEnabled in that line instead, lmk if you want me to change that.

santiperone:
Actually after a second look using "nodes.table.streamEnabled" just seems more reliable.

santiperone:
any news on this? @jayair @fwang

vimtor:
just to confirm: the issue only happens when you try to subscribe to a table after using `sst.aws.Dynamo.get`, right?

santiperone:
> just to confirm: the issue only happens when you try to subscribe to a table after using `sst.aws.Dynamo.get`, right?

Yes, that's right

vimtor (approved):

vimtor (changes requested):
the problem with the current implementation is that the exception is not thrown even if the stream is not enabled:

Image

this probably happens because the `this.nodes.table.streamEnabled` is an `Output<boolean>` and not a boolean and so it exists

santiperone:
Oh you're right, my bad. Thanks for catching that! This should be working now:
image
Great to finally see some movement here btw, thanks for taking this!

platform/src/components/aws/dynamo.ts
```diff
@@ -616,11 +614,12 @@ export class Dynamo extends Component implements Link.Linkable {
...
     // Validate stream is enabled
-    if (!this.isStreamEnabled)
-      throw new Error(
+    if (!this.nodes.table.streamEnabled.apply((streamEnabled) => {
+      if (!streamEnabled) throw new Error(
         `Cannot subscribe to "${sourceName}" because stream is not enabled.`,
       );
-
+    }))
+
```
vimtor: we should probably do something like this as we do in other parts of the codebase:


```ts
    return all([this.nodes.table.streamEnabled]).apply(([streamEnabled]) => {
      if (!streamEnabled) {
        throw new Error(`Cannot subscribe to "${sourceName}" because stream is not enabled.`);
      }
      
      return isFunctionSubscriber(...
    }))
    
```

vimtor (changes requested):
last comment @santiperone

thank you

santiperone:
@vimtor should be ready now

vimtor (approved):
approved!

<details>

<summary>proof</summary>

Image

Image

Image

</details>

PR #5921, opened 2025-06-23
Add `Nuxt` streaming support

RiFrost:
This pull request introduces the addition of the `streaming` property to NuxtServerArgs. The streaming option will be exposed directly through the server prop, rather than being accessed via transform. I achieved this by adding an Intersection Type (not sure if this approach suits you). I also added the functionality of properly handling intersection types to the `generate.ts`. For now, the intersection type will be rendered as an object with all the nested properties flattened and rendered. I have also added an example to demonstrate how to set up simple streaming with Nuxt and SST.

Open to discussion and changes!


jayair:
Thanks!

jayair:
This is pretty comprehensive!

vimtor (approved):
thanks for your contribution @RiFrost 

i made some minor changes to use the nitro config as we do elsewhere

<details>

https://github.com/user-attachments/assets/96e2d79d-f861-42c2-885c-56202bfdd28d

https://github.com/user-attachments/assets/238857ae-211b-4613-a816-33be8ec83767

</details>

PR #5952, opened 2025-07-08
Fix tasks not receiving host OS env vars on `sst dev`

sockthedev:
## Problem

Task dev mode fails with errors like `node: not found` because the system PATH is not available to the spawned process.

## Fix

- Simplify environment merging in task.go to follow the standard codebase pattern
- Use `latest` bridge-task image which blacklists PATH from being forwarded from the ECS container

🤖 Collaboration with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/cmd/sst/mosaic/aws/task.go
+++ b/cmd/sst/mosaic/aws/task.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"log/slog"
 	"os"
-	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -135,7 +134,7 @@ func task(ctx context.Context, input input) {
 				}
 				cmd := process.Command(fields[0], fields[1:]...)
 				cmd.Dir = task.Directory
-				cmd.Env = mergeEnvironment(body.Environment)
+				cmd.Env = append(os.Environ(), body.Environment...)
 				stdout, _ := cmd.StdoutPipe()
 				stderr, _ := cmd.StderrPipe()
 				go func() {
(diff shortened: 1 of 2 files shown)


sockthedev:
Need to verify this still.

...

verified. ✅

sockthedev:
@vimtor - i rounded back on this. greatly simplified the fix.

would appreciate if we could get this merged, as i still can't run `Task` locally without this patch.

cmd/sst/mosaic/aws/task.go
```diff
@@ -133,7 +134,7 @@ func task(ctx context.Context, input input) {
 				}
 				cmd := process.Command(fields[0], fields[1:]...)
 				cmd.Dir = task.Directory
-				cmd.Env = body.Environment
+				cmd.Env = append(os.Environ(), body.Environment...)
```
sockthedev, line 137: This makes sure we pass the environment variables from the host OS, alongside the environment variables from the Task definition, when running the Task locally.

Super key for making sure things like `PATH` etc are available for local dev.

platform/src/components/aws/task.ts
```diff
@@ -326,9 +326,7 @@ export class Task extends Component implements Link.Linkable {
             return [
               {
                 ...v[0],
-                image: output(
-                  "ghcr.io/anomalyco/sst/bridge-task:20241224005724",
-                ),
+                image: output("ghcr.io/anomalyco/sst/bridge-task:latest"),
```
sockthedev, line 329: The bridge-task was not pulling from latest builds, which I assume we should always do, or at least keep updating this to be inline with the published version closest to the respective SST CLI version?

sockthedev: If the later approach makes more sense, then it should be [20250624172413](https://github.com/orgs/anomalyco/packages/container/sst%2Fbridge-task/446034433?tag=20250624172413)

Needs a checklist though for updating this version number with every SST publish to ensure the bridge code is always aligned.

vimtor:
i finally got the issue: to reproduce it you need to set the `latest` version of the `bridge-task`, which blacklisted the PATH env var and made go stop inheriting system vars.

problem is the latest image hasn't been pushed in 8 months. just saw the ci shows a silent error:

```
ERROR: failed to build: failed to solve: failed to push ghcr.io/anomalyco/sst/bridge-task:20260223101712: denied: permission_denied: write_package
```

https://github.com/anomalyco/sst/actions/runs/22301779985/job/64511131698#step:11:235

i'll fix ci first, then review the pr

sockthedev:
you legend @vimtor !

i'm mega looking forward to not having to run my own fork of sst.

this component is so powerful, amazing to be able to reach for this capability so easily. dx just needs some love.

sockthedev:
another dx experience downer is the truncation of logs in the Task output tab of the sst dev server. i'll see how this rollout goes and then raise another issue/pr to address that.

vimtor (approved):
thanks for your contribution @sockthedev 

<details>

Image

</details>

PR #5980, opened 2025-07-16
fix: support S3 to SQS notifications with KMS-encrypted queues by updating queue policy

pbalenzuela:
## Problem

When using `bucket.notify(...)` to create an S3 → SQS notification, SST automatically sets a queue policy allowing s3.amazonaws.com to send messages. However, this deployment silently fails if the SQS queue is encrypted with an AWS-managed KMS key (e.g., `alias/aws/sqs`).

S3 requires the SQS queue policy to include specific conditions to validate the destination when encryption is enabled. Without these conditions, the `PutBucketNotificationConfiguration` API returns:

```sh
InvalidArgument: Unable to validate the following destination configurations
```

## Why this happens only with KMS?

When a queue uses an AWS-managed key, S3 must call `kms:GenerateDataKey` on the queue’s KMS key. AWS only allows this if the queue policy includes:

```json
"Condition": {
  "ArnEquals": {
    "aws:SourceArn": "<bucket-arn>"
  }
}
```

Notice that in the approach I'm also restricting to the current aws account. This is a good security measure considering that the bucket arns are global

## The Fix

This PR updates bucket.notify(...) to automatically add the required `aws:SourceArn` condition to the SQS queue policy.

Adding these conditions unconditionally is safe because:
	•	They only restrict permissions for `s3.amazonaws.com`, which is a best practice for security.
	•	They do not interfere with other services like SNS or EventBridge.
	•	S3 requires these conditions for validating notifications but does not fail if they are present when not strictly needed.

### DIFF
--- a/platform/src/components/aws/bucket-notification.ts
+++ b/platform/src/components/aws/bucket-notification.ts
@@ -159,7 +159,7 @@ export class BucketNotification extends Component {
             const policy = Queue.createPolicy(
               `${name}Notification${n.name}Policy`,
               arn,
-              { parent: self },
+              { parent: self, sourceArns: [bucket.arn] },
             );
             return { args: n, queueArn: arn, dependsOn: policy };
           }
--- a/platform/src/components/aws/queue.ts
+++ b/platform/src/components/aws/queue.ts
@@ -13,7 +13,7 @@ import { VisibleError } from "../error";
 import { hashStringToPrettyString, logicalName } from "../naming";
 import { parseQueueArn } from "./helpers/arn";
 import { QueueLambdaSubscriber } from "./queue-lambda-subscriber";
-import { iam, lambda, sqs } from "@pulumi/aws";
+import { iam, lambda, sqs, getCallerIdentity } from "@pulumi/aws";
 import { DurationHours, DurationMinutes, toSeconds } from "../duration";
 import { permission } from "./permission.js";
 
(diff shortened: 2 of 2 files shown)


vimtor:
thanks for the contribution @pbalenzuela

could you provide a sample `sst.config.ts` for reproducing the issue?

pbalenzuela:
> thanks for the contribution @pbalenzuela
> 
> could you provide a sample `sst.config.ts` for reproducing the issue?

It seems it is working now. I tried the following and it works:

```ts

export default $config({
  app(input) {
    return {
      name: 'downloads',
      removal: input?.stage === 'production' ? 'retain' : 'remove',
      protect: ['production'].includes(input?.stage),
      home: 'aws',
      providers: {
        aws: {
          region: 'us-west-2',
          profile: 'admin',
        },
      },
    };
  },
  async run() {
    const bucket = new sst.aws.Bucket('test-bucket');
    const queue = new sst.aws.Queue('test-queue', {
      transform: { queue: { kmsMasterKeyId: 'alias/aws/sqs' } },
    });
    bucket.notify({
      notifications: [
        {
          name: 'MySubscriber',
          queue,
        },
      ],
    });
  },
});

```

pbalenzuela:
> thanks for the contribution @pbalenzuela
> 
> could you provide a sample `sst.config.ts` for reproducing the issue?

In my project I ended up not using `bucket.notify`, I managed the `aws.s3.BucketNotification` myself to avoid the problem.
The problem might still be there but I would not know how to reproduce it anymore since I followed a different approach

PR #6004, opened 2025-07-26
Replace latest tag with component name in python docker Image

doron98:
All newly pushed Python Docker images to ECR currently use the `latest` tag. This causes issues because it makes it harder to identify images in ECR and to implement simple lifecycle policies, such as keeping only the most recent image for each type. As a result, images accumulate over time, leading to non-negligible storage costs.

For the Service and Task components, the default tag is the component name and can be customized. Ideally, we should handle Python Docker images similarly. But as a quick fix, I propose changing the tag of newly pushed Python Docker images to use the component name instead of `latest`.


vimtor:
related to #6359

PR #6014, opened 2025-07-30
Fix bridge task alpine platform mismatch

### LINKED ISSUE #6013: Bridge Task - Arm architecture issue when running SST Dev on Arm architecture Tasks
Found a bug in the bridge-task where the alpine image that is used for the final image is the wrong architecture when building for arm tasks.

### The Problem
The bridge-task that runs as the bridge between my local machine and the cloud in `sst dev` doesn't run properly when my SST Task is configured to use the arm architecture.

### What does this affect?
Whenever your SST Task is configured to use arm architecture, the bridge software that gets put in place of your task (in AWS, for instance) crashes. Things like environment variable injection no longer work as well as any dev commands stop working.

### The Fix?
The problem, from what I can tell, comes from a mismatch of the architecture between the built Go project and the alpine distro used to run the executable.  

https://github.com/sst/sst/blob/dev/platform/support/bridge-task/Dockerfile

Taking a look at the Dockerfile that builds the the bridge task, the builder stage properly switches what architecture to build the bridge task with (this is also in part thanks to @thdxr who initially took a look at this issue from my questions on Discord 🙏🏻). 

Where I believe we are still crashing is that the second stage is using the base Alpine distro which uses amd64 architecture. We would need to use arm64 architecture for the runner so that the built Go project doesn't have conflicting architecture.

The arm64 version of the alpine distro is found at: https://hub.docker.com/r/arm64v8/alpine/

I recommend that we expand the Dockerfile to include switching which architecture to use for running the Bridge-Task Go project. I will create a PR and link it here in a bit with my proposed fix.

### Example Task
I wanted to throw a simple code example here to illustrate what it takes to see this issue.

```ts
new sst.aws.Task(
	"DummyTask",
	{
		cluster,
		architecture: "arm64",
		image: {
			context: "packages/tasks/src/dummy",
			dockerfile: "Dockerfile",
		},
		dev: {
			directory: "packages/tasks/src/dummy",
			command: `bun index.ts`,
		},
	}
);
```

Here, we are just creating a task that used arm64 as the architecture. When running in dev mode, my bun command never runs because the bridge task fails to start in the cloud. When inspecting the task logs in AWS ECS, I see `[FATAL tini (7)] exec /bin/sh failed: Exec format error`. In other words, the architecture is mismatching between the compiled go binary and the alpine image running it.

Thanks for reading the writeup and I look forward to getting this fixed up so I can use Arm Tasks in dev mode 🙏🏼.

ChristianPayne:
Closes https://github.com/anomalyco/sst/issues/6013

The changes here are around the Dockerfile that builds and runs the Bridge Task Go project.
The different architectures that can be built are passed into the Dockerfile arg from the build script's --platform flag. 
The change in particular is to the runtime and which one is chosen depending on the architecture being used.

If there is a better way to go about this or errors in my logic, I would love to make changes to get this resolved.

Thank you!


vimtor:
closes #6013

blocked by #6308

vimtor:
@ChristianPayne are you still running into this issue?

this seems strangey because `alpine` is already multiarch so this change should be needed in principle

i can't test since i have arm, did this change work for you?

ChristianPayne:
To be honest, my company switched away from SST for this use case because we couldn't get this resolved. 

I do find it strange that you can't test it. The computer that it would be erroring on is in the cloud. You can configure an arm server via sst so anyone can do it. @vimtor

ChristianPayne:
I do see what you mean with multiarch. I actually didn't know Docker had that capability. This would change the solution a bit for sure, in fact, I think it would simplify the solution.

ChristianPayne:
So this change reverts a lot of the hard coded image selection I had proposed initially. Now we can just pass the target arch to docker and multiarch should pick the right one.

I tested multiarch platform selecting in a dummy Dockerfile and it worked. I have not tested this in the full build for SST's Bridge.

Thoughts?

vimtor:
@ChristianPayne i'm testing locally and i believe `FROM alpine` already builds a multi-arch build

in fact, support for it was added in https://github.com/anomalyco/sst/commit/f2535915fd

do you get a different behaviour with this pr code vs the existing one?

ChristianPayne:
@vimtor For one, I appreciate you digging into this. 

I think my original problem was that the Go project was built using the TARGETARCH variable. So that would for sure build with ARM if that is the arch we target. 

Where my original problem's emphasis was on was that just saying `FROM alpine`, doesn't give any info to Docker to make sure it grabs the ARM version. I believe by default it is x86. By adding the platform flag, we can hint to grab the arm version.

That being said, I don't want to discredit your effort on locally testing. I will see if I can get a project scaffold going to isolate as best I can. I'll try main branch first to see if I can replicate then try the PR if I still get the issue.

ChristianPayne:
**TLDR:** This PR isn't needed. The bug I was hitting was caused by v3.17.x pinning to an older amd64-only image. The fix already exists at the CI build pipeline level, and v4.x works correctly.

---

@vimtor First off, thank you for pushing back and testing on your end. It got me to actually trace through the whole thing and I learned a lot about how the build pipeline works.

Here's what I found and the journey I took to get there:

**The setup:**
I created a bare SST app with a single arm64 Task with a console.log, nothing else.

My work project is on v3.17.8, so I started there. That version still referenced the old `ghcr.io/sst/sst/...` URL (pre org rename), so the image pull 403'd. Bumped to v3.17.34 which has the `anomalyco` URL fix — image pulled fine, but the task immediately crashed with `[FATAL tini (6)] exec /bin/sh failed: Exec format error`. The exact bug I originally reported.

Then I bumped to v4.2.7 and... it just worked. No error. So something changed between v3.17.34 and v4.

**Tracing the fix:**
I dug into why. The difference is that v3.17.34 pins to a specific timestamped image tag — `ghcr.io/anomalyco/sst/bridge-task:20241224005724`. That image was built in December 2024, before any multi-arch support existed. It's amd64-only.

v4.x references `:latest`. I inspected that manifest:
```
docker manifest inspect ghcr.io/anomalyco/sst/bridge-task:latest
```
It has both `amd64` and `arm64` variants. Docker resolves the right one at pull time based on the ECS task's configured architecture.

**Where the multi-arch comes from:**
You pointed me to commit `f2535915fd` — that's the one. It added `docker buildx --platform linux/amd64,linux/arm64` to the CI build script (`platform/scripts/build`). That flag is what produces the multi-arch manifest when the image is built in CI. So `FROM alpine` in the Dockerfile doesn't need `--platform` because `buildx` is already handling it externally.

**What I got wrong:**
My original assumption was that `FROM alpine` would default to x86_64 and we needed to explicitly pass the platform in the Dockerfile. That's true if you're building with a plain `docker build`, but not when the CI uses `docker buildx --platform` — buildx overrides the base image platform for each target arch. So the Dockerfile is fine as-is.

**Bottom line:**
The bug I was experiencing was real, but it was specific to v3.17.x pinning to a pre-multi-arch image. The fix at the build pipeline level already solved it for any version that references `:latest`. This Dockerfile change would be redundant.

**For anyone else landing here with the same `exec format error` on arm64 tasks in dev mode:** the fix is to upgrade to v4.x. The multi-arch bridge-task image is already published and `:latest` just works.

The only thing this PR would protect against is if someone needed to build the bridge-task Dockerfile directly outside of the CI pipeline. If that's not a concern, I'm good with closing this out. Does all of that track with what you're seeing on your end?

Thanks again for the time on this.

vimtor:
great! thanks for letting me know

PR #6016, opened 2025-07-31
Fix #6011: Inconsistent behavior of dev processes in `sst dev`

### LINKED ISSUE #6011: Inconsistent startup of `dev` resources (`sst.aws.StaticSite`, `sst.aws.Service`, `sst.x.DevCommand`...)
When `sst dev` is ran process startup is inconsistent - all processes do not always appear in the multiplexer in the desired state. This behavior has been observed across team members working in our project. 

Without sharing our repo directly - we're working on a fairly standard web app. Have a `StaticSite` along with a typical ECS `Cluster` + `Service` backend and `Postgres` for storage. Everything lives in a VPC, using private VPC links and APIGV2 for public access to backend resources. A couple of loose Cognito, SES, and Lambda resources - but generally nothing out of the ordinary. Otherwise - we have 5-10 `sst.x.DevCommand` instances to surface various testing and development scripts in local development. Brief summary - happy to share more details if it's helpful - just wanted to get an issue out. 

My best guess as to why it's happening is that there's some sort of race condition we're seeing around sst server startup - the `CompleteEvents` that the command looks for seem to never be received... but I'm sure I'm missing the nuance of how it all fits together. I think that I have an answer - but might be misguided. 

Lifted the code that lives in the `EventProcess` case of the event switch/case in `multiplexer.go` into a function that can be called ahead of the subscription to these events in `mosaic.go` (all in `multi.Start()`).  We can then walk through the output of `p.GetCompleted(c.Context)` and for any completed processes - add them in the exact same way processes are added in the `CompleteEvent` case of the switch/case defined after the subscription. 

I'll make a PR shortly and link to this issue - I'm sure my 6AM plain text description isn't super helpful. I'm not convinced that this approach handles the initial deploy case appropriately - ie with an `sst remove` and `sst dev` call certain resources never come up - however after a second run all appear as expected.

ryan-warner-foreflight:
A possible fix for the behavior laid out in #6011 

## Description  
Fixes startup inconsistency in the dev command by ensuring that already completed events are processed immediately when starting sst, rather than waiting for new events to arrive.

## Changes:  
- **Refactored event processing logic**: Extracted common event processing functions (`processCompleteEventForMultiplexer` and `processCompleteEventForMonoplexer`) to eliminate code duplication between events processed before server start and after
- **Added immediate startup processing**: Both multiplexer and monoplexer now check for and process any already completed events (`p.GetCompleted()`) before subscribing to new events
- **Added InitProcess**: Moved process initialization logic from the event handler into a dedicated `InitProcess` method for better separation of concerns

## Considerations:  
- I'll attach some additional videos to this when I get a chance to compare current behavior and behavior with these fixes. As mentioned in the issue - not 100% certain this is THE fix - but it certainly solves most of what we're seeing today. I am not convinced this provides the correct experience on the first `sst dev` run for a stage.

### DIFF
--- a/cmd/sst/mosaic.go
+++ b/cmd/sst/mosaic.go
@@ -247,12 +247,44 @@ func CmdMosaic(c *cli.Cli) error {
 		defer func() {
 			multi.Exit()
 		}()
+
+		processCompleteEventForMultiplexer := func(evt *project.CompleteEvent, addProcess func(key string, args []string, icon string, title string, cwd string, killable bool, autostart bool, env ...string)) {
+			for _, d := range evt.Devs {
+				if d.Command == "" {
+					continue
+				}
+				dir := filepath.Join(cwd, d.Directory)
+				title := d.Title
+				if title == "" {
+					title = d.Name
+				}
+				addProcess(d.Name, append([]string{currentExecutable, "dev"}), "→", title, dir, true, d.Autostart, append([]string{"SST_CHILD=" + d.Name}, multiEnv...)...)
+			}
+
+			for name := range evt.Tunnels {
+				addProcess("tunnel", []string{currentExecutable, "tunnel", "--stage", p.App().Stage}, "⇌", "Tunnel", "", true, true, append(
+					multiEnv,
+					"SST_LOG="+p.PathLog("tunnel_"+name),
+				)...)
+			}
+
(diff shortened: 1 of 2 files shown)


vimtor:
thanks for your contribution

closed in favour of https://github.com/anomalyco/sst/pull/6402

PR #6029, opened 2025-08-07
Replace EventBus with Bus

nepeanwjdw:
In the migrate from v2 guide it specifies to use the EventBus construct but it actually should be Bus instead


jayair:
Thanks!

PR #6032, opened 2025-08-08
Fix Cloudflare zone ID resolution

### LINKED ISSUE #5853: SST CDN Cloudflare SSL, CAA failing if subdomain is used
Using SST 3.17.0 and Cloudflare 6.2.0

This code is supposed to create cloudfront accessible with subdomainXXX.MAIN-DOMAIN.com
```
export const router = new sst.aws.Router("router", {
  // Configure the domain for the router with Cloudflare DNS management
  ...(process.env.API_SUBDOMAIN && process.env.API_DOMAIN_NAME && process.env.CLOUDFLARE_ZONE_ID && {
    domain: {
      name: `${process.env.API_SUBDOMAIN}.${process.env.API_DOMAIN_NAME}`,
      // aliases: [],
      dns: sst.cloudflare.dns({
        zone: process.env.CLOUDFLARE_ZONE_ID,
        // proxy: false, // whether to drive traffic through Cloudflare or not (false = direct to AWS)
      })
    }
  }),
```

**But we always end up with:**
```
|  Created     router sst:aws:Router → routerCdnSslCertificate aws:acm:Certificate (8.0s)
|  Error       router sst:aws:Router → routerCdnSslCAA**subdomainXXX main-domain:com**Record:sst:cloudflare:DnsRecord
Cloudflare API error: Could not route to /client/v4/zones/dns_records, perhaps your object identifier is invalid?
|  Error       router sst:aws:Router → routerCdnSslCAAWildcard**subdomainXXX main-domain:com**Record:sst:cloudflare:DnsRecord
Cloudflare API error: Could not route to /client/v4/zones/dns_records, perhaps your object identifier is invalid?
```

Notice the " " between subdomain and domain while creating cloudflare resources.

vimtor:
Fixes #5853

### DIFF
--- a/platform/src/components/cloudflare/dns.ts
+++ b/platform/src/components/cloudflare/dns.ts
@@ -110,7 +110,7 @@ export function dns(args: DnsArgs = {}) {
         zoneId: args.zone,
       });
       return {
-        id: zone.id,
+        id: zone.zoneId,
         name: zone.name,
       };
     }


vimtor:
Just in case, I don't believe this should be made retrocompatible since the latest SST version requires Cloudflare Pulumi version >6 and it changed the result of zone from `zone.id` to `zone.zoneId`: https://www.pulumi.com/registry/packages/cloudflare/api-docs/getzone/#zoneid_nodejs

PR #6047, opened 2025-08-16
Fix Cloudflare DNS TypeScript error

vimtor:
my latest merged PR #6032 gave a TS error after merging

the Cloudflare Pulumi SDK seems to be typed incorrectly when compared with their docs

this time i double checked by running `tsc` locally

### DIFF
--- a/platform/src/components/cloudflare/dns.ts
+++ b/platform/src/components/cloudflare/dns.ts
@@ -110,7 +110,7 @@ export function dns(args: DnsArgs = {}) {
         zoneId: args.zone,
       });
       return {
-        id: zone.zoneId,
+        id: args.zone,
         name: zone.name,
       };
     }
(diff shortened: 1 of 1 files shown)


platform/src/components/cloudflare/dns.ts
```diff
@@ -110,7 +110,7 @@ export function dns(args: DnsArgs = {}) {
         zoneId: args.zone,
       });
       return {
-        id: zone.zoneId,
+        id: output(args.zone),
```
vimtor, line 113: we could also do something like this:

```ts
        id: all([zone.zoneId, args.zone]).apply(([resultId, defaultId]) => {
          return resultId ?? defaultId;
        }),
```

but seemed a bit redundant since we already know the zone ID

PR #6061, opened 2025-08-23
Improve Python support

### LINKED ISSUE #5190: Can't call an API route from another route with auth iam enable in local
In Local : I created an api gateway v2 and 2 routes one in typescript and one in python. The second route is protected with auth with IAM.
When the first route is invoked, it calls the second route with the sigv4 and the call fails because of an 403 error forbidden.

I use @aws-sdk/credential-provider-node, from the documentation : it will attempt to find credentials from the following sources (listed in order of precedence):

- Environment variables exposed via process.env
- SSO credentials from token cache
- Web identity token credentials
- Shared credentials and config ini files
- The EC2/ECS Instance Metadata Service

In my code , I use this block of code

```typescript
import { defaultProvider } from "@aws-sdk/credential-provider-node";

console.log("ENV", process.env);

const apiUrl = `${process.env.API_URL}/test2`;

const signer = new SignatureV4({
  credentials: defaultProvider(), //NOT WORKING
  // credentials: defaultProvider({profile: "default"}),
  sha256: Sha256,
  region: "eu-west-1",
  service: "execute-api",
});
```

By digging and displaying the signed request and also the environement variable, it seems that the calls use an access key id of type ASIA (Temporary Access Keys) that seems to be coming from the the environment variable of the lambda.
If I force the profile in the default provider, it works 

I'm not sure if it's a bug : shouldn't the access key be populated with your access key id in local ?
I tried also to specify the profile in the sst.config.ts but it seems the access key id in the environment variable is still an temporary one 

I reproduced on a small example with 2 typescript lambda :

[example](https://github.com/TheUncharted/test_lambda)

Currently, I invoke directly the lambda and its working so it's an urgent issue

### LINKED ISSUE #5704: Incorrect Handler Path When Using Workspace Python Functions
When deploying a Python Lambda function within a workspace (using `tool.uv.workspace`), the generated handler path includes the workspace directory (`python/`), which doesn't exist in the deployed package. This results in a broken Lambda due to an invalid handler path.

**Steps to Reproduce:**  
1. Create a Python SST app with a workspace structure:
   ```toml
   # pyproject.toml (root)
   [tool.uv.workspace]
   members = ["python/*"]
   ```
2. Define a Lambda function in a workspace member (`python/lambda1`):
   ```js
   const lambda1 = new sst.aws.Function('Lambda1', {
     runtime: 'python3.12',
     handler: 'python/lambda1/handler.handler'
   });
   ```
3. Deploy the app. The Lambda's handler path becomes `python/lambda1/handler.handler`, but the deployed code only includes `lambda1/handler.py`.

Image

**Expected Behavior:**  
The handler path should be `lambda1/handler.handler` (without the `python/` prefix) to match the deployed code structure.

### LINKED ISSUE #5821: SST Lambda Python Deployment Not Installing Dependencies
I'm trying to deploy a Lambda function that forwards Cloudwatch Logs to Datadog. [link](https://docs.datadoghq.com/logs/guide/forwarder/?tab=cloudformation) 

By default this service uses requirements.txt however I know SST uses uv. I created a pyproject.toml for the lambda and am deploying it that way. However in SST console I get an error 

```[ERROR] Runtime.ImportModuleError: Unable to import module 'lambda_function': No module named 'requests'
Traceback (most recent call last):
INIT_REPORT Init Duration: 229.24 ms	Phase: invoke	Status: error	Error Type: Runtime.ImportModuleError
```

This is odd because requests is defined in my pyproject.toml 

```toml
[project]
name = "datadog-forwarder-lambda"
version = "0.1.0"
description = "Datadog Forwarder Lambda"
requires-python = "==3.12.*" # Match your Lambda runtime
dependencies = [
  "attrs",
  "bytecode",
  "cattrs",
  "certifi",
  "charset-normalizer",
  "datadog-lambda==6.104.0",
  "datadog==0.50.2",    
  "ddsketch==3.0.1",
  "ddtrace==2.17.3",
  "deprecated",
  "envier",
  "exceptiongroup",
  "idna==3.7",
  "importlib-metadata",
  "opentelemetry-api",
  "protobuf",
  "requests-futures",
  "requests",
  "six",
  "typing-extensions",
  "urllib3>=1.26.19,<3.0",
  "wrapt==1.14.0",
  "xmltodict",
  "zipp",
  "ujson"
]

[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"
backend-path = ["."] # Tells build system to look in current directory.


```

Here are the imports for the lamba for reference 

```python
# Copyright 2021 Datadog, Inc.

import json
import os
import boto3
import logging
import requests
from hashlib import sha1

from datadog_lambda.wrapper import datadog_lambda_wrapper
from datadog import api
from enhanced_lambda_metrics import parse_and_submit_enhanced_metrics
from steps.parsing import parse
from steps.enrichment import enrich
from steps.transformation import transform
from steps.splitting import split
from caching.cache_layer import CacheLayer
from forwarder import Forwarder
from settings import (
```

Here is my sst function file for the lamba function. 

```ts
import { DatadogApiKey } from "../aws-services/secrets";

const forwardHelperBucket = new sst.aws.Bucket("DatadogForwarderHandler", {});

export function DatadogForwarder() {
  const lambda = new sst.aws.Function("DatadogForwarder", {
    bundle: "functions/datadog-forwarder/logs_monitoring",
    environment: {
      DD_API_KEY: DatadogApiKey.value,
      DD_FETCH_LAMBDA_TAGS: "true",
      DD_FETCH_LOG_GROUP_TAGS: "true",
      DD_LOG_LEVEL: "info",
      DD_S3_BUCKET_NAME: forwardHelperBucket.name,
      DD_SITE: "datadoghq.com",
      DD_STORE_FAILED_EVENTS: "true",
    },
    handler: "lambda_function.lambda_handler",
    link: [DatadogApiKey, forwardHelperBucket],
    memory: "512 MB",
    permissions: [
      // Permissions for the DD_S3_BUCKET_NAME
      {
        actions: [
          "s3:GetObject",
          "s3:PutObject",
          "s3:ListBucket",
          "s3:DeleteObject",
        ],
        resources: [
          forwardHelperBucket.arn,
          $interpolate`${forwardHelperBucket.arn}/*`,
        ],
      },
      // Permissions for fetching tags
      {
        actions: ["tag:GetResources"], // For DD_FETCH_LAMBDA_TAGS
        resources: ["*"], // Scope down if possible
      },
      {
        actions: ["logs:ListTagsLogGroup"], // For DD_FETCH_LOG_GROUP_TAGS (Datadog docs also mention logs:ListTagsForResource as an alternative)
        resources: ["*"], // Scope down if possible
      },
    ],
    runtime: "python3.12",
    timeout: "300 seconds",
  });

  return {
    datadogForwarder: lambda,
  };
}
```

The docs aren't very detailed on how things are installed when using Python and SST. Am I doing something wrong?

### LINKED ISSUE #5932: Task dev command execution environment issues
sst version: 3.17.6

I'm having some issues running the dev command for the Task resource. They appear to related to the execution environment in which the [dev](https://sst.dev/docs/component/aws/task/#dev) command is run.  Here is the snippet outlining what I would ideally like to achieve:

```typescript
const task = new sst.aws.Task("MyTask", { 
  cluster,
  image: {
    context: "services/taskRepo",
    dockerfile: "Dockerfile",
  },
  dev: {
    // suboptimal - we should be able to hit the cache and use Nx
    command: "uv run --no-cache python -m task.main",
    // command: "pnpm nx run task:serve", // doesn't work with error "env: node: No such file or directory"
    // command: "uv run python -m task.main", // doesn't work with with permissions error to write to uv cache
  },
});
```

In the end a workaround is possible (run the Python task directly without cache) but naturally I would prefer to either:

1. Run the Python task, using uv, with cache enabled
2. Run the task using Nx for consistency and to take advantage of their caching mechanism

Many thanks for taking a look - running tasks like this in with a Live dev mode is an incredible feature!

### LINKED ISSUE #6050: Python Lambda dev command errors when packages aren't directly inside root directory
`sst dev` fails to deploy lambdas with following file structure:

Image

Error:
```bash
|  Error       packages/embedding/src/embedding/embed.handler
|  No module named 'embed'
|  ↳ Traceback (most recent call last):
|  ↳ File "/Users/slmn/Work/Personal/pychat/.sst/artifacts/filesNotificationsNotificationembedder-dev/lambdaric_python_bridge.py", line 50, in <module>
|  ↳ module = importlib.import_module(module_name)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/opt/homebrew/Cellar/python@3.12/3.12.11/Frameworks/Python.framework/Versions/3.12/lib/python3.12/importlib/__init__.py", line 90, in import_module
|  ↳ return _bootstrap._gcd_import(name[level:], package, level)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
|  ↳ File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
|  ↳ File "<frozen importlib._bootstrap>", line 1324, in _find_and_load_unlocked
|  ↳
```

It's happening because we are adding wrong path to `sys.path`:
https://github.com/sst/sst/blob/dev/platform/functions/python-runtime/index.py#L46

we are supposed to add:
`~/project/.sst/artifacts/filesNotificationsNotificationembedder-dev/embedding/`
but sst adds this instead:
`~/project/.sst/artifacts/filesNotificationsNotificationembedder-dev/packages/embedding/`

Image

### LINKED ISSUE #6054: AWS Lambda Python Runtime ImportModuleError
When deploying the Python Lambda function using the [aws-python SST example](https://github.com/sst/sst/tree/dev/examples/aws-python) the function fails at runtime with a GLIBC version compatibility error related to the cryptography package. This error is persistent in all other python functions I tried to deploy that imports the `SST` module (`from sst import Resource`)

Image

The Lambda runtime environment appears to have an older version of GLIBC that doesn't include the symbols required by the cryptography package - `AWS Python 3.11`

subssn21:
This PR implements comprehensive fixes for Python Lambda deployment artifacts, addressing critical issues with Project Layout, missing modules, oversized packages, and deployment reliability. Also adds Python 3.13 support.

## Issues Fixed
- Fixes #6054 - AWS Lambda Python Runtime ImportModuleError
- Fixes #6050 - Python Lambda dev command errors when packages aren't directly inside root directory
- Fixes #5821 - SST Lambda Python Deployment Not Installing Dependencies
- Fixes #5704 - Incorrect Handler Path When Using Workspace Python Functions
- Closes #5932 - Task dev command execution environment issues
- Closes #5190 - Python: `.venv` in function handler dir conflicting with Docker `.venv`
- Closes #5126 - Python 3.13 runtime support implemented

## Key Improvements
- ✅ Comprehensive build validation and error handling
- ✅ Intelligent content filtering to prevent oversized packages
- ✅ Support for flexible Python project layouts
- ✅ Enhanced caching and performance optimizations
- ✅ Python 3.13 runtime support
- ✅ Extensive testing and documentation

## Testing
- Comprehensive test suite with integration tests
- Validated with real Lambda deployments
- Backward compatibility maintained

*This Pull request and code was created with the help of an AI agent*


dsonet:
@thdxr Could you please take a look at this PR? It's a significant improvement for python deployment, and it would be great to include it in the new version.

subssn21:
I know it looks big, but of the 117 files changed only 23 are actual system code, the rest is documentation, examples and tests.

syhner:
@walln in case you'd like to take a look

moAlobaidi:
@thdxr @jayair can we get this across the finish line happy to help

subssn21:
I have updated the code to fix some layout issues and caching issues I found when I started testing against real projects instead of just the example projects. I am now running a functions/src/functions style project as well as a large monorepo off of the new Python code, so it should be even better tested now

dan-hook:
Are you sure this closes [https://github.com/sst/sst/issues/5704](https://github.com/sst/sst/issues/5704)?  I tried this fork on my project and it seems like its an issue.

subssn21:
> Are you sure this closes #5704? I tried this fork on my project and it seems like its an issue.

Any chance you can send me a simple example project that reproduces the issue? The workspace example at examples/python-layouts/workspace-layout was supposed to test this exact issue. Is there something wrong with its setup?

dan-hook:
> Any chance you can send me a simple example project that reproduces the issue? The workspace example at examples/python-layouts/workspace-layout was supposed to test this exact issue. Is there something wrong with its setup?

Ah, I was following the aws-python example with pyproject.toml files at two different levels, but using a uv workspace.  Following the workspace-layout worked.

subssn21:
I have been using the new code successfully with two actual SST projects for a while now and have tweaked the code to make it work even better for Python. There are probably still a few minor issues, but all in all it works well.

subssn21:
This now has a much dev live experience for Python users. When using a standard workspace setup (Not the legacy function/src/function) We now don't do anything in the build step at all making it really fast. We just point the Bridge to the handler in your source and execute it like you would any other Python code. We then just have to keep track of when to restart the Python processes before an invoke so that you get the changed code.

Baalakay:
+1 for this fix. Been cussing at my Cursor agent for weeks thinking it's an invalid config somewhere. Can't wait for this fix to be merged. Thx

dan-hook:
I've been using this branch for several weeks now and it's been working very well for me.

pkg/runtime/python/content_filter.go
```diff
@@ -0,0 +1,542 @@
...
+		"docker-compose.yml",
+		"docker-compose.yaml",
+
+		// Test directories and files
+		"tests",
+		"tests/**",
+		"test",
+		"test/**",
+		"*_test.py",
+		"test_*.py",
```
dan-hook: Excluding the directories is a good idea, but the module exclusions led to some rather mysterious errors when my module `test_user.py` wasn't included.  I think the module exclusions should be removed, or if not removed then they should be properly documented in python-best-practices.md.

docs/python-best-practices.md
```diff
@@ -0,0 +1,606 @@
...
+- Package structure files (`__init__.py`)
+- Static files and assets in common directories (`data/`, `config/`, `templates/`, `static/`, `assets/`)
+- Essential configuration files (`requirements.txt`)
+
+**❌ Automatically Excluded:**
+- **Build artifacts**: `__pycache__/`, `*.pyc`, `*.pyo`, `*.pyd`, `*.egg-info/`
+- **Version control**: `.git/`, `.gitignore`, `.gitattributes`
+- **Virtual environments**: `.venv/`, `venv/`, `env/`
+- **IDE files**: `.vscode/`, `.idea/`, `*.swp`, `.DS_Store`
+- **Test artifacts**: `.pytest_cache/`, `.coverage`, `htmlcov/`
```
dan-hook: As the code is now, this should also include:
`test_*.py`, `*_test.py`
But I'm hoping you'll remove those exclusions.  

```suggestion
- **Test artifacts**: `.pytest_cache/`, `.coverage`, `htmlcov/`
- **Tests**: `tests/`, `test/`
```

subssn21: Hey Dan,

So to understand what you are suggesting. remove the individual file filters for test_ and _test and rely on the test directory exclusion and be sure to properly document it?

dan-hook: Exactly.  Thanks!

examples/python-layouts/monorepo-layout/README.md
dan-hook, line 1: I'm having trouble getting this example to deploy.  `uv sync` runs in each the services directories, then the builder tries to copy the .venv directory created by `uv sync`.  I think that should be excluded.  The copy fails because there's a symlink lib64->lib.    

I'm chasing the bug down now.

```
Error: incremental build failed: [error:project_structure] | failed to install dependencies: failed to copy source files: failed to copy directory .venv: read /workspaces/redacted/sst/examples/python-layouts/monorepo-layout/services/auth/.venv/lib64: is a directory | Context: originalContext=build execution, packagesPlanned=1, estimatedDuration=15.1s | Suggestions: Check build logs for specific package failures
    at IncomingMessage.<anonymous> (file:///workspaces/redacted/sst/examples/python-layouts/monorepo-layout/.sst/platform/src/components/rpc/rpc.ts:42:22)
    at IncomingMessage.emit (node:events:520:35)
    at endReadableNT (node:internal/streams/readable:1701:12)
    at processTicksAndRejections (node:internal/process/task_queues:89:21) {
  promise: Promise { <rejected> [Circular *1] }
```

subssn21: I think I have a fix for the issue. It should have ignored the .venv directory, but it didn't. I will test a little bit more and try to get it out tomorrow

dan-hook: I put up a PR to your PR with a fix, plus more.

dan-hook: This is still an issue in your latest commit.  I believe that after you merged my PR, you removed all of its commits with your next force commit.  

The problem seems to be that incremental builder, for non-simple layouts, iterates over all directories in the workspace and adds them to include without checking the content filter.  Here's [one place the content filter should be added](https://github.com/dan-hook/sst/blob/feat/uv-workspace-package-support/pkg/runtime/python/incremental_builder.go#L1269)

subssn21: So sorry, I didn't mean to do that, Let me see if I can get it fixed.

dan-hook: My [second PR](https://github.com/subssn21/sst/pull/2) includes all the same changes as the first one, plus better uv workspace support.

subssn21: I think I have your commits successfully merged now, Sorry about that

julen:
Hey Michael, amazing work here, can't wait for it to be merged and released! Thanks for all the effort ❤️ 

QQ: do you think it'd make sense to incorporate [support for Python 3.14 as part of this PR](https://github.com/sst/sst/pull/6061/files#diff-ea83e80a8a278f701ea3333102457d15c16ace0944c7c7dfd408da21ae33fdb4)? FWIW both [AWS Lambda](https://docs.aws.amazon.com/lambda/latest/dg/lambda-python.html) as well as [uv](https://astral.sh/blog/python-3.14) support it.

subssn21:
Does everyone feel that this is in a good enough state to merge?

There are some things that could be better but they would require changes outside of the Python runtime code, and would therefore require more work/coordination with others.

examples/python-modern-uv/README.md
```diff
@@ -0,0 +1,123 @@
...
+- Cross-package imports work in Lambda
+
+## Known Issues to Check
+
+1. **PYTHONPATH in dev mode**: Does SST set PYTHONPATH to include src/?
+2. **uv sync before dev**: Does SST call `uv sync` before starting dev mode?
+3. **Deployment bundling**: Are workspace packages bundled correctly?
+4. **Editable installs**: Does deployment convert editable installs to regular installs?
+
+## Success Criteria
```
dan-hook: My biggest motivation for working on this layout was so that the lambdas would have distinct sets of dependencies.   I think that worker should include a dependency that WorkspaceHandler should have a dependency that PackageHandler does not, and that one of the Success Criteria should be **Minimal Deployment Size**: Dependencies scoped per function.

subssn21 (commented):
@dan-hook Does this cover your concern?

dan-hook:
> @dan-hook Does this cover your concern?

No.  The test just checks to see if the package is in the pyproject.toml.  We want to test the dependency resolver/installer pipeline in incremental_builder.go and make sure that it performs correctly.  A hypothetical test for this is to deploy the example, then look in .sst/artifacts/PackageHandler/code.zip and ensure that it doesn't contain arrow.

subssn21:
> > @dan-hook Does this cover your concern?
> 
> No. The test just checks to see if the package is in the pyproject.toml. We want to test the dependency resolver/installer pipeline in incremental_builder.go and make sure that it performs correctly. A hypothetical test for this is to deploy the example, then look in .sst/artifacts/PackageHandler/code.zip and ensure that it doesn't contain arrow.

I have added a new test

sphorner-stolle:
Any possibilities that this will be approved and merged @thdxr @fwang in the near future?

subssn21:
This should be in a pretty good state now. I have been using with a number of different layouts in real code and it its working quite well now. The weird issues we were seeing before are gone. THe code has been cleaned up significantly and should be much better.

pkg/runtime/python/uv_runner.go
```diff
@@ -0,0 +1,1295 @@
...
+	// Add suggestions based on common error patterns
+	if strings.Contains(result.Stderr, "package") && strings.Contains(result.Stderr, "not found") {
+		errorMsg += "\nSuggestion: Check if the package name is correct and exists in the workspace"
+	} else if strings.Contains(result.Stderr, "permission") {
+		errorMsg += "\nSuggestion: Check file permissions for the output directory"
+	} else if strings.Contains(result.Stderr, "lock") {
+		errorMsg += "\nSuggestion: Run 'uv sync' first to ensure dependencies are resolved"
+	}
+
+	return fmt.Errorf(errorMsg)
```
julen: I attempted to make a CI build in my fork to try test this branch, but the `goreleaser/goreleaser-action` step failed with:
```
Error:     │ pkg/runtime/python/uv_runner.go:552:20: non-constant format string in call to fmt.Errorf
Error:     │ pkg/runtime/python/uv_runner.go:723:20: non-constant format string in call to fmt.Errorf
Error:     │ pkg/runtime/python/uv_runner.go:1294:20: non-constant format string in call to fmt.Errorf
```

This happened while running `go test`.

My understanding is a non-constant format string needs to be used in `fmt.Errorf`, [because of the diagnostics run during `go test`](https://tip.golang.org/doc/go1.24#vet).

```suggestion
	return fmt.Errorf("%s", errorMsg)
```

As you see in the error message, the same applies to lines 723 and 1294 in this file.

julen:
@vimtor is this PR in your pipeline to review and merge? I'm bringing this to your attention as the set of fixes introduced here would be of tremendous help for folks deploying Python projects.

vimtor:
@julen i didn't have this on my pipeline

i'm not super familiar with the python ecosystem, i can give it a try. ideally someone with more sst + python experience helps me test the changes.

pkg/runtime/python/uv_runner.go
subssn21: Should be fixed in latest push

subssn21 (commented):
I had to add a change to the base api for the runtime manager that runs dev runners for the various language runtimes. The default method of running all the processes on a rebuild was causing huge issues in python since we can't track what files were affected the way node does. We just changed it to lazy start the workers on invoke instead so now it works like real lambda with a cold start on first run after a rebuild. We made it configurable so that all other runtimes continue to function as before. If a runtime wants to opt in to the new behavior they just need to change what they return from ShouldRunEagerly

dsonet (approved):

ericmillsio:
@vimtor I know this isn't your area of expertise, but I can attest to @subssn21's work as him and I talked extensively through the problems and edge cases, and I sent a ton of test case situations to him that I was running into. The current state of python in sst is pretty broken/slow without this, if we can stabilize it at the very least (which is this PR), it can clear a lot of headaches. 

Since it seems there is also movement to a 4.0 as well (yay!), it might be a good time to introduce these changes? Several other users have ran @subssn21 branch, and I hope his work (which was a lot, 6 months in the making) doesn't go to waste. Python is critical to an AI stack (which is everyone now).

Cheers, and thanks for picking up the load to keep sst going!

sphorner-stolle:
@subssn21 just wanted to thank you for all your work on this and [the Python improvements PR](https://github.com/anomalyco/sst/pull/5132). Using your 5132 PR branch kept us able to deploy for the last few months before it was finally merged to dev.

vimtor:
@subssn21 can you add me on discord? i'm on the sst server too

i would like to have a quick call with you to review this pr and merge it by next week

package-lock.json
vimtor, line 1: can you uncommit this file? the global `package.json` should be installed using `bun install` not `npm`

python.test
vimtor, line 1: i guess this file should be deleted?

test_imports.py
vimtor, line 1: i guess this file should be deleted?

pkg/project/project.go
```diff
@@ -127,17 +127,26 @@ func New(input *ProjectConfig) (*Project, error) {
...
-			node.New(input.Version),
-			worker.New(),
-			python.New(),
-			golang.New(),
-			rust.New(),
-		),
 	}
 	tmp := proj.PathWorkingDir()
 
+	// Create Python runtime with caching enabled
```
vimtor: minor style change: probably we can move these declarations above the `&Project{` so we don't assign the runtime after the struct creation

pkg/runtime/python/sst_module.py
vimtor, line 1: if i'm not mistaken, this file isn't being used

pkg/events/function.go
```diff
@@ -0,0 +1,8 @@
+package events
+
+// FunctionBuildProgressEvent represents a build progress event for the SST live console
+type FunctionBuildProgressEvent struct {
```
vimtor: not 100% sure why we need this new event instead of using the existing FunctionBuildEvent

platform/src/components/aws/function.ts
```diff
@@ -2400,7 +2401,7 @@ export class Function extends Component implements Link.Linkable {
 
           // Calculate hash of the zip file
           const hash = crypto.createHash("sha256");
-          hash.update(await fs.promises.readFile(zipPath, "utf-8"));
+          hash.update(await fs.promises.readFile(zipPath));
```
vimtor: i guess this change is performance related?

platform/src/components/aws/function.ts
```diff
@@ -2475,6 +2476,7 @@ export class Function extends Component implements Link.Linkable {
           concurrency,
           dev,
         ]) => {
+
```
vimtor:
```suggestion
```

platform/functions/python-runtime/index.py
```diff
@@ -1,10 +1,29 @@
...
+import logging
 import os
 import sys
 import traceback
 import time
 import requests
+from pathlib import Path
+
+
+# Configure Python logging to output to stdout so it appears alongside print statements
```
vimtor: are these logs for `sst dev`?

examples/python-modern-uv/pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+description = "Test modern uv project structures"
+dependencies = ["requests>=2.31.0"]
+requires-python = ">=3.11"
+
+[tool.uv]
+package = true
+
+[tool.uv.workspace]
+members = ["packages/*"]
+
```
vimtor, line 12: i had to add the following to make this example work:

```suggestion

[tool.hatch.build.targets.wheel]
packages = ["src/myapp"]

```

examples/python-modern-uv/sst.config.ts
```diff
@@ -0,0 +1,36 @@
...
+      name: "python-modern-uv",
+      removal: input?.stage === "production" ? "retain" : "remove",
+      home: "aws",
+    };
+  },
+  async run() {
+    // Test 1: Root entry point with src/ layout
+    const rootHandler = new sst.aws.Function("RootHandler", {
+      handler: "handler.lambda_handler",
+      runtime: "python3.11",
```
vimtor, line 72: i had to add

```suggestion
      runtime: "python3.11",
      url: true
```

to the functions in this example otherwise i got `Function URL is not enabled. Enable it with "url: true".` error

examples/python-modern-uv/sst.config.ts
```diff
@@ -0,0 +1,36 @@
...
+    // Test 3: Workspace member importing another member
+    const workspaceHandler = new sst.aws.Function("WorkspaceHandler", {
+      handler: "packages/worker/src/worker/handler.lambda_handler",
+      runtime: "python3.11",
+    });
+
+    return {
+      rootHandler: rootHandler.url,
+      packageHandler: packageHandler.url,
+      workspaceHandler: workspaceHandler.url,
```
vimtor, line 91: tried deploying these and got the following:

### `sst dev`

1. root: `Root handler with src/ imports: Processed: {'test': 'data'}`
2. package: `Internal Server Error`

<details>

<summary>logs</summary>

```
|  Build       RunComplete: Started packages/api/src/api/handler.lambda_handler in 23.058125ms
|  +2562047h   Failed to load handler: /Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/PackageHandler-d
ev/packages/api/handler.lambda_handler
|  +2562047h     Error: No module named 'packages.api.handler'
|  +2562047h     Module: Users.vimtor.Developer.sst.python-deployment-fixes.examples.python-modern-uv..sst.artifacts.PackageHandler-dev.packages.api
.handler
|  +2562047h     Working dir: /Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/PackageHandler-dev
|  +2562047h     PYTHONPATH: not set
|  +2562047h     sys.path: ['/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/PackageHandler-dev', '/Use
rs/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python311.zip', '/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-
aarch64-none/lib/python3.11', '/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/lib-dynload', '/Users/vimtor/D
eveloper/sst.python-deployment-fixes/examples/python-modern-uv/.venv/lib/python3.11/site-packages', '/Users/vimtor/Developer/sst.python-deployment-f
ixes/examples/python-modern-uv/src']
|  Error       packages/api/src/api/handler.lambda_handler
|  No module named 'packages.api.handler'
|  ↳ Traceback (most recent call last):
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/PackageHandler-dev/lambdaric_python_bridge.p
y", line 173, in <module>
|  ↳ module, handler_function = resolve_handler_legacy(handler, artifact_dir)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/PackageHandler-dev/lambdaric_python_bridge.p
y", line 139, in resolve_handler_legacy
|  ↳ module = importlib.import_module(python_module_path)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/importlib/__init__.py", line 126, in import_module
|  ↳ return _bootstrap._gcd_import(name[level:], package, level)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
|  ↳ File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
|  ↳ File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
```

</details>

3. workspace: `Internal Server Error`

<details>

<summary>logs</summary>

```
|  Build       RunComplete: Started packages/worker/src/worker/handler.lambda_handler in 21.073334ms
|  +2562047h   Failed to load handler: /Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/WorkspaceHandler
-dev/packages/worker/handler.lambda_handler
|  +2562047h     Error: No module named 'packages.worker.handler'
|  +2562047h     Module: Users.vimtor.Developer.sst.python-deployment-fixes.examples.python-modern-uv..sst.artifacts.WorkspaceHandler-dev.packages.w
orker.handler
|  +2562047h     Working dir: /Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/WorkspaceHandler-dev
|  +2562047h     PYTHONPATH: not set
|  +2562047h     sys.path: ['/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/WorkspaceHandler-dev', '/U
sers/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python311.zip', '/Users/vimtor/.local/share/uv/python/cpython-3.11.13-maco
s-aarch64-none/lib/python3.11', '/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/lib-dynload', '/Users/vimtor
/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.venv/lib/python3.11/site-packages', '/Users/vimtor/Developer/sst.python-deployment
-fixes/examples/python-modern-uv/src']
|  Error       packages/worker/src/worker/handler.lambda_handler
|  No module named 'packages.worker.handler'
|  ↳ Traceback (most recent call last):
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/WorkspaceHandler-dev/lambdaric_python_bridge
.py", line 173, in <module>
|  ↳ module, handler_function = resolve_handler_legacy(handler, artifact_dir)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/WorkspaceHandler-dev/lambdaric_python_bridge
.py", line 139, in resolve_handler_legacy
|  ↳ module = importlib.import_module(python_module_path)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/importlib/__init__.py", line 126, in import_module
|  ↳ return _bootstrap._gcd_import(name[level:], package, level)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
|  ↳ File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
|  ↳ File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
```

</details>

### `sst deploy`

1. root: `Internal Server Error`

<details>

<summary>logs</summary>

```
[ERROR] Runtime.ImportModuleError: Unable to import module 'handler': No module named 'src'
```

</details>

2. package: `Response: API handler works!`
3. workspace: `Response: Worker importing from API package`

examples/aws-python/sst.config.ts
vimtor, line 1: <details>

<summary>during `sst dev` i get `Internal Server Error` with these logs</summary>

```
h64-none/lib/python3.11', '/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/lib-dynload', '/Users/vimtor/.cach
e/uv/builds-v0/.tmpGNqEh9/lib/python3.11/site-packages', '/Users/vimtor/.cache/uv/archive-v0/uKckjt25Yxnx0CK5FThXr/lib/python3.11/site-packages', '/
Users/vimtor/Developer/sst.python-deployment-fixes/examples/aws-python/.venv/lib/python3.11/site-packages']
|  Error       functions/src/functions/api.handler
|  No module named 'sst'
|  ↳ Traceback (most recent call last):
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/aws-python/.sst/artifacts/MyPythonFunction-dev/lambdaric_python_bridge.py",
line 173, in <module>
|  ↳ module, handler_function = resolve_handler_legacy(handler, artifact_dir)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/aws-python/.sst/artifacts/MyPythonFunction-dev/lambdaric_python_bridge.py",
line 139, in resolve_handler_legacy
|  ↳ module = importlib.import_module(python_module_path)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/importlib/__init__.py", line 126, in import_module
|  ↳ return _bootstrap._gcd_import(name[level:], package, level)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
|  ↳ File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
|  ↳ File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
|  ↳ File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
|  ↳ File "<frozen importlib._bootstrap_external>", line 940, in exec_module
|  ↳ File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/aws-python/.sst/artifacts/MyPythonFunction-dev/functions/api.py", line 1, in
 <module>
|  ↳ from sst import Resource
```

</details>

update: i had to run `uv sync --all-packages` to make it work in dev (not sure if this is common in python)

<details>

<summary>`sst deploy` i get `Internal Server Error` with these logs in AWS</summary>

```
[ERROR] Runtime.ImportModuleError: Unable to import module 'functions/src/functions/api': /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /var/task/cryptography/hazmat/bindings/_rust.abi3.so)
```

</details>

pkg/runtime/python/content_filter.go
vimtor: good point

@subssn21 what do you think?

docs/python-best-practices.md
```diff
@@ -0,0 +1,611 @@
+# Python Lambda Best Practices
```
vimtor: i'll merge a separate pr with Reference/Runtime/Node.js so we can move most of this docs there

i'll let you know once it's merged so you can update with dev

examples/aws-python-huggingface/functions/pyproject.toml
```diff
@@ -7,7 +7,7 @@ dependencies = ["transformers>=4.44.2", "torch==2.3.1"]
 
 # It is recommended to specify your python version to match your Lambda runtime otherwise you may
 # encounter issues with dependencies.
-requires-python = "==3.12.*"
+requires-python = "==3.13.*"
```
vimtor: had to downgrade this to make it work because my cli was telling me that `torch==2.3.1` doesn't have 3.13 wheels

```suggestion
requires-python = "==3.12.*"
```

examples/aws-python-huggingface/pyproject.toml
```diff
@@ -7,7 +7,7 @@ dependencies = []
 
 # It is recommended to specify your python version to match your Lambda runtime otherwise you may
 # encounter issues with dependencies.
-requires-python = "==3.12.*"
+requires-python = "==3.13.*"
```
vimtor: had to downgrade this to make it work because my cli was telling me that `torch==2.3.1` doesn't have 3.13 wheels

```suggestion
requires-python = "==3.12.*"
```

pkg/events/function.go
vimtor: i believe this is also responsible for cluttering the logs with `RunBuild` and things like that:

Image

also the errors are not formatted as errors (they appear blue) and are shown twice

pkg/runtime/python/python.go
```diff
@@ -99,23 +320,70 @@ func (r *PythonRuntime) Build(ctx context.Context, input *runtime.BuildInput) (*
 
 	file, err := r.getFile(input)
 	if err != nil {
-		return nil, fmt.Errorf("handler not found: %v", err)
+		bus.Publish(&events.FunctionBuildProgressEvent{
```
vimtor: related to https://github.com/anomalyco/sst/pull/6061/changes#r2877878837

not sure we need these anymore, they seem to be for debugging purposes

we don't have the equivalent for node.js

examples/aws-python-huggingface/sst.config.ts
vimtor, line 1: after `sst deploy` i get `Internal Server Error` as well with the following logs in AWS:

```
[ERROR] Runtime.ImportModuleError: Unable to import module 'functions.src.functions.api': cannot import name '_regex' from partially initialized module 'regex' (most likely due to a circular import) (/var/task/regex/__init__.py)
```

examples/python-layouts/flat-layout/sst.config.ts
vimtor, line 1: i got the following warning for this example:

```
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.11`
```

pkg/runtime/python/python.go
vimtor: to be more precise about the "noise"

here's an example of me in `sst dev` modifying and invoking a lambda again.

i get the following logs:

```
|  Build       services/api/handler.main
|  Build       services/auth/handler.main
|  Build       RunComplete: Started services/api/handler.main in 16.2755ms
|  +22.589s    Loaded services/api/handler.main
|  Invoke      services/api/handler.main
|  Done        took +457ms
```

it seems to me that the following are not necessary?

```
|  Build       RunComplete: Started services/api/handler.main in 16.2755ms
|  +22.589s    Loaded services/api/handler.main
```

examples/aws-python/test-install/.lock
vimtor, line 1: not sure if this `.lock` file needs to be committed

examples/python-layouts/monorepo-layout/sst.config.ts
```diff
@@ -0,0 +1,42 @@
...
+    // API service - uses root pyproject.toml
+    const api = new sst.aws.Function("ApiService", {
+      handler: "services/api/handler.main",
+      runtime: "python3.12",
+      timeout: "30 seconds",
+      url: true, // Enable function URL
+    });
+
+    // Worker service - has its own pyproject.toml in services/worker/
+    const worker = new sst.aws.Function("WorkerService", {
```
vimtor, line 49: is it possible that this example is incomplete? it feels like the worker functions should be an sqs subscription bu there is no queue defined in the config

cmd/sst/mosaic/ui/function_build_progress_integration_test.go
vimtor, line 1: this test is probably unnecessary (besides the fact that the `FunctionBuildProgressEvent` will probably get deleted

platform/functions/docker/python.Dockerfile
```diff
@@ -2,30 +2,11 @@
...
-  else \
-  yum install -y git gcc; \
-  fi
-
-# Copy requirements and install dependencies
-COPY requirements.txt ${LAMBDA_TASK_ROOT}/requirements.txt
-
-# Mount the uv image to install the dependencies - uv will not be installed in the final image
-RUN --mount=from=ghcr.io/astral-sh/uv,source=/uv,target=/bin/uv \
-  uv pip install -r requirements.txt --target ${LAMBDA_TASK_ROOT} --system --compile-bytecode
```
vimtor, line 20: i believe `-compile-bytecode` prevents lambda cold starts? might be wrong

pkg/runtime/python/python.go
```diff
@@ -11,13 +11,45 @@ import (
...
 )
 
+// Global sync tracker shared across all builds
+var (
+	globalSyncCompleted = make(map[string]bool)
+	globalSyncMutex     sync.RWMutex
+
+	// Global build semaphore to limit concurrent builds
+	// This prevents system overload when Pulumi tries to build 100+ functions in parallel
+	globalBuildSemaphore = make(chan struct{}, 4) // Allow max 4 concurrent builds
```
vimtor: shouldn't the parallel build follow the `SST_CONCURRENCY` flags?

pkg/runtime/python/python.go
vimtor, line 1: <details>

<summary>in general i found a lot of dead code</summary>

```
❯ /Users/vimtor/golang/bin/deadcode -filter 'pkg/runtime/python' ./cmd/sst/
pkg/runtime/python/content_filter.go:39:6: unreachable func: NewContentFilterWithPatterns
pkg/runtime/python/dependency_analyzer.go:534:31: unreachable func: DependencyAnalyzer.isInTypicalPackageLocation
pkg/runtime/python/dependency_analyzer.go:551:31: unreachable func: DependencyAnalyzer.followsPackageNamingConventions
pkg/runtime/python/dependency_analyzer.go:815:31: unreachable func: DependencyAnalyzer.walkPackageDirectory
pkg/runtime/python/dependency_analyzer.go:820:31: unreachable func: DependencyAnalyzer.walkPackageDirectoryRecursive
pkg/runtime/python/dependency_analyzer.go:904:31: unreachable func: DependencyAnalyzer.shouldSkipSourceDirectory
pkg/runtime/python/dependency_analyzer.go:1311:31: unreachable func: DependencyAnalyzer.dependenciesEqual
pkg/runtime/python/dependency_analyzer.go:1346:31: unreachable func: DependencyAnalyzer.localPackagesEqual
pkg/runtime/python/dependency_analyzer.go:1384:31: unreachable func: DependencyAnalyzer.externalDependenciesEqual
pkg/runtime/python/errors.go:272:6: unreachable func: NewProjectStructureError
pkg/runtime/python/errors.go:288:6: unreachable func: NewCacheCorruptedError
pkg/runtime/python/errors.go:346:6: unreachable func: NewDependencyFailedError
pkg/runtime/python/errors.go:357:6: unreachable func: NewBuildOutputMissingError
pkg/runtime/python/errors.go:375:6: unreachable func: NewExtractionFailedError
pkg/runtime/python/errors.go:413:6: unreachable func: NewModuleMoveFailedError
pkg/runtime/python/errors.go:446:6: unreachable func: NewValidationFailedError
pkg/runtime/python/errors.go:463:6: unreachable func: NewArtifactOversizedError
pkg/runtime/python/errors.go:487:6: unreachable func: NewModuleMissingError
pkg/runtime/python/errors.go:570:34: unreachable func: ErrorRecoveryManager.RecoverFromError
pkg/runtime/python/errors.go:582:34: unreachable func: ErrorRecoveryManager.executeRecoveryAction
pkg/runtime/python/factory.go:78:27: unreachable func: RuntimeFactory.CreateDefaultProgressCallback
pkg/runtime/python/helpers.go:17:24: unreachable func: PathHelpers.GetCacheDir
pkg/runtime/python/helpers.go:30:24: unreachable func: PathHelpers.GetSstDir
pkg/runtime/python/helpers.go:35:24: unreachable func: PathHelpers.IsWithinProject
pkg/runtime/python/incremental_builder.go:1829:31: unreachable func: IncrementalBuilder.installDependenciesFresh
pkg/runtime/python/incremental_builder.go:1850:31: unreachable func: IncrementalBuilder.parseDependencyList
pkg/runtime/python/incremental_builder.go:1882:31: unreachable func: IncrementalBuilder.shouldUseDependencyCache
pkg/runtime/python/incremental_builder.go:2074:31: unreachable func: IncrementalBuilder.sortRequirementsFile
pkg/runtime/python/incremental_builder.go:2446:31: unreachable func: IncrementalBuilder.containsPythonContent
pkg/runtime/python/incremental_builder.go:2937:31: unreachable func: IncrementalBuilder.copyWorkspacePackageSources
pkg/runtime/python/incremental_builder.go:3349:31: unreachable func: IncrementalBuilder.isWorkspacePackage
pkg/runtime/python/incremental_builder.go:3359:31: unreachable func: IncrementalBuilder.shouldSkipPackage
pkg/runtime/python/progress.go:23:6: unreachable func: NewBuildProgressReporter
pkg/runtime/python/progress.go:32:35: unreachable func: BuildProgressReporter.ReportStage
pkg/runtime/python/progress.go:61:35: unreachable func: BuildProgressReporter.ReportSuccess
pkg/runtime/python/progress.go:98:35: unreachable func: BuildProgressReporter.ReportWarning
pkg/runtime/python/progress.go:124:35: unreachable func: BuildProgressReporter.ReportModuleIncluded
pkg/runtime/python/progress.go:136:35: unreachable func: BuildProgressReporter.ReportArtifactSummary
pkg/runtime/python/progress.go:149:35: unreachable func: BuildProgressReporter.ReportBuildComplete
pkg/runtime/python/progress.go:170:35: unreachable func: BuildProgressReporter.getStageTimings
pkg/runtime/python/progress.go:202:35: unreachable func: BuildProgressReporter.ValidateAndReportModules
pkg/runtime/python/python.go:574:6: unreachable func: appendIfMissing
pkg/runtime/python/python.go:910:25: unreachable func: PythonRuntime.getHandlerForFunction
pkg/runtime/python/python.go:957:25: unreachable func: PythonRuntime.isFileRelatedToFunction
pkg/runtime/python/python.go:1342:6: unreachable func: fileExists
pkg/runtime/python/python.go:1348:6: unreachable func: copyDirectory
pkg/runtime/python/python.go:1396:6: unreachable func: shouldSkipFile
pkg/runtime/python/python.go:1487:25: unreachable func: PythonRuntime.getWorkspaceDirectory
pkg/runtime/python/python.go:1538:25: unreachable func: PythonRuntime.getPackageName
pkg/runtime/python/python.go:1573:25: unreachable func: PythonRuntime.adjustHandlerPath
pkg/runtime/python/python.go:1614:25: unreachable func: PythonRuntime.calculateArtifactSummary
pkg/runtime/python/python.go:1651:25: unreachable func: PythonRuntime.updateCacheAfterBuild
pkg/runtime/python/python.go:1753:25: unreachable func: PythonRuntime.copySourceFiles
pkg/runtime/python/python.go:1774:25: unreachable func: PythonRuntime.copyPythonFilesWithProgress
pkg/runtime/python/python.go:1822:25: unreachable func: PythonRuntime.copyPythonFiles
pkg/runtime/python/python.go:1885:25: unreachable func: PythonRuntime.installLocalPackages
pkg/runtime/python/python.go:2099:25: unreachable func: PythonRuntime.copyDirectory
```

</details>



<details>

<summary>with another command</summary>

```
❯ /Users/vimtor/golang/bin/staticcheck -checks U1000 ./pkg/runtime/python/...
pkg/runtime/python/dependency_analyzer.go:534:31: func (*DependencyAnalyzer).isInTypicalPackageLocation is unused (U1000)
pkg/runtime/python/dependency_analyzer.go:551:31: func (*DependencyAnalyzer).followsPackageNamingConventions is unused (U1000)
pkg/runtime/python/dependency_analyzer.go:815:31: func (*DependencyAnalyzer).walkPackageDirectory is unused (U1000)
pkg/runtime/python/dependency_analyzer.go:820:31: func (*DependencyAnalyzer).walkPackageDirectoryRecursive is unused (U1000)
pkg/runtime/python/dependency_analyzer.go:904:31: func (*DependencyAnalyzer).shouldSkipSourceDirectory is unused (U1000)
pkg/runtime/python/dependency_analyzer.go:1311:31: func (*DependencyAnalyzer).dependenciesEqual is unused (U1000)
pkg/runtime/python/dependency_analyzer.go:1346:31: func (*DependencyAnalyzer).localPackagesEqual is unused (U1000)
pkg/runtime/python/dependency_analyzer.go:1384:31: func (*DependencyAnalyzer).externalDependenciesEqual is unused (U1000)
pkg/runtime/python/incremental_builder.go:1829:31: func (*IncrementalBuilder).installDependenciesFresh is unused (U1000)
pkg/runtime/python/incremental_builder.go:1850:31: func (*IncrementalBuilder).parseDependencyList is unused (U1000)
pkg/runtime/python/incremental_builder.go:1882:31: func (*IncrementalBuilder).shouldUseDependencyCache is unused (U1000)
pkg/runtime/python/incremental_builder.go:2074:31: func (*IncrementalBuilder).sortRequirementsFile is unused (U1000)
pkg/runtime/python/incremental_builder.go:2446:31: func (*IncrementalBuilder).containsPythonContent is unused (U1000)
pkg/runtime/python/incremental_builder.go:2937:31: func (*IncrementalBuilder).copyWorkspacePackageSources is unused (U1000)
pkg/runtime/python/incremental_builder.go:3349:31: func (*IncrementalBuilder).isWorkspacePackage is unused (U1000)
pkg/runtime/python/incremental_builder.go:3359:31: func (*IncrementalBuilder).shouldSkipPackage is unused (U1000)
pkg/runtime/python/python.go:26:2: var globalSyncCompleted is unused (U1000)
pkg/runtime/python/python.go:27:2: var globalSyncMutex is unused (U1000)
pkg/runtime/python/python.go:574:6: func appendIfMissing is unused (U1000)
pkg/runtime/python/python.go:910:25: func (*PythonRuntime).getHandlerForFunction is unused (U1000)
pkg/runtime/python/python.go:957:25: func (*PythonRuntime).isFileRelatedToFunction is unused (U1000)
pkg/runtime/python/python.go:1348:6: func copyDirectory is unused (U1000)
pkg/runtime/python/python.go:1396:6: func shouldSkipFile is unused (U1000)
pkg/runtime/python/python.go:1487:25: func (*PythonRuntime).getWorkspaceDirectory is unused (U1000)
pkg/runtime/python/python.go:1538:25: func (*PythonRuntime).getPackageName is unused (U1000)
pkg/runtime/python/python.go:1573:25: func (*PythonRuntime).adjustHandlerPath is unused (U1000)
pkg/runtime/python/python.go:1614:25: func (*PythonRuntime).calculateArtifactSummary is unused (U1000)
pkg/runtime/python/python.go:1651:25: func (*PythonRuntime).updateCacheAfterBuild is unused (U1000)
pkg/runtime/python/python.go:1753:25: func (*PythonRuntime).copySourceFiles is unused (U1000)
pkg/runtime/python/python.go:1774:25: func (*PythonRuntime).copyPythonFilesWithProgress is unused (U1000)
pkg/runtime/python/python.go:1822:25: func (*PythonRuntime).copyPythonFiles is unused (U1000)
pkg/runtime/python/python.go:1885:25: func (*PythonRuntime).installLocalPackages is unused (U1000)
pkg/runtime/python/python.go:2099:25: func (*PythonRuntime).copyDirectory is unused (U1000)
```

</details>

i guess we should remove most of this?

pkg/runtime/python/incremental_builder.go
```diff
@@ -0,0 +1,3523 @@
...
+				currentDir := filepath.Dir(pyprojectDir)
+
+				slog.Info("walking up to find parent pyproject.toml",
+					"pyprojectDir", pyprojectDir,
+					"startingDir", currentDir,
+					"sourceRoot", projectInfo.SourceRoot)
+
+				// Walk up to find any parent pyproject.toml
+				iterCount := 0
+				for currentDir != "/" && currentDir != "." && currentDir != projectInfo.SourceRoot {
```
vimtor: i believe `currentDir != "/"` will break on windows

sdk/python/src/sst/__init__.py
```diff
@@ -2,7 +2,7 @@
 import json
 import base64
 from typing import Dict, Any
-from cryptography.hazmat.primitives.ciphers.aead import AESGCM
+from Crypto.Cipher import AES
```
vimtor: could we update this to prevent possible namespace collisions with the pycrypto module?

```suggestion
from Cryptodome.Cipher import AES
```

pkg/runtime/python/python.go
```diff
@@ -164,9 +530,15 @@ func (r *PythonRuntime) Run(ctx context.Context, input *runtime.RunInput) (runti
 		return nil, fmt.Errorf("failed to create stderr pipe: %v", err)
 	}
 
-	slog.Info("starting worker", "env", cmd.Env, "args", cmd.Args)
 	cmd.Start()
```
vimtor: this is a pre-existing error but we shoulnd't discard the err returned by `cmd.Start`

pkg/runtime/python/python.go
```diff
@@ -176,339 +548,1658 @@ func (r *PythonRuntime) Run(ctx context.Context, input *runtime.RunInput) (runti
...
+	}
+
+	slog.Info("installLocalPackages completed", "elapsed", time.Since(startTime), "processedPackages", len(packages))
+	return nil
+}
+
+// flattenWorkspaceLayouts detects and flattens package/src/package structures for all legacy projects
+func (r *PythonRuntime) flattenWorkspaceLayouts(artifactDir, functionID string) error {
+	// Create debug log file
+	logFile, err := os.Create(filepath.Join(artifactDir, "flatten_debug.log"))
```
vimtor: should this `flatten_debug.log` file be included in the uploaded artifact?

vimtor (changes requested):
thanks again for your contribution @subssn21 

i've left a bunch of comments. some general remarks:

1. the thing that worries me the most is all the dead code introduced by ai. we should get rid of it and simplify as much as we can. i'm waiting for it to get into the build details
2. some of the examples are incomplete. i'm sure you were using them for testing. feel free to simplify them if you don't want to bother (the sqs one is a good example of this). the examples are important for me so i can test the changes since i don't have any big python project besides them. also it's important that they work for new users trying out sst

platform/src/components/aws/function.ts
subssn21: Here's what I found:
Why the first option is incorrect
ZIP files are binary data, not text. Using "utf-8" forces Node.js to decode the raw bytes into a UTF-8 string. 
Data Corruption: If the binary file contains byte sequences that are invalid in UTF-8, they will be replaced by the "replacement character" (`` or \uFFFD).
Incorrect Hash: Because the data has been altered during decoding, the resulting hash will not match the actual file's checksum. 

Why the second option is correct
Raw Buffer: When you omit the encoding, readFile returns a Buffer containing the exact, raw bytes of the file.
Integrity: Hashing a Buffer ensures that every single bit of the original file is processed exactly as it exists on disk, which is essential for verifying file integrity.

sdk/python/src/sst/__init__.py
subssn21: Cryptodome uses Crypto as its namespace because it was meant as a drop in replacement for pycyrpto, but since we aren't using it that way here, I'll switch the package to pycryptodomex which uses the Cryptodome namespace.

examples/aws-python/sst.config.ts
subssn21: So this is an interesting one. So the biggest fix was that I just changed the way it pulls python wheels (The binary builds of python projects that have c dependencies so the glibc issue is fixed). So the deploy issue is fixed. In dev we don't automatically call uv sync in dev (I think you automatically call npm install or the equivalent for node stuff, at least for frontend), We could probably add that. It only needs to be called once per time the dependencies change so tracking it may be a pain. The typical use case though would be in dev a user would run uv pip install on the new dependency, and uv sync on setting up a new project, but obviously there are cases that would miss like update from git tha has a new dependency, and you would have to run uv sync again. It's pretty much the same as npm install in this case as far as what it does/how it works

examples/python-layouts/nested-layout/test_capabilities.py
vimtor, line 1: i believe we can safely delete this file

examples/python-layouts/nested-layout/sst.config.ts
```diff
@@ -0,0 +1,93 @@
...
+    // Auth function in another nested location
+    const auth = new sst.aws.Function("AuthFunction", {
+      handler: "app/functions/auth/handler.main",
+      runtime: "python3.11",
+      timeout: "15 seconds",
+      url: true  // Enable function URL for auth endpoints
+    });
+
+    // Test function to validate deployment capabilities
+    const test = new sst.aws.Function("TestFunction", {
```
vimtor: this function fails in production with:

```
[ERROR] Runtime.ImportModuleError: Unable to import module 'app/functions/test/handler': No module named 'app.functions.test'
```

examples/python-modern-uv/sst.config.ts
vimtor, line 1: <details>

<summary>during `sst dev` the package url fails with these logs</summary>

```
|  +2562047h   Failed to load handler: /Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/PackageHandler-d
ev/packages/api/handler.lambda_handler
|  +2562047h     Error: No module named 'packages.api.handler'
|  +2562047h     Module: Users.vimtor.Developer.sst.python-deployment-fixes.examples.python-modern-uv..sst.artifacts.PackageHandler-dev.packages.api
.handler
|  +2562047h     Working dir: /Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/PackageHandler-dev
|  +2562047h     PYTHONPATH: not set
|  +2562047h     sys.path: ['/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/PackageHandler-dev', '/Use
rs/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python311.zip', '/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-
aarch64-none/lib/python3.11', '/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/lib-dynload', '/Users/vimtor/D
eveloper/sst.python-deployment-fixes/examples/python-modern-uv/.venv/lib/python3.11/site-packages', '/Users/vimtor/Developer/sst.python-deployment-f
ixes/examples/python-modern-uv/packages/api/src', '/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/src', '/Users/vimto
r/Developer/sst.python-deployment-fixes/examples/python-modern-uv/packages/worker/src']
|  Error       packages/api/src/api/handler.lambda_handler
|  No module named 'packages.api.handler'
|  ↳ Traceback (most recent call last):
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/PackageHandler-dev/lambdaric_python_bridge.p
y", line 173, in <module>
|  ↳ module, handler_function = resolve_handler_legacy(handler, artifact_dir)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/PackageHandler-dev/lambdaric_python_bridge.p
y", line 139, in resolve_handler_legacy
|  ↳ module = importlib.import_module(python_module_path)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/importlib/__init__.py", line 126, in import_module
|  ↳ return _bootstrap._gcd_import(name[level:], package, level)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
|  ↳ File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
|  ↳ File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
|  ↳
```

</details>

<details>

<summary>and the workspace url fails with these logs</summary>

```
|  +2562047h   Failed to load handler: /Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/WorkspaceHandler
-dev/packages/worker/handler.lambda_handler
|  +2562047h     Error: No module named 'packages.worker.handler'
|  +2562047h     Module: Users.vimtor.Developer.sst.python-deployment-fixes.examples.python-modern-uv..sst.artifacts.WorkspaceHandler-dev.packages.w
orker.handler
|  +2562047h     Working dir: /Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/WorkspaceHandler-dev
|  +2562047h     PYTHONPATH: not set
|  +2562047h     sys.path: ['/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/WorkspaceHandler-dev', '/U
sers/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python311.zip', '/Users/vimtor/.local/share/uv/python/cpython-3.11.13-maco
s-aarch64-none/lib/python3.11', '/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/lib-dynload', '/Users/vimtor
/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.venv/lib/python3.11/site-packages', '/Users/vimtor/Developer/sst.python-deployment
-fixes/examples/python-modern-uv/packages/api/src', '/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/src', '/Users/vim
tor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/packages/worker/src']
|  Error       packages/worker/src/worker/handler.lambda_handler
|  No module named 'packages.worker.handler'
|  ↳ Traceback (most recent call last):
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/WorkspaceHandler-dev/lambdaric_python_bridge
.py", line 173, in <module>
|  ↳ module, handler_function = resolve_handler_legacy(handler, artifact_dir)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/Users/vimtor/Developer/sst.python-deployment-fixes/examples/python-modern-uv/.sst/artifacts/WorkspaceHandler-dev/lambdaric_python_bridge
.py", line 139, in resolve_handler_legacy
|  ↳ module = importlib.import_module(python_module_path)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "/Users/vimtor/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/importlib/__init__.py", line 126, in import_module
|  ↳ return _bootstrap._gcd_import(name[level:], package, level)
|  ↳ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|  ↳ File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
|  ↳ File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
|  ↳ File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
|  ↳
```

</details>

pkg/runtime/python/build_test.go
```diff
@@ -178,3 +177,191 @@ boto3>=1.34.0`
 		}
 	})
 }
+
+// --- Content filter tests (merged from content_filter_test.go) ---
```
dsonet, line 149: I suppose this comment should be removed, it could cause confusion.

examples/aws-python/sst.config.ts
vimtor, line 1: just to confirm, it makes sense that for this example i need to run `uv sync --all-packages` right? since it is a monorepo

vimtor: just to confirm, did we review these existing jsdocs? are these things still the case?

vimtor: maybe we should document this somewhere? or is it really common for python folks to know this?

platform/src/components/aws/function.ts
vimtor: did this issue occur with python specifically? if yes we can keep it, otherwise we should open a new issue and revert these changes

examples/python-layouts/monorepo-layout/services/api/handler.py
```diff
@@ -0,0 +1,63 @@
...
+logger = logging.getLogger(__name__)
+
+
+def main(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
+    """Main API service handler."""
+    try:
+        path = event.get("rawPath") or event.get("path", "/")
+        method = event.get("requestContext", {}).get("http", {}).get("method") or event.get("httpMethod", "GET")
+
+        if method == "GET" and path == "/api/health":
```
vimtor: could we simplify all the new examples code a bit?

returning just a "hello world" style response (like the other examples) from each route should be enough

.github/workflows/check.yml
```diff
@@ -50,6 +50,11 @@ jobs:
         env:
           DOCKER_PUSH: false
 
+      - name: Setup uv
```
vimtor, line 56: just to confirm: do we still need this after all the tests we deleted?

examples/python-layouts/monorepo-layout/vendored_sst/pyproject.toml
vimtor, line 1: what is this `vendored_sst` folder? maybe it was pushed by mistake?

examples/python-layouts/nested-layout/sst.config.ts
```diff
@@ -0,0 +1,84 @@
...
+      timeout: "30 seconds",
+      url: true  // Enable function URL
+    });
+
+    // Worker function in different nested location
+    const worker = new sst.aws.Function("WorkerFunction", {
+      handler: "app/functions/worker/handler.main",
+      runtime: "python3.11",
+      timeout: "5 minutes"
+      // No URL needed for worker function
```
vimtor: can we do a pass to remove these types of useless comments from everywhere?

i believe most of the comments can be removed or made more concise

platform/functions/python-runtime/index.py
vimtor: @subssn21 is this still needed? i from my testing i haven't seen any logs following this format, maybe they are parsed somewhere else?

platform/functions/python-runtime/index.py
```diff
@@ -1,12 +1,22 @@
...
+import logging
 import os
 import sys
 import traceback
 import time
 import requests
 
 
+# Configure Python logging to output to stdout so it appears alongside print statements
+logging.basicConfig(
```
vimtor, line 13: another minor logging issue that i've noticed:

for node.js `sst dev` the logs mention the file without a `./` prefix:

```
|  Invoke      src/index.handler
|  Done        took +664ms
```

but in python we get:

```
|  Invoke      ./functions/src/functions/api.handler
|  Done        took +747ms
```

nothing critical, just mentioning in case it's easy to fix

vimtor (changes requested):

examples/aws-python/sst.config.ts
subssn21: Should be pretty common, but it might make sense, You are referring o when you run sst dev the dependencies aren't there?

vimtor: yep. but it feels common, maybe adding a little mention in the troubleshooting guide

vimtor:
as an update for everyone else, the legend @subssn21 has been working hard to simplify the code and get it production ready

i believe we're really close to merging this

thanks your patience!

examples/aws-python/sst.config.ts
subssn21: I think the:
:::note
 * Builds currently do not tree shake so lots of workspaces can make the build larger than necessary.

Is inaccurate. (It doesn't tree shake like js, but we don't include all workspaces in every lambda, only what's listed in the pyproject.toml. Everything in a workspace will get included though even if not used no treeshaking) so I think the comment is confusing

.github/workflows/check.yml
subssn21: Yes, it's still needed. TestDeploymentWorkflow calls pythonRuntime.Build() which goes through the full build pipeline — that ends up invoking uv build and uv export via the UvCommandRunner in build.go (lines 1486 and 1530). So go test ./... in CI will fail without uv on the PATH.

pkg/runtime/python/python.go
```diff
@@ -242,9 +253,38 @@ func (r *PythonRuntime) Run(ctx context.Context, input *runtime.RunInput) (runti
...
+	for _, ext := range []string{".py", ".toml", ".lock", ".cfg"} {
+		if strings.HasSuffix(file, ext) {
+			return true
+		}
+	}
+	for _, name := range []string{"pyproject.toml", "requirements.txt", "uv.lock", "setup.py", "setup.cfg"} {
+		if filepath.Base(file) == name {
+			return true
+		}
+	}
```
dsonet, line 285: Would using regexp here make it much more efficient?

vimtor: i'm going to simplify this

thanks for pointing out

subssn21:
Latest changes look good and working in all my environments

GringoDotDev:
Just wanted to give a huge thank you to everyone who worked on this. This is amazing!

vimtor:
special shoutout to @subssn21 for putting in the work to make this even better

please @GringoDotDev and everyone, share your experience with the new version. i'm sure we have a lot to improve

PR #6063, opened 2025-08-25
Allow Lambda Function URL on a Router to restrict down to Cloudfront

ReeceXW:
### Problem

At the moment we have two choices:

1. A public function url
2. A private function url but requiring real AWS credentials

We can't restrict down to Cloudfront which means I can't have a URL behind a WAF.

### Solution

Add metadata on the router that exposes the type of URL for the Cloudfront function. We can then enable the no override option on the origin access control configuration.

1. If type is AWS_IAM, we restrict down to Cloudfront
2. If type is AWS_IAM and a permission is set by a user, nothing changes?
3. If type is NONE, OAC still triggers but it's not relevant :)

### DIFF
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -2527,6 +2527,7 @@ export class Function extends Component implements Link.Linkable {
             entries: fnUrl.functionUrl.apply((fnUrl) => ({
               metadata: JSON.stringify({
                 host: new URL(fnUrl).host,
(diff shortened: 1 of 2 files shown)


MrHertal:
Related PR: https://github.com/sst/sst/pull/5581

vimtor:
thanks for your contribution @ReeceXW 

closing in favour of https://github.com/anomalyco/sst/pull/6443

PR #6065, opened 2025-08-26
feat: Add configurable ignore patterns for file watcher

MarkMarine:
This PR adds support for configurable ignore patterns in the SST file watcher to prevent infinite rebuild loops caused by build artifacts.

When using SST with Python projects (or other build systems), package builds create artifacts like `.egg-info` directories and `functions-*.tar.gz` files that trigger SST's file watcher, causing infinite rebuild loops during `sst dev`. Makes local dev not work.

So, this:
  - Added `ignore` field to the App struct in `pkg/project/project.go`
  - Modified file watcher in `cmd/sst/mosaic/watcher/watcher.go` to accept custom ignore patterns
  - Updated `cmd/sst/mosaic.go` to pass ignore patterns from project config
  - Added ignore filtering in AWS function service to prevent rebuilds on ignored files
  - Uses substring matching for pattern detection (e.g., `".egg-info"` matches any path containing `.egg-info`)

  Configure ignore patterns in your `sst.config.ts`:

  ```typescript
  export default $config({
    app(input) {
      return {
        name: "my-app",
        home: "aws",
        ignore: [".egg-info", "functions-"] // Ignore Python build artifacts
      };
    }
  });

I tested:

  - Verified no crash loops with Python build artifacts
  - Confirmed successful deployment and working Lambda endpoints
  - Tested that ignored files don't trigger rebuilds during sst dev

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,24 @@ Learn more about some of the key concepts:
 - [Console](https://sst.dev/docs/console)
 - [Components](https://sst.dev/docs/components)
 
+## Configuration
+
+Your SST app is configured through the `sst.config.ts` file. You can specify app settings like ignore patterns for the file watcher:
+
+```typescript
+export default $config({
+  app(input) {
+    return {
+      name: "my-app",
+      home: "aws",
+      ignore: [".egg-info", "functions-"] // Ignore Python build artifacts
+    };
+  }
+});
+```
+
+See the [configuration reference](https://sst.dev/docs/reference/config) for all available options.
+
 ## Contributing
 
 Here's how you can contribute:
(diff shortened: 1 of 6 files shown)


wilfredjonathanjames:
While this is pending, SST watcher ignores . prefixed files and symlinks: https://github.com/sst/sst/blob/v3.17.23/cmd/sst/mosaic/watcher/watcher.go#L38

This may be acceptable to some.

E.g. Rust runtime "too many open files" is solved by renaming ./target to ./.target: https://github.com/sst/sst/issues/6237#issuecomment-3573027872

fwang:
@MarkMarine catching up on this.. do u mind resolving the conflicts? I can merge after that.

vimtor:
closing in favor of https://github.com/anomalyco/sst/pull/6675

PR #6067, opened 2025-08-28
chore: update pulumi/aws 6.83

natac13:
highest version before major update required


vimtor:
closing in favour of #6259

PR #6069, opened 2025-08-28
Improve renaming command

smith558:
`npx replace-in-file /monorepo-template/g MY_APP **/*.* --verbose` will not work in `zsh` or `fish`, `npx replace-in-file "monorepo-template" "MY_APP" "./**/*" --verbose` will work everywhere


vimtor (approved):

vimtor (changes requested):
some files are not updated:

Image

PR #6087, opened 2025-09-04
Fix NAT EIP associations not using parent provider

### LINKED ISSUE #6086: VPC creation with NAT `ec2` fails in other regions
I'm trying to setup a multi-region ECS deployment. For it I had to create a VPC in multiple regions by doing this:

```ts
const provider = new aws.Provider(name, { region });

const vpc = new sst.aws.Vpc(`Vpc${name}`, { nat: "ec2" }, { provider });

const cluster = new sst.aws.Cluster(
  `Cluster${name}`,
  { vpc, forceUpgrade: "v2" },
  { provider },
);
```

However, this resulted in the following errors: 

```
VpcApSoutheast1NatInstanceEipAssociation1 aws:ec2:EipAssociation sdk-v2/provider2.go:515: sdk.helper_schema: creating EC2 EIP Association: operation error EC2: AssociateAddress, https response error StatusCode: 400, RequestID: cb268286-7c6b-436c-a34c-6e2bf913e6e9, api error InvalidAllocationID.NotFound: The allocation ID 'eipalloc-0aa4e63961b562f49' does not exist: provider=aws@6.66.2

VpcApSoutheast1NatInstanceEipAssociation2
 aws:ec2:EipAssociation                  
sdk-v2/provider2.go:515: sdk.helper_schema: creating EC2 EIP Association: operation error EC2: AssociateAddress, https response error StatusCode: 400, RequestID: 6c6d3447-443a-4bf9-a609-d3790b9abc54, api error InvalidAllocationID.NotFound: The allocation ID 'eipalloc-09568cf40a9b6fa2b' does not exist: provider=aws@6.66.2

VpcApSoutheast1NatInstanceEipAssociation2 aws:ec2:EipAssociation
Association: operation error EC2: AssociateAddress, https response error StatusCode: 400, RequestID: 6c6d3447-443a-4bf9-a609-d3790b9abc54, api error InvalidAllocationID.NotFound: The allocation ID 'eipalloc-09568cf40a9b6fa2b' does not exist
```

vimtor:
Fixes #6086

### DIFF
--- a/platform/src/components/aws/vpc.ts
+++ b/platform/src/components/aws/vpc.ts
@@ -1018,10 +1018,14 @@ export class Vpc extends Component implements Link.Linkable {
               ),
             );
 
-            new ec2.EipAssociation(`${name}NatInstanceEipAssociation${i + 1}`, {
-              instanceId: instance.id,
-              allocationId: elasticIps[i]?.id ?? nat.ip![i],
-            });
+            new ec2.EipAssociation(
+              `${name}NatInstanceEipAssociation${i + 1}`,
+              {
+                instanceId: instance.id,
+                allocationId: elasticIps[i]?.id ?? nat.ip![i],
+              },
+              { parent: self },
+            );
 
             return instance;
           }),
--- a/platform/src/components/aws/vpc.ts
+++ b/platform/src/components/aws/vpc.ts
@@ -20,6 +20,8 @@ import { Vpc as VpcV1 } from "./vpc-v1";
 import { Link } from "../link";
 import { VisibleError } from "../error";
 import { PrivateKey } from "@pulumi/tls";
+import * as pulumi from "@pulumi/pulumi";
+
 export type { VpcArgs as VpcV1Args } from "./vpc-v1";
 
 export interface VpcArgs {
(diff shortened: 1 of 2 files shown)


jamesgibbons92:
This change has a side effect of causing the component to be replaced/recreated for users that already have this deployed. Now I would expect the create to fail as the allocation already exists, but I've just tested it and interestingly the create succeeds and then the deletion of the old components just wipes out the elastic ip associations. So your instances are left un-associated with the elastic ip but the state thinks it exists. If you run a refresh then deploy again it will successfully recreate the association. This seems like a bug in the provider to me. 

image
image

I'm not sure what the best way to navigate this is, but it's potentially a nasty bug/breaking change for anyone relying on the elastic IP addresses on the nat instances.

vimtor:
good catch @jamesgibbons92 

probably that’s why they didn’t include it in the first place. i wonder what the correct solution is here

vimtor:
can you test again with the latest change? @jamesgibbons92

i found a workaround using some pulumi magic

jamesgibbons92:
@vimtor LGTM! I had no idea about `aliases` that's really useful

---- Testing evidence ----
I ran the same steps as above with the update (previous existing VPC), and this time saw no diff / replacements on the EIP association. 
Once deployed I exported the state and the parent correctly migrated from the root to the vpc.
(note: the parent alias will always be root in this case, as there is no option to transform this component - even if there was, the parent would already have been overwritten anyway. so there are no further edge cases to consider)

2025-12-15-101518_hyprshot

Issue #6099, opened 2025-09-11
Raw response in live Hono streaming function with better-auth


pauljasperdev:
I am running a Hono backend better-auth for an Expo client. If I don't disable streaming explicitly, I get a raw session object from `useSession` when running Function live. A deployed streaming Function works fine.

If this is expected behavior, the [Hono streaming example](https://sst.dev/docs/examples/#aws-hono-streaming) should be updated with `streaming: $dev ? false : true,` hint.

### Server

```ts
export const auth = betterAuth({
  database: drizzleAdapter(db, {
    provider: 'pg',
    schema,
  }),
  trustedOrigins: [...],
  secret: Resource.BetterAuthSecret.value,
  emailAndPassword: {
    enabled: true,
  },
  plugins: [expo()],
  ...
});
```

```ts
const app = new Hono();
app.use(logger());

...

app.on(['POST', 'GET'], '/api/auth/**', (c) => auth.handler(c.req.raw));

export const handler = process.env.SST_LIVE ? handle(app) : streamHandle(app);
```

```ts
export const api = new sst.aws.Function('Api', {
  url: {
    router: { instance: routerApi, domain: domainApi },
  },
  streaming: $dev ? false : true,
  // streaming: true,
  handler: 'apps/server/src/index.handler',
});
```

### Client

```ts
export const authClient = createAuthClient({
  baseURL: process.env.EXPO_PUBLIC_SERVER_URL,
  plugins: [
    expoClient({
      storagePrefix: 'immo-app',
      storage: SecureStore,
      scheme: 'immo-app',
    }),
  ],
});
```

```ts
export default function MyComponent() {
  const { data: session, isPending } = authClient.useSession();

  console.log('session', { session });
  console.log('session.user', session?.user);
  ...
  }
```

### Response with `streaming: true` in live

```bash
 LOG  session {"session": {"body": "{\"session\":{\"expiresAt\":\"2025-09-18T06:19:38.413Z\",\"token\":\"W7LNkYe9BLZHgGWYhN8E25DBoBmSp
6XJ\",\"createdAt\":\"2025-09-11T06:19:38.414Z\",\"updatedAt\":\"2025-09-11T06:19:38.414Z\",\"ipAddress\":\"80.146.119.172\",\"userAge
[...]

PR #6100, opened 2025-09-11
docs: update hono streaming example (#6099)

pauljasperdev:
Updated docs of Hono example to disable streaming explicitly in live Function to prevent unexpected behavior.


jayair:
Thanks!

PR #6115, opened 2025-09-17
run uv sync in workspace directory not root directory

severi:
When working with Python containers in AWS Lambda functions, the current implementation runs the `uv sync`  command (and apparently also `uv export` and `uv build`) in the project root.

However, this assumption doesn’t always hold — especially in monorepo setups where the Python service is located in a subdirectory.

Example:

The following project structure does not work, since pyproject.toml is not at the root:

```
.
├── README.md
├── package.json
├── sst.config.ts
├── node-functions
│   └── hello-world
│       ├── index.ts
└── python-services
    └── service-A
        ├── Dockerfile
        ├── main.py
        ├── uv.lock
        ├── sst.pyi
        └── pyproject.toml
 ```
 
In this case, uv sync fails because it’s run from the wrong directory.
This change updates the logic to run `uv sync` from the workspace root instead (i.e. where the closest `pyproject.toml`) file is

### DIFF
--- a/pkg/runtime/python/python.go
+++ b/pkg/runtime/python/python.go
@@ -204,8 +204,13 @@ func (r *PythonRuntime) CreateBuildAsset(ctx context.Context, input *runtime.Bui
 	if arch != "x86_64" && arch != "arm64" {
 		return nil, fmt.Errorf("invalid architecture %q - must be x86_64 or arm64 - %v", arch, string(input.Properties))
 	}
-	workingDir := path.ResolveRootDir(input.CfgPath)
 
+
+	// Get the workspace directory (where pyproject.toml is located)
+	workingDir, err := r.getWorkspaceDirectory(input)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get workspace directory: %v", err)
+	}
 	// 1. Generate non-local package index
 	syncCmd := process.CommandContext(ctx, "uv", "sync", "--all-packages")
 	syncCmd.Dir = workingDir

### CODE REVIEW

**Comment by severi** [comment]
closing this pr as i think the underlying issue would require a more comprehensive solution. My observations from trying to get python services (with and without containers) running in our - otherwise-nodejs- monorepo is  that I would like to have the opportunity to

### 1. Explicitly defining the Python project root

It should be possible to specify the root directory of a Python project directly. For example:
```
    python: {
      rootDir: python-services
    },
    handler: 'service-A/main.handler'
```

This would avoid assumptions about project layout and make monorepo setups more flexible.

### 2. Overriding the default Docker command

Currently, I couldn’t properly use the [AWS Lambda Web Adapter](https://github.com/awslabs/aws-lambda-web-adapter) because I wasn’t able to change the default command used to start the container.

It would be useful to allow something like:
```
    python: {
      container: true,
      rootDir: python-services
      cmd: ["python", "main.py"]
    },
 ```
      
### 3. Dependency tree-shaking (or lack thereof)

According to the docs, dependencies don’t tree-shake during build ([reference](https://github.com/sst/sst/blob/dev/examples/aws-python/sst.config.ts#L14)).

I haven’t verified this, but if true, it means that having one Python function with large dependencies that requires a container would force all Python functions to include those dependencies as well—even if they don’t need them.

A possible solution could be to support multiple independent UV projects in a single repo. That way, projects with heavy dependencies can be isolated from smaller ones. This might be achievable via the rootDir attribute I mentioned earlier.

PR #6118, opened 2025-09-17
fix: ensure StaticSite uploads html after other assets

Karavil:
**Why it matters:** CloudFront served a brand-new `index.html` while the referenced `app.[hash].js` files were still on their way to S3, so early visitors hit 403s and saw a white screen.

```
 deploy start
   ├─▶ (reverse iteration) upload index.html
   │       └─▶ HTML points at app.[hash].js (missing)
   └───── lag ──▶ upload app.[hash].js
                           ↓
                       viewer → S3 403 → sad user
```

**What changed:** we keep user-defined cache semantics, but split `fileOptions` into HTML vs. non-HTML buckets so SST always ships hashed assets before HTML.

```
 configure: [ ...non-html..., ...html... ]
 reverse ▶  [ html-group, non-html-group ]
 reorder ▶  [ non-html-group, html-group ]
 serve   ▶  assets uploaded → index.html uploaded → no 403 gap
```

**Testing:** not run (library change).

### DIFF
--- a/platform/src/components/aws/static-site.ts
+++ b/platform/src/components/aws/static-site.ts
@@ -942,13 +942,29 @@ export class StaticSite extends Component implements Link.Linkable {
           const nonHtmlFileOptions: typeof reversedFileOptions = [];
 
           /**
-           * StaticSite historically replays `fileOptions` in reverse so the
-           * last config wins. That works fine for cache headers, but it means
-           * HTML often syncs *before* hashed JS/CSS when users leave the
-           * defaults in place. We bucket options by whether they target HTML
-           * and run those after everything else, preserving the "last wins"
-           * semantics while ensuring every referenced asset is in S3 before a
-           * new `index.html` ships.
+           * StaticSite historically replays `fileOptions` in reverse order, so
+           * the *last* entry in config becomes the *first* upload pass. With
+           * the default options that put HTML last, the runtime actually pushes
+           * `index.html` to S3 *before* its hashed JS/CSS siblings:
+           *
+           * ```text
+           * deploy start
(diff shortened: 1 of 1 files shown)


vimtor:
thanks for your contribution @Karavil 

i'm closing this in favor of #6289 since your pr didn't actually fixed the issue

Karavil:
> thanks for your contribution @Karavil 
> 
> 
> 
> i'm closing this in favor of #6118 since your pr didn't actually fixed the issue

Hey Vimtor, what do you mean it didn't actually fix the issue?

Karavil:
> thanks for your contribution @Karavil 
> 
> 
> 
> i'm closing this in favor of #6118 since your pr didn't actually fixed the issue

Also could you point me to the PR that you're merging? That points to the current one #6118

vimtor:
sorry, it's this one #6289

vimtor:
@Karavil more context on your pr:

it had a bug because by processing the non-html option first you accidentally include the html files with the first glob `**` thus setting the wrong cache-control header for them

CleanShot 2025-12-18 at 18 38 53@2x

also, as you can see in the linked pr i found another issue that's more likely to be the root cause

Karavil:
> @Karavil more context on your pr:
> 
> 
> 
> it had a bug because by processing the non-html option first you accidentally include the html files with the first glob `**` thus setting the wrong cache-control header for them
> 
> 
> 
> CleanShot 2025-12-18 at 18 38 53@2x
> 
> 
> 
> also, as you can see in the linked pr i found another issue that's more likely to be the root cause

Got it! I appreciate the response and review :)

PR #6133, opened 2025-09-25
fix: kill entire process groups to prevent dev server zombie processes on timeout

sQVe:
### Summary

SST dev command was leaking child processes when timing out. Each timeout left an entire process tree running, eventually exhausting system resources. This fix ensures all spawned processes die with their parents.

#### Background
 
Our team continuously experiences memory exhaustion and database connection leaks from sst dev, even when idling. During normal development work (hot reloads, rebuilds), it spawns new processes without cleaning up old ones, accumulating 200+ processes consuming over 100% system memory in a single dev session.

#### The Bug

When `sst dev` times out (default 50 mins), it prints `[timeout]` but never kills the running process. The timeout handler in [mosaic.go:85-91](https://github.com/sst/sst/blob/9ced491d45eb971842c67e2b87ba43df817bc241/cmd/sst/mosaic.go#L85-L90) just continues the loop without calling `process.Kill()`. When the loop eventually restarts, only the parent dies - children survive as zombies. 🧟 

**Reproduction:**
```bash
#!/bin/bash
node -e "setInterval(() => console.log('child alive'), 2000)" &
wait
```

1. Lower the timeout in `mosaic.go` for debugging.
1. Run above script with `sst dev`
1. Wait for the timeout
1. Check `ps aux | grep node` for accumulating orphaned processes (parent PID is 1)

#### The Fix

1. **Create process groups:** Set `Setpgid: true` for spawned processes on Unix ([mosaic.go:126-130](https://github.com/sst/sst/pull/6133/files#diff-d596e671dba70e5c2adde994d2c251da1f11e48cc3b2179f0a97c1c00e6f4049R126-R130))
1. **Kill entire groups:** Use `syscall.Kill(-pid, signal)` to terminate all processes in group ([process.go:111](https://github.com/sst/sst/pull/6133/files#diff-d821ab8e2837e9704c2828e47ea4558eea087c46e1870d71f51842b5764b633cR111))
1. **Graceful fallback:** If group kill fails, fall back to individual process termination

#### Test Plan

- [x] Run `sst dev` with a script that spawns children, verify all processes die on timeout
- [x] Test Ctrl+C kills entire process tree
- [x] Verify `ps aux` shows no orphaned processes after SST exits
- [ ] Confirm Windows builds work (excluded from Unix-specific changes)
- [x] Check runtime processes (Python/Go/Node/Rust) still terminate correctly

### DIFF
--- a/cmd/sst/mosaic.go
+++ b/cmd/sst/mosaic.go
@@ -90,7 +90,7 @@ func CmdMosaic(c *cli.Cli) error {
 				go func() {
 					evts <- true
 				}()
-				fmt.Println(ui.TEXT_DIM.Render("[timeout]"))
+				fmt.Println("\n"+ui.TEXT_DIM.Render("[timeout]"))
 				timer.Reset(timeout)
 				continue
 			case _, ok := <-evts:
--- a/cmd/sst/mosaic.go
+++ b/cmd/sst/mosaic.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	goruntime "runtime"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/kballard/go-shellquote"
@@ -126,6 +127,11 @@ func CmdMosaic(c *cli.Cli) error {
 					cmd.Env = append(cmd.Env, "FORCE_COLOR=1")
 					for k, v := range nextEnv.Env {
 						cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", k, v))
+					}
+					if goruntime.GOOS != "windows" {
+						cmd.SysProcAttr = &syscall.SysProcAttr{
+							Setpgid: true,
+						}
 					}
 					cmd.Stdin = os.Stdin
 					cmd.Stdout = os.Stdout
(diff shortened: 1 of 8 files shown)


vimtor (changes requested):
the changes work on unix

before:

https://github.com/user-attachments/assets/6c12d717-386b-4d4d-bebf-13f525b9cf6f

after:

https://github.com/user-attachments/assets/32fea16c-ba17-4363-891d-cb9fe276cc78

but sadly the build breaks on windows

```sh
GOOS=windows GOARCH=amd64 go build ./pkg/process/

# github.com/sst/sst/v3/pkg/process
pkg/process/process.go:111:21: undefined: syscall.Kill
pkg/process/process.go:134:21: undefined: syscall.Kill
```

this happens because `syscall.Kill` is not available on windows

@sQVe could you extract the kill process logic into 2 separate files using go build directives

there are similar solutions in the repo, for example:
https://github.com/sst/sst/blob/dev/cmd/sst/mosaic/multiplexer/tcell-term/vt_unix.go
https://github.com/sst/sst/blob/dev/cmd/sst/mosaic/multiplexer/tcell-term/vt_windows.go

sQVe:
@vimtor Cheers for taking a look ❤️. I will look into that ASAP, either tonight or tomorrow! 👍🏼

vimtor (approved):
thanks for your contribution @sQVe 

made some minor changes and added tests

<details>

Image

Image

</details>

PR #6134, opened 2025-09-26
fix: update launchdarkly link to actual provider from sst/provider

mjdiloreto:
The LaunchDarkly provider in [sst/provider](https://github.com/sst/provider/blob/24e18e8a1ab99e94b6b36df387c4e7b66493bd2d/metadata/launchdarkly.toml#L2) is different than the one the docs point to.

Docs here are simply outdated, they point to an archive of an unofficial pulumi implementation.


jayair:
Thank you!

PR #6138, opened 2025-09-30
feat(auth): add router transform support

MrHertal:
**Summary**

  - Add transform property to customize the Router (enables WAF attachment)

**Changes**

  - Router transform: New transform.router prop allows customizing the underlying Router/CloudFront,
  enabling WAF attachment via webAclId.

**Usage**

```
  // WAF attachment
  new sst.aws.Auth("MyAuth", {
    issuer: "src/auth.handler",
    domain: "auth.example.com",
    transform: {
      router: (args) => {
        args.transform = {
          cdn: { transform: { distribution: { webAclId: "..." } } }
        };
      }
    }
  });
```


platform/src/components/aws/auth.ts
```diff
@@ -304,14 +390,21 @@ export class Auth extends Component implements Link.Linkable {
...
+      // Apply transform if provided
+      if (args.transform?.router) {
+        if (typeof args.transform.router === "function") {
+          args.transform.router(routerArgs, {}, `${name}Router`);
+        } else {
+          Object.assign(routerArgs, args.transform.router);
+        }
+      }
+
+      const router = new Router(`${name}Router`, routerArgs, { parent: self });
```
vimtor: transforms are usually applied using the `transform` utility

any reason not to do it here?

vimtor (changes requested):
thanks for the contribution @MrHertal 

i just left some minor suggestions

platform/src/components/aws/auth.ts
```diff
@@ -264,14 +343,19 @@ export class Auth extends Component implements Link.Linkable {
...
     function createTable() {
-      return new Dynamo(
-        `${name}Storage`,
-        {
-          fields: { pk: "string", sk: "string" },
-          primaryIndex: { hashKey: "pk", rangeKey: "sk" },
-          ttl: "expiry",
-        },
-        { parent: self },
+      if (args.table) {
```
vimtor: if it isn't too hard, could we throw a `VisibleError` if the input table has the required schema?

platform/src/components/aws/auth.ts
MrHertal: no reason, my bad!
I applied the changes: https://github.com/sst/sst/pull/6138/commits/d6db2510a7ce0c6d30a5035cfd00c3feb606df62

platform/src/components/aws/auth.ts
MrHertal: Good catch, done: https://github.com/sst/sst/pull/6138/commits/8a8beef25e0c6eebcda8b483cca0f6ef2ec3d627

platform/src/components/aws/auth.ts
```diff
@@ -264,14 +344,40 @@ export class Auth extends Component implements Link.Linkable {
...
```
vimtor: when working with pulumi inputs it's usually better not to do checks like this because `args.table` could be a promise or output that resolves to null for example

it's better practice to do something like this:

```ts
    function createTable() {
      return output(args.table).apply((table) => {
        if (table) {
          return all([
            table.nodes.table.hashKey,
            table.nodes.table.rangeKey,
            table.nodes.table.ttl,
          ]).apply(([hashKey, rangeKey, ttl]) => {
            if (hashKey !== "pk")
              throw new VisibleError(
                `The provided table must have "pk" as the hash key, but got "${hashKey}".`,
              );

            if (rangeKey !== "sk")
              throw new VisibleError(
                `The provided table must have "sk" as the range key, but got "${rangeKey}".`,
              );

            if (ttl?.attributeName !== "expiry")
              throw new VisibleError(
                `The provided table must have TTL enabled on the "expiry" attribute, but got "${ttl?.attributeName}".`,
              );

            return table;
          });
        }

        return output(
          new Dynamo(
            `${name}Storage`,
            {
              fields: { pk: "string", sk: "string" },
              primaryIndex: { hashKey: "pk", rangeKey: "sk" },
              ttl: "expiry",
            },
            { parent: self },
          ),
        );
      });
    }
```

https://github.com/sst/sst/pull/5908 had a good example on why you should avoid this

vimtor (changes requested):

platform/src/components/aws/auth.ts
MrHertal: Thanks for the advice, I've applied the changes: https://github.com/sst/sst/pull/6138/commits/a9184146cacdecdcdd533f0b535f8dec9b0936cf

vimtor:
one last thing @MrHertal

what happens if someone links a table to an existing auth instance? would it wipe the user data?

maybe we should add a warning in the docs

MrHertal:
> one last thing @MrHertal
> 
> what happens if someone links a table to an existing auth instance? would it wipe the user data?
> 
> maybe we should add a warning in the docs

@vimtor True, Pulumi would delete the original table and all user data would be lost. I've added a warning: https://github.com/sst/sst/pull/6138/commits/458e93278c783e2bac8f17654e5f1c4ce28ddd28

vimtor (commented):
hey @MrHertal happy new year!

i'm coming back to this. after some thought, i'm not sure whether exposing the table argument is a good idea. i feel like some people could shoot themselves in the foot. what is the use case? is it linking the underlying table? because if that's the case, maybe you can do it via `auth.nodes.table`

what do you think? can we just leave the transform for now and open a separate pr/issue to discuss the table part?

MrHertal:
Hey @vimtor, happy new year to you too!

You make a good point. My original use case was for multi-region deployments with a single replicated DynamoDB table, but I agree there's potential for users to accidentally lose data when switching from an auto-created table to a custom one.

I've removed the table prop and kept only the transform.router feature for now. We can discuss the custom table functionality separately if there's enough demand for it - perhaps with better safeguards or a different approach.

PR #6148, opened 2025-10-03
docs: fixed opensearch static get example to use id instead of arn

eric-k-chu:
Updated opensearch get example to use the id instead of the arn


jayair:
Thanks!

PR #6163, opened 2025-10-10
docs(blog): update links in `blog/moving-away-from-cdk` to point to v2

neo:
I was reading this blog post (that's filled with great insights) and found the links are supposed to point to the docs from v2


jayair:
Appreciate it!

PR #6165, opened 2025-10-11
unwrap Input<string[]> before spreading in server install config

danielfyhr:
release is [broken](https://github.com/sst/sst/actions/runs/18287399258/job/52065926413)

```
Run cd platform && bun tsc --noEmit
src/components/aws/ssr-site.ts(1137,24): error TS2488: Type 'Input<string[]>' must have a '[Symbol.iterator]()' method that returns an iterator.
error: "tsc" exited with code 2
```

this pr maybe fixes that:
changed from direct spread of args.server?.install to using output(args.server).apply() to unwrap
the Input type first like its done repeatedly in that file

### DIFF
--- a/platform/src/components/aws/ssr-site.ts
+++ b/platform/src/components/aws/ssr-site.ts
@@ -1133,10 +1133,10 @@ async function handler(event) {
                 nodejs: {
                   ...planServer.nodejs,
                   format: "esm" as const,
-                  install: [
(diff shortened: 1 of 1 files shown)


danielfyhr:
this is fixed in 73e3d01cbecfec1fdb7aa34f771cda77dd03f1e8

PR #6169, opened 2025-10-14
Add QueueConsumer component for Cloudflare Queues

bachiitter:
This PR adds `QueueConsumer` component for Cloudflare Queues allowing user to separate queue message processing into dedicated workers.

## Example Usage
```ts
const queue = new sst.cloudflare.Queue("MyQueue");

new sst.cloudflare.QueueConsumer("MyConsumer", {
  queueId: queue.id,
  processor: "consumer.ts"
});

### CODE REVIEW

**Review by vimtor — COMMENTED** [review]
thanks for your contribution @bachiitter 

i'm thinking that we should probably follow the same API as the AWS `queue.subscribe()` instead of exposing the queue consumer directly

also, it would be amazing if you could add an example in the `examples/` directory

what do you think?

**Comment by bachiitter** [comment]
> thanks for your contribution @bachiitter
> 
> i'm thinking that we should probably follow the same API as the AWS `queue.subscribe()` instead of exposing the queue consumer directly
> 
> also, it would be amazing if you could add an example in the `examples/` directory
> 
> what do you think?

will create an example, queue.subscribe() seems better though i don't have much experience so what ever fits better

**Comment by vimtor** [comment]
yep, i prefer `queue.subscribe` since it's more consistent with what we already have

**Comment by bachiitter** [comment]
@vimtor updated to queue.subscribe and also added an example

File: examples/cloudflare-queue/producer.ts [hunk 1]
```diff
@@ -0,0 +1,8 @@
+import { Resource } from "sst";
+
+export default {
+  async fetch(request, env) {
+    await env.MyQueue.send({ hello: "world" });
```
vimtor: shouldn't we use this instead?

```suggestion
    await Resource.MyQueue.send({ hello: "world" });
```

platform/src/components/cloudflare/queue.ts
```diff
@@ -49,6 +129,96 @@ export class Queue extends Component implements Link.Linkable {
...
+   * });
+   * ```
+   */
+  public subscribe(
+    subscriber: Input<string | WorkerArgs>,
+    args?: QueueSubscribeArgs,
+    opts?: ComponentResourceOptions,
+  ) {
+    if (this.isSubscribed) {
+      throw new Error(
```
vimtor: we should throw a `VisibleError` type so it gets printed

```suggestion
      throw new VisibleError(
```

platform/src/components/cloudflare/queue.ts
```diff
@@ -49,6 +129,96 @@ export class Queue extends Component implements Link.Linkable {
...
+            queueId: this.queue.id,
+            scriptName: worker.script.scriptName,
+            settings: args?.settings,
+            type: "worker",
+          },
+          { parent, ...opts },
+        ),
+      );
+
+      return { worker, consumer };
```
vimtor: not sure about this return type

to be consistent with the aws queue we should create a `QueueWorkerSubscriber` component and return that instead

vimtor (changes requested):
looking great @bachiitter!

i've left a couple of comments with some minor changes

we should be ready to merge this after fixing them

bachiitter:
> looking great @bachiitter!
> 
> i've left a couple of comments with some minor changes
> 
> we should be ready to merge this after fixing them

just copy example from another test project will update it and will make other fixes

vimtor (commented):
<details>

<summary>your code works like a charm @bachiitter</summary>

Image

Image

</details>

however it gives the following error when removing the deployment:

```
MyQueueSubscriber sst:cloudflare:QueueWorkerSubscriber → MyQueueSubscriberConsumer cloudflare:index:QueueConsumer
failed to make http request: missing required consumer_id parameter
```

this comes from the cloudflare provider https://github.com/cloudflare/terraform-provider-cloudflare/issues/6405

so we'll have to wait for it to get fixed before merging

bachiitter:
> your code works like a charm @bachiitter
> Image Image
> however it gives the following error when removing the deployment:
> 
> ```
> MyQueueSubscriber sst:cloudflare:QueueWorkerSubscriber → MyQueueSubscriberConsumer cloudflare:index:QueueConsumer
> failed to make http request: missing required consumer_id parameter
> ```
> 
> this comes from the cloudflare provider [cloudflare/terraform-provider-cloudflare#6405](https://github.com/cloudflare/terraform-provider-cloudflare/issues/6405)
> 
> so we'll have to wait for it to get fixed before merging

another thing added to list of thing cloudflare needs fixing

platform/src/components/cloudflare/queue.ts
```diff
@@ -18,17 +21,95 @@ export interface QueueArgs {
   };
 }
 
+export interface QueueSubscribeArgs {
+  /**
+   * The dead letter queue to send messages that fail processing.
+   */
+  deadLetterQueue?: Input<string>;
```
vimtor: shouldn't the dlq be an arg to the queue itself?

platform/src/components/cloudflare/queue.ts
```diff
@@ -18,17 +21,95 @@ export interface QueueArgs {
...
+  /**
+   * Consumer settings like batch size, retries, and visibility timeout.
+   */
+  settings?: cloudflare.QueueConsumerArgs["settings"];
```
vimtor: could we directly expose the settings ourselves? similar to how we do it in the aws queue? with nice short names and defaults. ideally we will mirror the same field structure

examples/cloudflare-queue/bun.lock
vimtor, line 1: could you remove this file?

(sorry this should be ignored by default)

vimtor (changes requested):
@bachiitter good news! with the latest cloudflare provider the removing issue no longer happens

i left a couple of minor comments, once resolved we should be good to merge

platform/src/components/cloudflare/queue.ts
bachiitter: it's on consumer
https://www.pulumi.com/registry/packages/cloudflare/api-docs/queueconsumer/

platform/src/components/cloudflare/queue.ts
bachiitter: sure we can do that, will look at aws queue

examples/cloudflare-queue/bun.lock
bachiitter: > could you remove this file?
> 
> (sorry this should be ignored by default)

sure sorry my bad

platform/src/components/cloudflare/queue.ts
vimtor: i see

given we can only have one consumer, i'll probably still put it in the queue just to match the existing aws queue component

what do you think? open to suggestions here

vimtor:
@bachiitter forgot to mention that we should surface it in the docs in this pr as well

platform/src/components/cloudflare/queue.ts
vimtor: @bachiitter thoughts?

.gitignore
```diff
@@ -3,6 +3,7 @@ dist
 node_modules
 Pulumi.*.yaml
 Pulumi.yaml
+bun.lock
```
vimtor: sorry we shouldn't ignore this i misspoke 

we should only ignore it for the example 

don't worry i will fix it in a separate pr

examples/cloudflare-queue/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "cloudflare-queue",
+  "version": "0.0.0",
+  "dependencies": {
+    "sst": "^3.1.49"
```
vimtor: can we set it to ^4 like the other examples?

platform/src/components/cloudflare/queue-worker-subscriber.ts
```diff
@@ -0,0 +1,196 @@
...
+   */
+  subscriber: Input<string | WorkerArgs>;
+  /**
+   * The dead letter queue to send messages that fail processing.
+   */
+  dlq?: {
+    /**
+     * The name of the dead letter queue.
+     */
+    queue?: Input<string>;
```
vimtor: when dlq is provided the queue name should be required

platform/src/components/cloudflare/queue.ts
```diff
@@ -18,17 +22,172 @@ export interface QueueArgs {
...
 
+export interface QueueSubscribeArgs {
+  /**
+   * The dead letter queue to send messages that fail processing.
+   */
+  dlq?: {
+    /**
+     * The name of the dead letter queue.
+     */
+    queue?: Input<string>;
```
vimtor: same thing here

www/astro.config.mjs
```diff
@@ -217,6 +217,8 @@ const sidebar = [
           "docs/component/cloudflare/cron",
           "docs/component/cloudflare/d1",
           "docs/component/cloudflare/kv",
+          "docs/component/cloudflare/queue",
+          "docs/component/cloudflare/static-site",
```
vimtor: let's remove this one for the moment

i've created a separate issue of it

vimtor (changes requested):

platform/src/components/cloudflare/queue.ts
bachiitter: sure we can do that only problem is we can't do that for the queue-worker-subscriber

platform/src/components/cloudflare/queue-worker-subscriber.ts
bachiitter: got it

platform/src/components/cloudflare/queue.ts
bachiitter: will fix it

www/astro.config.mjs
bachiitter: got it

.gitignore
bachiitter: got it

Issue #6156, opened 2025-10-08
Aurora PG Serverless Version mismatch


ericmillsio:
```
Postgres sst:aws:Aurora → PostgresCluster aws:rds:Cluster
sdk-v2/provider2.go:515: sdk.helper_schema: updating RDS Cluster (xxxx-xxxx-postgrescluster-xxxxxxxxx): operation err
or RDS: ModifyDBCluster, https response error StatusCode: 400, RequestID: dab60308-8b77-45bb-a3f3-8e7c0a309d27, api err
or InvalidParameterCombination: Cannot upgrade aurora-postgresql from 16.8 to 16.6: provider=aws@6.66.2
```

```
export const db = new sst.aws.Aurora('Postgres', {
  vpc,
  engine: 'postgres',
  version: '16.6',
  scaling: {
    min: '0 ACU',
    max: '1 ACU',
    pauseAfter: '1 hour',
  },
});
```

I'm getting an error of a failed upgrade of Aurora serverless when I have a fixed version for some reason. I also have different staging DBs that ended up in the newer versions, when IMO it should stay fixed to that version (unless AWS doesn't allow new DBs in old versions?).

Another user reports a similar issue on Discord:

"yeah, i get the same. i had to go into the instance in aws console and manually update it to the target version, matching the version change in my sst config. after that, my next sst deploy worked. this only works for minor updates tho. 
i need to figure out how to upgrade the major version.
ideally it could all just go via the sst config, but i tested that out, and it failed for me."

vimtor:
if you are okay with Postgres version `16.8` you can get rid of the error by changing the `version` field to `16.8`

this error probably occurs because AWS setting for auto upgrading minor version is set to `true` by default

i've opened a PR to fix this error for future users

ericmillsio:
Yeah I'm forcing things through with an upgrade for now. I have other stages that are still behind with other devs though, so will still be important for future version changes. Thanks!

AlfieJones:
You can also set the version without the minor version number, e.g `version: ’16’`, which still allows AWS to auto upgrade

vimtor:
probably fixed by #6259

PR #6170, opened 2025-10-14
Disable `autoMinorVersionUpgrade` for AWS databases

vimtor:
AWS databases by default auto upgrade minor versions

this ends up causing a mismatch with the Pulumi specified version

leading to confusion like in #6156


vimtor:
closing in favor of #6207

PR #6174, opened 2025-10-15
Fix Remix unused '@ts-expect-error' directive

### LINKED ISSUE #6168: Unused '@ts-expect-error' directive
The latest sst (3.17.19) has a typecheck error:

.sst/platform/src/components/aws/remix.ts:481:13 - error TS2578: Unused '@ts-expect-error' directive.

vimtor:
closes #6168 #6182

this error happens because `vite` gets installed when using Remix

making it available and triggering the `@ts-expect-error` unused warning

### DIFF
--- a/platform/src/components/aws/remix.ts
+++ b/platform/src/components/aws/remix.ts
@@ -478,7 +478,7 @@ export class Remix extends SsrSite {
           if (!file) return;
 
           try {
-            // @ts-expect-error
+            // @ts-ignore
             const vite = await import("vite");
             const config = await vite.loadConfigFromFile(
               { command: "build", mode: "production" },


dani-mp:
@vimtor thanks! hopefully they can get in soon!

platform/src/components/aws/remix.ts
```diff
@@ -478,7 +478,7 @@ export class Remix extends SsrSite {
           if (!file) return;
 
           try {
-            // @ts-expect-error
+            // @ts-ignore
```
dani-mp, line 481: the error that i get on my side is
```
Error: .sst/platform/src/components/aws/remix.ts(481,13): error TS2578: Unused '@ts-expect-error' directive.
```
so should we expect an error there at all?

PR #6176, opened 2025-10-16
fix: add 'immutable' suffix to svelte-kit app plan assets

### LINKED ISSUE #6040: SvelteKit Remote Functions: Not working on lambda due to CloudFront KVS
### EDIT: 
I overlooked the fact that it's a devDependency, apologies! 

The issue still persists, though. My current assumption is that lies with how entries in the KeyValueStore are created. 

Here is what I _believe_ is happening:

1. SvelteKit bundles the remote.ts files into the client and in the server modules - i guess so you can import them on both sides?
2. SST creates a KV entry for all files in the client and prerendered modules.
3. because SST has created a KV entry, CloudFront takes precedence in the request and tries to fetch the file from S3 - which actually should return something, i guess... - because it was bundled
4. The request from client is never passed through to the server because CloudFront intercepted it.

The CloudFlare adapter for SvelteKit seems to have solved this by mapping routes individually, instead of a catch-all approach for anything not in client/ prerendered. 

I will do my best to provide another update as I try to pinpoint the exact issue.

--------

**Context**

The svelte-kit-sst adapter currently uses 

```
"@sveltejs/kit": "^1.5.0"
````

since then SvelteKit had a major release and is as of Aug 13, 2025 at v2.29.0.

**Request**
SvelteKit now includes a built-in RPC implementation (still experimental but soon to be released for production). Please bump the dependency version for SvelteKit in the adapter from ^1.x.x to ^2.x.x

**Breaking Changes**
The breaking changes from v1 to v2 primarily affect application code (routes, hooks, load functions). There are no changes to how network requests are handled, thus the adapter implementation should not need any modification.

### PR DESCRIPTION
**Pull request by weston-harper-rox**
resolves #6040

SvelteKit's new remote functions feature does not work with the current default configuration of SST's `SvelteKit` component assets plan. The issue is related to the handler function errantly routing requests sent by SvelteKit to `_app/remote` to the client-side static asset handler.

The community has found a workaround, which works when modified manually in CloudFront. However, when a router change is deployed, the KV store is updated, which negates all prior manual changes. Additionally, consumers should expect the SvelteKit component to work without modification.

I followed the pattern that resolves a similar issue in NextJS apps and used deep routing within SvelteKit's `_app` directory. I've added a new demo app that uses experimental features and have verified that deployments work as expected when built using the change.

### CODE REVIEW

**Comment by weston-harper-rox** [comment]
I've done some digging in the SvelteKit repo and I'd like to test this setup with prerendered pages and other static assets to make sure it all works. I'll create a new SvelteKit demo app with experimental features turned on.

File: platform/src/components/aws/svelte-kit.ts [hunk 1]
```diff
@@ -474,6 +474,7 @@ export class SvelteKit extends SsrSite {
             to: "",
             cached: true,
             versionedSubDir: "_app",
+            deepRoute: "_app",
```
weston-harper-rox, line 477: This is the entirety of the fix.

vimtor (approved):
approved!

before:

Image

after:

Image

i've also tried deploying the existing Sveltekit example and it worked

leon:
Any news on when this could get merged?

vimtor:
> Any news on when this could get merged?

this week

PR #6186, opened 2025-10-21
Make Rust SDK `SST_KEY` environment variables optional

### LINKED ISSUE #6161: SST_KEY environment variable isn't set while working with SDK
I am using the Rust SST SDK and `Resource::init()` returns `ResourceError::EnvError(NotPresent)` because it [can't find the environment variable  `SST_KEY`](https://github.com/sst/sst/blob/a0a12d15ce5d363768a089bdf0678a5b4e7b98e1/sdk/rust/src/lib.rs#L33). I'm not sure why the variable isn't being set, but even `sst shell env | grep '^SST.*'` only shows the SST resource and no sign of `SST_KEY` nor `SST_KEY_FILE`. It looks like the absence of `SST_KEY`/`SST_KEY_FILE` is fine for the [Python SDK](https://github.com/sst/sst/blob/a0a12d15ce5d363768a089bdf0678a5b4e7b98e1/sdk/python/src/sst/__init__.py#L23) and [Javascript SDK](https://github.com/sst/sst/blob/a0a12d15ce5d363768a089bdf0678a5b4e7b98e1/sdk/js/src/resource.ts#L29), but not for the [Go SDK](https://github.com/sst/sst/blob/a0a12d15ce5d363768a089bdf0678a5b4e7b98e1/sdk/golang/resource/resource.go#L18). Should this behavior differs between the SDKs?

I am using SST version 3.17.14 with NodeJS 24.

vimtor:
closes #6161

the Rust SDK behaves differently than the JS SDK because it required the `SST_KEY` and `SST_KEY_FILE` env vars to be present

this PR makes them optional

i also added some tests to prevent this issue in the future

### DIFF
--- a/sdk/rust/Cargo.lock
+++ b/sdk/rust/Cargo.lock
@@ -44,6 +44,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
+name = "bitflags"
+version = "2.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "812e12b5285cc515a9c72a5c1d3b6d46a19dac5acfef5265968c166106e31dd3"
+
+[[package]]
 name = "cfg-if"
 version = "1.0.0"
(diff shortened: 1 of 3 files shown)


wilfredjonathanjames:
Can we get this merged?

PR #6205, opened 2025-10-30
Support providing a package name that is different from the provider name

### LINKED ISSUE #6203: Provider configuration doesn't work with non-@pulumi/@sst-provider, etc packages
When using a package that has to be fully named in the providers section (e.g. `@paynearme/pulumi-jetstream`) as follows:

```
app(input) {
    return {
      name: "test-project",
      home: "aws",
      providers: {
        "@paynearme/pulumi-jetstream": {
          servers: "localhost:4222"
        },
      },
    };
  },
```

The config that gets sent to pulumi uses the full provider name rather than the pulumi specific namespace (e.g. `@paynearme/pulumi-jetstream:servers=localhost:4222` instead of `jetstream:servers:localhost:4222`).

The problem seems to be centered around https://github.com/sst/sst/blob/291039bcd81ca3b13bf1124502b66acef2f459ae/pkg/project/run.go#L291-L307.

yoerayo:
Fixes https://github.com/sst/sst/issues/6203 by adding a `package` configuration option for `providers` that allows specifying the NPM package name for the provider.

### DIFF
--- a/cmd/sst/main.go
+++ b/cmd/sst/main.go
@@ -591,7 +591,7 @@ var root = &cli.Command{
 						return err
 					}
 				}
-				entry, err := project.FindProvider(pkg, pkg, "latest")
+				entry, err := project.FindProvider(pkg, "latest", pkg)
 				if err != nil {
 					return util.NewReadableError(err, "Could not find provider "+pkg)
 				}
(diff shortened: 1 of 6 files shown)


vimtor (approved):
thanks for your contribution @yoerayo 

had to do some minor changes but everything works like a charm

<details>

Image

Image

</details>

vimtor:
<details>

<summary>also verified the aws v6 to 7 upgrade path</summary>

CleanShot 2026-03-02 at 13 44 07@2x

</details>

PR #6207, opened 2025-11-02
Update default Postgres major version to 17

### LINKED ISSUE #6195: Can't create Postgres db
I am trying to deploy a pg db with:
```typescript
new sst.aws.Postgres("PanelsDB", {
  vpc,
  proxy: true,
})
```
However I am getting the following error:
```
sdk-v2/provider2.go:515: sdk.helper_schema: creating RDS DB Instance (panels-production-panelsdbinstance-mrkntvtc): operation error RDS: CreateDBInstance, http
s response error StatusCode: 400, RequestID: 6f1854cf-1594-44c2-9404-52656491f99b, api error InvalidParameterCombination: Cannot find version 16.4 for postgres
: provider=aws@6.66.2
```
I'm using the latest sst version (3.17.21)

vimtor:
closes #6195 #6156

### DIFF
--- a/platform/src/components/aws/aurora.ts
+++ b/platform/src/components/aws/aurora.ts
@@ -40,7 +40,7 @@ export interface AuroraArgs {
   /**
    * The version of the Aurora engine.
    *
-   * The default is `"16.4"` for Postgres and `"3.08.0"` for MySQL.
+   * The default is `"17"` for Postgres and `"3.08.0"` for MySQL.
    *
    * Check out the [available Postgres versions](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Concepts.Aurora_Fea_Regions_DB-eng.Feature.ServerlessV2.html#Concepts.Aurora_Fea_Regions_DB-eng.Feature.ServerlessV2.apg) and [available MySQL versions](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Concepts.Aurora_Fea_Regions_DB-eng.Feature.ServerlessV2.html#Concepts.Aurora_Fea_Regions_DB-eng.Feature.ServerlessV2.amy) in your region.
    *
@@ -55,11 +55,11 @@ export interface AuroraArgs {
    * - Aurora PostgresSQL 13.15 and higher
    * - Aurora MySQL 3.08.0 and higher
    *
-   * @default `"16.4"` for Postgres, `"3.08.0"` for MySQL
+   * @default `"17"` for Postgres, `"3.08.0"` for MySQL
    * @example
    * ```js
    * {
-   *   version: "16.3"
+   *   version: "17.3"
    * }
    * ```
    */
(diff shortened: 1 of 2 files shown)


arjunyel:
Since Postgres 18 is out now, probably worth changing the default version to 17

vimtor:
i thought about it @arjunyel but it's a bit trickier than it seems

we don't want to change the default for existing users who don't have the `version` specified

ideally we only apply this change to people who are on the `16.x` version but i'm not sure what's the best way to do that

let me investigate a bit

vimtor:
we could do something like "if the component already exists use 16 otherwise use 17"

the problem is it gets messy when going 17 to 18, 18 to 19, etc

probably better to merge this PR since it fixes the most immediate issues

long term i'd suggest making the `version` argument required

jamesgibbons92:
To not impact current users deployed instances, you could pass engineVersion into the [ignoreChanges](https://www.pulumi.com/docs/iac/concepts/resources/options/ignorechanges/#resource-option-ignorechanges) prop in the underlying pulumi rds instance if no version has been explictly set. That way if users upgrade sst and have not set the version prop, they will stay on 16 or whatever the default was when they first deployed it.

vimtor:
this is what i've tested:

1. create a `Postgres` with the existing sst and the default `version`

```
const vpc = new sst.aws.Vpc("MyVpc", { bastion: true });

const postgres = new sst.aws.Postgres("MyDatabase", {
  vpc,
});
```

CleanShot 2025-12-06 at 10 48 55@2x


2. switch to my branch and run `sst dev` again. no changes are detected

CleanShot 2025-12-06 at 10 51 42@2x

3. create a new Postgres database from my branch

CleanShot 2025-12-06 at 14 42 28@2x

other notes:
- i set the default postgres version to 17 since it's current default in AWS. thankfully the current setup lets us keep upgrading the default version without breaking existing setups
- i tested the same changes with the Aurora component to solve the linked issue. i'll wait a few weeks to see if any unexpected issues arise. if none show up i'll open a new pr to apply this versioning trick to all the components (redis, mysql, etc.) so we can keep them up to date too
- if you are wondering why didn't i test a major version upgrade, it's because it is currently broken #6261

platform/src/components/aws/aurora.ts
```diff
@@ -989,7 +992,10 @@ Listening on "${dev.host}:${dev.port}"...`,
               ...(proxy ? { "sst:ref:proxy": proxy.id } : {}),
             })),
           },
-          { parent: self },
+          {
+            parent: self,
+            ignoreChanges: args.version ? [] : ["engineVersion"],
```
jamesgibbons92, line 997: Having thought about this more, and maybe I am being overly cautious, but this is potentially still a 'breaking' change - as if the user has passed in their own `ignoreChanges` into the rds instance `transforms`, then this prop will be overwritten as they are not merged, and so the users database would upgrade in that scenario. 
I wonder if it makes sense to make sure the ignoreChanges here are merged with the users opts.ignoreChanges transforms ...

vimtor: good catch!

luckily major upgrades are disabled by default until we do #6261 so in this case it would just error

maybe it makes sense to update the `transform` utility directly? i mean, if a component has `ignoreChanges` defined, the user shouldn't be able to override it

PR #6209, opened 2025-11-03
feat(bucket): add lifecycle config to built-in bucket component

mkilp:
Hi guys,

initially i just thought about adding docs with an example (https://github.com/sst/sst/pull/6206) but after thinking some more I feel like this is such a common use case that it warrants first party support.

This PR adds just that.

In this basic version I expose expiration and filters, not the full blown [pulumi lifecycle rules](https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketlifecycleconfigurationv2/#bucketlifecycleconfigurationv2rule) for simplicity and ease of configuration. 

Happy to expand / iterate on everything if you think its required.

Cheers ✌️

### DIFF
--- a/platform/src/components/aws/bucket.ts
+++ b/platform/src/components/aws/bucket.ts
@@ -10,7 +10,7 @@ import { Component, Prettify, Transform, transform } from "../component";
 import { Link } from "../link";
 import type { Input } from "../input";
 import { FunctionArgs, FunctionArn } from "./function";
-import { Duration, toSeconds } from "../duration";
+import { Duration, DurationDays, toSeconds } from "../duration";
 import { VisibleError } from "../error";
 import { parseBucketArn } from "./helpers/arn";
 import { BucketLambdaSubscriber } from "./bucket-lambda-subscriber";
@@ -97,6 +97,51 @@ interface BucketCorsArgs {
    * ```
    */
   maxAge?: Input<Duration>;
+}
+
+interface BucketLifecycleArgs {
+  /**
+   * An S3 object key prefix that the lifecycle rule applies to.
+   * @example
+   * Applies to all the objects in the `images/` folder.
+   * ```js
+   * {
+   *   prefix: "images/"
+   * }
+   * ```
+   */
+  prefix?: Input<string>;
+  /**
+   * Whether the lifecycle rule is enabled.
+   * @example
+   * ```js
+   * {
+   *  enabled: true
+   * }
+   * ```
+   * @default `true`
+   */
+  enabled?: Input<boolean>;
+  /**
+   * Days after which the objects in the bucket should expire.
+   * @example
+   * ```js
+   * {
+   *  expiresIn: "30 days"
+   * }
+   * ```
+   */
+  expiresIn?: Input<DurationDays>;
+  /**
+   * Date after which the objects in the bucket should expire. Defaults to midnight UTC time.
+   * @example
+   * ```js
+   * {
+   *  expiresAt: "2023-08-22"
+   * }
+   * ```
+   */
+  expiresAt?: Input<string>;
 }
 
 export interface BucketArgs {
@@ -385,6 +430,22 @@ export interface BucketArgs {
    */
   cors?: Input<false | Prettify<BucketCorsArgs>>;
   /**
+   * The lifecycle configuration for the bucket.
+   * @example
+   * Delete objects in the "/tmp" directory after 30 days.
+   * ```js
+   * {
+   *   lifecycle: [
+   *     {
+   *       prefix: "/tmp",
+   *       expiresIn: "30 days"
+   *     }
+   *   ]
+   * }
+   * ```
+   */
+  lifecycle?: Input<Input<Prettify<BucketLifecycleArgs>>[]>;
+  /**
    * Enable versioning for the bucket.
    *
    * Bucket versioning enables you to store multiple versions of an object, protecting
@@ -420,6 +481,10 @@ export interface BucketArgs {
      * Transform the S3 Bucket versioning resource.
      */
     versioning?: Transform<s3.BucketVersioningV2Args>;
+    /**
+     * Transform the S3 Bucket lifecycle resource.
+     * */
+    lifecycle?: Transform<s3.BucketLifecycleConfigurationV2Args>;
     /**
      * Transform the public access block resource that's attached to the Bucket.
      *
@@ -783,6 +848,7 @@ export class Bucket extends Component implements Link.Linkable {
     const publicAccessBlock = createPublicAccess();
     const policy = createBucketPolicy();
     createCorsRule();
+    createLifecycle();
 
     // Ensure the policy is created when the bucket is used in another component
     // (ie. bucket.name). Also, a bucket can only have one policy. We want to ensure
@@ -847,6 +913,39 @@ export class Bucket extends Component implements Link.Linkable {
               versioningConfiguration: {
                 status: "Enabled",
               },
+            },
+            { parent },
+          ),
+        );
+      });
+    }
+
+    function createLifecycle() {
+      return output(args.lifecycle).apply((lifecycleRules) => {
+        if (!lifecycleRules || lifecycleRules.length === 0) return;
+
+        return new s3.BucketLifecycleConfigurationV2(
+          ...transform(
+            args.transform?.lifecycle,
+            `${name}Lifecycle`,
+            {
+              bucket: bucket.bucket,
+              rules: lifecycleRules.map((rule, index) => ({
+                id: `${name}LifecycleRule${index}`,
+                status: rule.enabled !== false ? "Enabled" : "Disabled",
+                expiration:
+                  rule.expiresIn || rule.expiresAt
+                    ? {
+                        days: rule.expiresIn
+                          ? toSeconds(rule.expiresIn) / 86400
+                          : undefined,
+                        date: rule.expiresAt
+                          ? `${rule.expiresAt}T00:00:00Z`
+                          : undefined,
+                      }
+                    : undefined,
+                filter: rule.prefix ? { prefix: rule.prefix } : undefined,
+              })),
             },
             { parent },
           ),
--- a/platform/src/components/component.ts
+++ b/platform/src/components/component.ts
@@ -165,6 +165,7 @@ export class Component extends ComponentResource {
               "aws:s3/bucketPolicy:BucketPolicy",
               "aws:s3/bucketPublicAccessBlock:BucketPublicAccessBlock",
               "aws:s3/bucketVersioningV2:BucketVersioningV2",
+              "aws:s3/bucketLifecycleConfigurationV2:BucketLifecycleConfigurationV2",
               "aws:s3/bucketWebsiteConfigurationV2:BucketWebsiteConfigurationV2",
               "aws:secretsmanager/secretVersion:SecretVersion",
               "aws:ses/domainIdentityVerification:DomainIdentityVerification",
--- a/platform/src/components/duration.ts
+++ b/platform/src/components/duration.ts
@@ -24,8 +24,10 @@ export type DurationHours = `${number} ${
   | "hour"
   | "hours"}`;
 
+export type DurationDays = `${number} ${"day" | "days"}`;
+
 export function toSeconds(
-  duration: Duration | DurationMinutes | DurationSeconds,
+  duration: Duration | DurationMinutes | DurationSeconds | DurationDays,
 ) {
   const [count, unit] = duration.split(" ");
   const countNum = parseInt(count);


dragonautdev:
This is code I've had to repeat over and over through multiple projects, so I'm really pumped this can be part of SST now.

vimtor:
note: if we merge this we should probably close #4941 and #6206

mkilp:
Agreed, at first i wanted a proper doc example but then decided to just add first party support. Hope we can get it merged soon once the SST team figures out the maintainer stuff.

vimtor:
@mkilp checking the API design you proposed and i'm wondering the following changes:

```ts
new sst.aws.Bucket("Example", {
  lifecycle: [
    prefix: "/tmp",
    expiresIn: "7 days",
    expiresAt: "2025-12-12",
  ]
});
```

the changes:

1. array instead of an object for defining multiple rules
2. rename `filterPrefix` to `prefix`
3. enabled is `true` by default
4. replace `expires` by top level properties:
  1. `expiresIn` uses common SST syntax for defining duration (could also be passed as "12 months" using the `duration.ts` utility)
  2. `expiresAt` as top level property
  3. `expiredObjectDeleteMarker` removed since people knowing what that is are probably going to need the Pulumi construct anyway
  

what do you think?

mkilp:
> @mkilp checking the API design you proposed and i'm wondering the following changes:
> 
> ```ts
> new sst.aws.Bucket("Example", {
>   lifecycle: [
>     prefix: "/tmp",
>     expiresIn: "7 days",
>     expiresAt: "2025-12-12",
>   ]
> });
> ```
> 
> the changes:
> 
> 1. array instead of an object for defining multiple rules
> 2. rename `filterPrefix` to `prefix`
> 3. enabled is `true` by default
> 4. replace `expires` by top level properties:
> 5. `expiresIn` uses common SST syntax for defining duration (could also be passed as "12 months" using the `duration.ts` utility)
> 6. `expiresAt` as top level property
> 7. `expiredObjectDeleteMarker` removed since people knowing what that is are probably going to need the Pulumi construct anyway
> 
> what do you think?

@vimtor appreciate the feedback, I asked the exact same thing in the discord if there is a standard api. I really like your proposed changes, I'll get that in.

mkilp:
@vimtor just changed the api and made a new union type for the expiration. Hopefully we can get that in now that PRs are being merged!

platform/src/components/aws/bucket.ts
```diff
@@ -99,6 +99,63 @@ interface BucketCorsArgs {
...
+   * ```
+   */
+  expiresAt: Input<string>;
+  expiresIn?: never;
+}
+
+type BucketLifecycleArgs =
+  | BucketLifecycleExpiresInArgs
+  | BucketLifecycleExpiresAtArgs;
+
```
vimtor:
```suggestion
interface BucketLifecycleArgs {
  /**
   * An S3 object key prefix that the lifecycle rule applies to.
   * @example
   * Applies to all the objects in the `images/` folder.
   * ```js
   * {
   *   prefix: "images/"
   * }
   * ```
   */
  prefix?: Input<string>;
  /**
   * Whether the lifecycle rule is enabled.
   * @example
   * ```js
   * {
   *  enabled: true
   * }
   * ```
   * @default `true`
   */
  enabled?: Input<boolean>;
  /**
   * Time after which the objects in the bucket should expire.
   * @example
   * ```js
   * {
   *  expiresIn: "30 days"
   * }
   * ```
   */
  expiresIn?: Input<`${number} days` | `${number} day`>;
  /**
   * Date after which the objects in the bucket should expire.
   * @example
   * ```js
   * {
   *  expiresAt: "2023-08-22"
   * }
   * ```
   */
  expiresAt?: Input<string>;
}

```

i had trouble generating the docs with the type union and the never  

less correct but simpler  

also, switched Duration to `${number} days` because the API only takes days at minimum. even better if you add `DurationDays` to `duration.ts` alongside the other specific durations

platform/src/components/aws/bucket.ts
```diff
@@ -384,6 +441,20 @@ export interface BucketArgs {
...
+   * This example will delete every object in the bucket after 30 days.
+   * ```js
+   * {
+   *  lifecycle: [{
+   *   enabled: true,
+   *   expiresIn: "30 days"
+   *  }]
+   * }
+   * ```
+   */
```
vimtor:
```suggestion
   * Delete objects in the "/tmp" directory after 30 days.
   * ```js
   * {
   *   lifecycle: [
   *     {
   *       prefix: "/tmp",
   *       expiresIn: "30 days"
   *     }
   *   ]
   * }
```

formatted the array so it's clearer it allows multiple rules and added a prefix to show a common use case

platform/src/components/aws/bucket.ts
```diff
@@ -854,6 +930,33 @@ export class Bucket extends Component implements Link.Linkable {
...
+            {
+              bucket: bucket.bucket,
+              rules: lifecycleRules.map((rule, index) => ({
+                id: `${name}LifecycleRule${index}`,
+                status: rule.enabled !== false ? "Enabled" : "Disabled",
+                expiration: rule.expiresIn
+                  ? { days: toSeconds(rule.expiresIn) / 86400 }
+                  : rule.expiresAt
+                    ? { date: rule.expiresAt }
+                    : undefined,
```
vimtor: if we allow both `expiresIn` and `expiresAt` we will need to change this

vimtor (changes requested):
one last thing, i'm maybe contradicting myself but i've noticed that the bucket already uses `filterPrefix` and `filterSuffix` for the notifications:

Image

probably we should use those names and their description to stay consistent

platform/src/components/aws/bucket.ts
mkilp: Added Duration days and will remove union type (sad)

platform/src/components/aws/bucket.ts
mkilp: Looking at https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Interface/LifecycleExpiration/ and https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketlifecycleconfigurationv2/#bucketlifecycleconfigurationv2ruleexpiration it seems that both values are allowed at the same time at the users discretion.

mkilp:
> one last thing, i'm maybe contradicting myself but i've noticed that the bucket already uses `filterPrefix` and `filterSuffix` for the notifications:
> 
> Image
> probably we should use those names and their description to stay consistent

Looking at pulumi AWS does not allow suffixes anymore, so I would be fine with staying with just "prefix"

platform/src/components/aws/bucket.ts
```diff
@@ -385,16 +430,29 @@ export interface BucketArgs {
...
+   * {
+   *   lifecycle: [
+   *     {
+   *       prefix: "/tmp",
+   *       expiresIn: "30 days"
+   *     }
+   *   ]
    * }
    * ```
    */
```
vimtor, line 446: something weird happened were this was placed below the property instead of above

```suggestion
   * The lifecycle configuration for the bucket.
   * @example
   * Delete objects in the "/tmp" directory after 30 days.
   * ```js
   * {
   *   lifecycle: [
   *     {
   *       prefix: "/tmp",
   *       expiresIn: "30 days"
   *     }
   *   ]
  * }
  * ```
  */
  lifecycle?: Input<Input<Prettify<BucketLifecycleArgs>>[]>;
```

platform/src/components/aws/bucket.ts
```diff
@@ -854,6 +917,33 @@ export class Bucket extends Component implements Link.Linkable {
...
+            {
+              bucket: bucket.bucket,
+              rules: lifecycleRules.map((rule, index) => ({
+                id: `${name}LifecycleRule${index}`,
+                status: rule.enabled !== false ? "Enabled" : "Disabled",
+                expiration: rule.expiresIn
+                  ? { days: toSeconds(rule.expiresIn) / 86400 }
+                  : rule.expiresAt
+                    ? { date: rule.expiresAt }
+                    : undefined,
```
vimtor: if it's possible to provide both properties `days` and `date` we should do it, otherwise mention it in the docs

vimtor (changes requested):
last requested changes @mkilp

platform/src/components/aws/bucket.ts
mkilp: fixed

platform/src/components/aws/bucket.ts
mkilp: Totally missed that, my bad! Should be fixed.

mkilp:
@vimtor Comments are all set! Sorry, my VSCode PR review panel really messed with me today.

platform/src/components/aws/bucket.ts
```diff
@@ -385,19 +430,22 @@ export interface BucketArgs {
    */
   cors?: Input<false | Prettify<BucketCorsArgs>>;
   /**
-   * Enable versioning for the bucket.
```
vimtor, line 388: @mkilp now the versioning comment got deleted 🥲

vimtor (changes requested):

platform/src/components/aws/bucket.ts
mkilp: Will get right on that, I'm so sorry I'm usually not this stupid. Note to self: Don't fiddle around with the VSCode PR view and just read the changes manually. :D

vimtor: don't worry the github ui is not the best one either haha

mkilp: @vimtor  all set now - also squashed all our back and forth into a single commit.

vimtor (changes requested):
some changes:

1. you will also need to add this line to 

```ts
"aws:s3/bucketLifecycleConfigurationV2:BucketLifecycleConfigurationV2",
```

to the app prefixed names:

https://github.com/sst/sst/blob/19b85a752620151232715a720331aeabd0c1ab4b/platform/src/components/component.ts#L168

otherwise it fails

2. the `expiresAt` need to be suffixed by `"T00:00:00Z"` otherwise it gives an error

mkilp:
> some changes:
> 
> 1. you will also need to add this line to
> 
> ```ts
> "aws:s3/bucketLifecycleConfigurationV2:BucketLifecycleConfigurationV2",
> ```
> 
> to the app prefixed names:
> 
> https://github.com/sst/sst/blob/19b85a752620151232715a720331aeabd0c1ab4b/platform/src/components/component.ts#L168
> 
> otherwise it fails
> 
> 2. the `expiresAt` need to be suffixed by `"T00:00:00Z"` otherwise it gives an error

I added the app prefixed name but I am not sure what you are seeing with the expiresAt error. According to https://www.pulumi.com/registry/packages/aws/api-docs/s3/bucketlifecycleconfigurationv2/#bucketlifecycleconfigurationv2ruleexpiration just using a date string should in theory work. (ref: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6)

But you are absolutely correct it currently doesnt. Even the AWS SDK states it takes a ISO 8601 format date and defaults the time to midnight (https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/command/PutBucketLifecycleConfigurationCommand/)

I will append the timestamp to midnight. Sorry for my lack of test and thanks for calling this out @vimtor! I will amend my commit and force push as well as clarifying the example that it will default to midnight. I will also look into raising an upstream PR into pulumi-aws for this if its not intended.

platform/src/components/aws/bucket.ts
```diff
@@ -854,6 +920,37 @@ export class Bucket extends Component implements Link.Linkable {
...
+              rules: lifecycleRules.map((rule, index) => ({
+                id: `${name}LifecycleRule${index}`,
+                status: rule.enabled !== false ? "Enabled" : "Disabled",
+                expiration:
+                  rule.expiresIn || rule.expiresAt
+                    ? {
+                        days: rule.expiresIn
+                          ? toSeconds(rule.expiresIn) / 86400
+                          : undefined,
+                        date: `${rule.expiresAt}T00:00:00Z`,
```
vimtor:
```suggestion
                        date: rule.expiresAt ? `${rule.expiresAt}T00:00:00Z` : undefined,
```

vimtor (changes requested):

platform/src/components/aws/bucket.ts
mkilp: done. sorry we had to go back and forth so much on this.

vimtor (approved):
working!

```ts
const bucket = new sst.aws.Bucket("MyBucket", {
  lifecycle: [
    {
      expiresIn: "30 days",
    },
    {
      prefix: "tmp1/",
      expiresIn: "30 days",
    },
    {
      prefix: "tmp2/",
      expiresAt: "2028-12-31",
    },
  ],
});
```

Image

PR #6232, opened 2025-11-19
fix(ssr-site): update autostart to respect autostart and dev: false

kilakewe:
Fixed the autostart behavior in SSR sites to properly respect the dev: false configuration. Previously, the autostart defaulted to true regardless of the dev setting, causing development servers to start even when explicitly disabled.

Changes
- Updated platform/src/components/aws/ssr-site.ts
- Modified the autostart logic to check if args.dev === false before defaulting to true
- New logic: devArgs.autostart ?? (args.dev === false ? false : true)

Problem

When users set dev: false in their SSR site configuration, the development server would still autostart because the autostart property only checked for devArgs.autostart and defaulted to true in all other cases.

Solution

The autostart now respects the dev configuration:
- If devArgs.autostart is explicitly set, use that value
- Otherwise, if dev: false, set autostart to false
- Otherwise, default to true

### DIFF
--- a/platform/src/components/aws/ssr-site.ts
+++ b/platform/src/components/aws/ssr-site.ts
@@ -927,7 +927,7 @@ async function handler(event) {
         outputs: {
           title: devArgs.title,
           command: output(devArgs.command ?? "npm run dev"),
-          autostart: output(devArgs.autostart ?? true),
+          autostart: output(devArgs.autostart ?? args.dev === false ? false : true),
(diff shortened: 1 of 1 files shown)


vimtor:
i believe this is the intended behaviour

if `dev: false` you are disabling the dev mode completely, so the website will build

if you need to turn off the autostart just pass `dev.autostart: false`

PR #6236, opened 2025-11-24
feat: add gcp home provider

shreyassanthu77:
- used aws and cloudflare as a reference to implement this
- uses gcp project id as the bucket name cuz it's already globally unique and puts it in gcs (project region)
- uses  google secrets manager to save secrets

## few considerations
- secrets manager is not free and needs to be enabled manually (super cheap but still) should i do it more like cloudflare and put it in the bucket?
- the gcp error message extraction was kind of annoying, feels really hacky but big pickle said it's the only good way :)
- also added couple  fixme comments for stuff i wasn't sure about what to do
- looks like it's everything is working fine

I will try to port a few my clients' pulumi automation setups to sst and try moving the common patterns to cloud build, cloud run, astro components and stuff if it works out well


shreyassanthu77:
also this permalink thing doesn't work but not sure if that is me being a dummy or what just putting it here before i forget
image

azzzy:
@shreyassanthu77 , Google's Always Free Tier will cover most of the use cases if rotation notifications are part of the deployment workflow.


Resource | Monthly free usage limits1
-- | --
Active secret versions | 6 versions
Access operations | 10,000 operations
Rotation Notifications | 3 rotations

shreyassanthu77:
> @shreyassanthu77 , Google's Always Free Tier will cover most of the use cases if rotation notifications are part of the deployment workflow.
> 
> 
> Resource | Monthly free usage limits1
> -- | --
> Active secret versions | 6 versions
> Access operations | 10,000 operations
> Rotation Notifications | 3 rotations
> 
> 

Yeah but 6 versions is not that many

shreyassanthu77:
Not sure though it's just a couple cents max for most people if any so should be fine ig?

PR #6246, opened 2025-11-29
[Docs] Update TanStack Start docs to reflect current required steps

dragonautdev:
This PR updates the docs for TanStack Start, so that they explain the current steps needed for TS Start to work and deploy correctly to AWS with SST. 

Biggest change is that now there's no need for `app.config.ts`, and settings are done straight from the vite.config.ts file


vimtor (commented):
thanks for your contribution @dragonautdev 

could you also check the tanstack examples in the examples folder? they probably need to be updated too

dragonautdev:
@vimtor I've updated both the `aws-tanstack-start` and `aws-tanstack-start-vite` examples. Thanks for calling those out. 

I *think* we can also delete the `aws-tanstack-start-alpha` example, as it's highly deprecated and can lead people to confusion. Do you agree? If so, I'll just remove it.

vimtor:
thanks @dragonautdev 

yep, let's remove that deprecated example too

dragonautdev:
@vimtor it's done. I believe this one covers all bases for TS Start now.

vimtor (changes requested):
@dragonautdev last change:

when trying to run the `-vite` example i got this error:

```
$ vite build
error during build:
TypeError: Cannot read properties of undefined (reading 'meta')
```

looking at the docs we should delete this `-vite` example since they are fairly similar and both use vite

dragonautdev:
@vimtor done.

vimtor (approved):
working!

deployed example:

Image

docs:

Image

sommeeeer:
Anyone succesfully deployed to SST with newest version of Tanstack Start BTW? (`1.142.5`)

jamesjoplin:
> Anyone successfully deployed to SST with newest version of Tanstack Start BTW? (`1.142.5`)

I followed these steps to render a deployment

- create a new tanstack start app via the CLI above
- `npm create @tanstack/start@latest`
- follow the steps per the picture above
- update the vite.config.ts file to include the awsLambda streaming settings like below
```typescript
const config = defineConfig({
  plugins: [
    devtools(),
    nitro(),
    // this is the plugin that enables path aliases
    viteTsConfigPaths({
      projects: ['./tsconfig.json'],
    }),
    tailwindcss(),
    tanstackStart(),
    viteReact(),
  ],
  nitro: { // this is the important detail
    preset: 'aws-lambda',
    awsLambda: {
      streaming: true,
    },
  },
})
```
- init an sst powered application
- `npx sst@latest init`
- update the sst.config.ts with a tanstack component
- ` new sst.aws.TanStackStart('MyWeb'),`

Note: If you wish to utilize the TODO example for server api based TODO's - change the path to be `/tmp/todos.json` -- otherwise it will not work since the default lambda path is read-only.

HTH

Issue #6033, opened 2025-08-09
Upgrade plan for @pulumi/aws to v7


abnud11:
Version 7 of @pulumi/aws got released with some breaking changes. According to https://www.pulumi.com/registry/packages/aws/how-to-guides/7-0-migration/, for the SST codebase, we need these steps:

1. Run `pulumi up --refresh --run-program` to update the state to be compatible with @pulumi/aws v7. We need to determine whether this process is automated when people upgrade SST and deploy/refresh, or if it's documented as part of the migration to v4 (which, in my opinion, should be the version of SST using @pulumi/aws v7, given the breaking change).
2. `s3.Bucket` and `s3.BucketV2` are now the same, including their subresources, e.g. aws.s3.BucketAccelerateConfigurationV2. That means that SST should just use `s3.Bucket`.
3. Many properties in many resources, including the Provider itself, were removed. We need to scan the codebase for these properties and replace them.

The primary concern here is how to upgrade without breaking people's infrastructure, state must be updated before any attempt to deploy when upgrading to v7, also since state isn't compatible between v6 and v7, we need to require v7 of @pulumi/aws so any person who is still using v6 will get an error message before any attempt at deployment.

My proposed solution involves checking the state's compatibility with v7 and failing fast if it's not. Then, it logs a message to run `sst refresh` first, which internally executes `pulumi up --refresh --run-program`. Also, the new version of SST using @pulumi/aws v7 must be v4, indicating that it's a breaking release since many people use `new aws.Resource` directly in their `sst.config.ts` file.

I'm open to suggestions and willing to help with TypeScript code if it's needed.

abnud11:
@fwang Aren't there any plans to migrate to @pulumi/aws v7?

abnud11:
@fwang I can help with anything Typescript related if that's required, so maybe we should push on this?

jamesgibbons92:
Working on this in https://github.com/sst/sst/pull/6259
I agree with your suggestions @abnud11 and have added some logic to check the provider versions and inform the user.
We'll probably do a sst minor version ie 3.17.X -> 3.18.0 with some info in the release notes as well.
I'll tidy it up once pulumi 3.210 is merged as this PR depends on it.

vimtor:
fixed in [v4.0.0](https://github.com/anomalyco/sst/releases/tag/v4.0.0)

PR #6254, opened 2025-12-03
Update aws provider to 7.12.0

jamesgibbons92:
https://www.pulumi.com/registry/packages/aws/how-to-guides/7-0-migration/

- [x]  tsc passes
- [ ]  verify upgrade path for each sst aws component doesn't introduce breaking changes, e.g. unexpected replacements


jamesgibbons92:
@vimtor if you can help test this that would be appreciated!
The main thing is to check whether the breaking changes in the provider impact any of the sst components.
Also we need to check whether upgrading from the current default provider causes any unexpected changes such as replacing resources.

One thing to note is the provider is changing the BucketV2 and Bucket components, currently the BucketV2 is deprecated in this new version and it will be removed in the next major version.

abnud11:
@jamesgibbons92 I actually did some work regarding this, you can see it here:
https://github.com/abnud11/sst/commit/98d59f8c10d517063cf309c2ba0fadd644bdc8b8

There are a lot of required changes than just what's made in this PR in my opinion, examples include:
1- BucketV2 -> Bucket
2- getRegionOutput().name => getRegionOutput().region
3- authTokenUpdateStrategy for ElastiCache ReplicationGroup changed its default, so we either need to set it explicitly to `ROTATE` or just use the default value(different from `ROTATE` in v7) but risk breaking apps
4- multiple places use `tagsAll` instead of `tags`

abnud11:
See this:
https://www.pulumi.com/registry/packages/aws/how-to-guides/7-0-migration/

abnud11:
Oh and we must tell the users to pass `--refresh --run-program` to `pulumi up` command, that is `sst` deploy command would take a new flag, pass that flag to `pulumi up`, otherwise pulumi state will break after the upgrade:

> After you have upgraded the AWS provider library to v7 you need to run Pulumi with the --refresh and --run-program CLI flags in order to migrate the state to be compatible with the new major version.

I'm not sure how to do this in a non-breaking way, we need to find a way to check for state compatibility with v7 and pass `--refresh --run-program` to pulumi when that happens

abnud11:
https://github.com/sst/sst/issues/6033

jamesgibbons92:
Oh great well done, we can use your fork in that case and take the discussion to the at issue - I had no seen it. Thanks!

abnud11:
> Oh great well done, we can use your fork in that case and take the discussion to the at issue - I had no seen it. Thanks!

Thanks but the branch has some changes not related to aws 7.12 upgrade but I will create a specific branch and provide a PR soon, but will be sometime before that happens, I'm currently working on an assignment for college

PR #6257, opened 2025-12-05
feat(aws): enable overriding cpu / memory / storage when running ECS tasks

achadha235:
Often, you need to override an existing ECS task definition's CPU / memory settings based your workload at runtime.

This change allows `sst.task.run` to pass these values through to the AWS SDK. 
Hopefully, this can help others save some time and optimize their ECS costs. 

PS: this is my first contribution to sst so please lmk if I missed something I would be happy to learn!

### DIFF
--- a/sdk/js/src/aws/task.ts
+++ b/sdk/js/src/aws/task.ts
@@ -78,15 +78,15 @@ export module task {
      */
     capacity?: "fargate" | "spot";
     /**
-     * The amount of memory allocated for this task. This value overrides the memory allocated in the task definition.
+     * Overrides the memory allocated for this task in the task definition.
      */
-    memory?: string;
+    memory?: `${number} GB`;
     /**
-     * The amount of CPU allocated for this task. This value overrides the CPU allocated in the task definition.
+     * Overrides the CPU allocated for this task in the task definition.
      */
-    cpu?: string;
+    cpu?: `${number} vCPU`;
     /**
-     * Ephemeral storage size in GiB allocated for this task. This value overrides the ephemeral storage allocated in the task definition.
+     * Overrides the ephemeral storage (in GiB) allocated for this task in the task definition.
      */
     storage?: number;
   }
--- a/sdk/js/src/aws/task.ts
+++ b/sdk/js/src/aws/task.ts
@@ -1,5 +1,4 @@
 import { AwsOptions, client } from "./client.js";
-
 /**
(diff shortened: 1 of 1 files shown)


sdk/js/src/aws/task.ts
```diff
@@ -247,6 +258,11 @@ export module task {
           },
         },
         overrides: {
+          ...(options?.cpu ? { cpu: options.cpu } : {}),
+          ...(options?.memory ? { memory: options.memory } : {}),
+          ...(options?.storage
+            ? { ephemeralStorage: { sizeInGiB: options.storage } }
+            : {}),
```
vimtor:
```suggestion
          ...(options?.cpu ? { cpu: (parseFloat(options.cpu.replace(" vCPU", "")) * 1024).toString() } : {}),
          ...(options?.memory ? { memory: (parseFloat(options.memory.replace(" GB", "")) * 1024).toString() } : {}),
          ...(options?.storage ? { ephemeralStorage: { sizeInGiB: parseInt(options.storage.replace(" GB", "")) } } : {}),

```

sdk/js/src/aws/task.ts
```diff
@@ -78,6 +77,18 @@ export module task {
...
+     */
+    memory?: `${number} GB`;
+    /**
+     * Overrides the CPU allocated for this task in the task definition.
+     */
+    cpu?: `${number} vCPU`;
+    /**
+     * Overrides the ephemeral storage (in GiB) allocated for this task in the task definition.
+     */
+    storage?: number;
```
vimtor:
```suggestion
    /**
     * Overrides the ephemeral storage allocated for this task in the task definition.
     */
    storage?: `${number} GB`;
```

vimtor (changes requested):
thanks for your contribution @achadha235 

sadly your PR gives the following error when launching a task:

```
{
  __type: "SerializationException",
  Message: "STRING_VALUE cannot be converted to Integer",
}
```

this is because the AWS API expects a different shape for those parameters

i've managed to make it work by doing the suggested changes

here's the proof:

```ts
await task.run(Resource.MyTask, {}, {
  cpu: "2 vCPU",
  memory: "4 GB",
  storage: "30 GB",
});
```

CleanShot 2025-12-06 at 17 53 09@2x

achadha235:
thanks for catching that @vimtor, I've added this fix.

vimtor (approved):

PR #6259, opened 2025-12-06
AWS v7 Provider upgrade

jamesgibbons92:
AWS provider upgade has some breaking changes which requires some manual intervention before the user should deploy using the new version. 
The steps are explained here https://www.pulumi.com/registry/packages/aws/how-to-guides/7-0-migration

- ~~Upgrade pulumi to latest version (this contains the new arg flag --run-program for refresh). (Thanks to @abnud11)~~ This will be merged separately in another PR.
- Upgrade AWS from v6 -> v7 (Thanks to @abdnud11)
- Add provider major upgrade check when running `sst deploy`. Which asks the user to explicitly run `sst refresh` to migrate their state to the new version. Deploy/diff will be blocked until this is done and no resources in your state are using an outdated version. 
- Add `--run-program` pulumi arg to refresh command, which is required for the migration.


jamesgibbons92:
Draft message for release notes: 

> This release updates the aws provider from v6 to v7. This major upgrade contains a number of breaking changes and deprecations. Follow the guide here https://www.pulumi.com/registry/packages/aws/how-to-guides/7-0-migration/ to view the changes and update any affected pulumi aws components you may have.
> There should be no breaking changes to your SST components (except for possibly any inputs you were transforming), however as part of the upgrade you must migrate your state using `sst refresh` before deploying with the new provider. It's advisable to also use `sst diff` after this command to ensure there are no unexpected changes.

jamesgibbons92:
One thing to consider, adding support for `--refresh` arg to the deploy command, to automatically run refresh before deploy. This would provide a way of getting around having to do a manual refresh step before deploy and may be useful for people that have lots of stages to update.

abnud11:
I hope we can push on this PR as a lot of features are blocked by it :+1: 
@jamesgibbons92 @vimtor

jamesgibbons92:
@vimtor Now that pulumi has been merged I've updated this PR so it's good to review now. 
I'm doing some more testing, but would appreciate a second pair of eyes on this as well!

austeane:
Would love for this to be merged, it's key for my project, and I've switched to using this branch as my base.

cmd/sst/deploy.go
```diff
@@ -79,6 +79,12 @@ var CmdDeploy = &cli.Command{
 			"sst deploy --dev",
 			"```",
 			"The `--dev` flag will deploy your resources as if you were running `sst dev`.",
+			"",
+			"```bash frame=\"none\"",
+			"sst deploy --refresh",
+			"```",
+			"Pass --refresh flag to refresh your state resources before deploying any changes.",
+			"This is useful for making sure your state reflects your cloud resources accurately before applying updates.",
```
vimtor: thinking about what's the difference between `sst deploy --refresh` and `sst refresh` i discovered that the `refresh` command is broken in this pr:

```
default_7_12_0 pulumi:providers:aws
pulumi:providers:aws resource 'default_7_12_0' has a problem: Missing region information
Make sure you have set your AWS region, e.g. `pulumi config set aws:region us-west-2`.
```

regardless of whether we need this new flag, the user should be able to do:
1. `sst refresh`
2. `sst deploy`

to upgrade their state right? @jamesgibbons92 

if that's the case we might want to keep it simple and don't introduce new flags

pkg/project/preflight.go
```diff
@@ -0,0 +1,95 @@
...
+	toVersion    string
+	message      string
+}
+
+var migrationNotices = []migrationNotice{
+	{
+		providerName: "aws",
+		fromVersion:  "6.0.0",
+		toVersion:    "7.0.0",
+		message:      "Detected AWS provider upgrade to v7.\n\nUpgrading from v6 to v7 introduces some breaking changes and requires state migration before deploying. Refer to the SST release notes and the pulumi migration guide to make the relevent changes before deploying: https://www.pulumi.com/registry/packages/aws/how-to-guides/7-0-migration \n\nThis notice will clear after migrating your state",
```
vimtor:
```suggestion
		message:      "Detected AWS provider upgrade to v7.\n\nUpgrading from v6 to v7 introduces some breaking changes and requires state migration before deploying. Refer to the SST release notes and the Pulumi migration guide to make the relevent changes before deploying: https://www.pulumi.com/registry/packages/aws/how-to-guides/7-0-migration \n\nThis notice will clear after migrating your state",
```

from this message it didn't become clear that:
1. the user needs to `refresh` the state
2. the migration is one-way only? (i tried going back and it didn't work)

www/src/content/docs/docs/examples.mdx
```diff
@@ -764,8 +764,8 @@ View the [full example](https://github.com/sst/sst/tree/dev/examples/aws-bun-red
 ---
 ## AWS Cluster custom autoscaling
 
-In this example, we'll create a cluster that can autoscales based on a custom
-metric; in this case, the number of messages in a queue.
+In this example, we'll create a cluster that autoscales based on a custom
```
vimtor: i believe the changes made to the `examples.mdx` file will be lost because it gets regenerated from every JSDocs found above the `sst.config.ts` inside the `examples/` folder

vimtor (changes requested):
hey @jamesgibbons92

i left a couple of minor comments about the migration steps

everything else seems to be working. great job!

will be testing more examples today to see if they still work

cmd/sst/deploy.go
jamesgibbons92: Yeh it's not necessary, I can remove it.
But you said the refresh command is broken? Or was it just when using as a deploy arg?

pkg/project/preflight.go
jamesgibbons92: Yes I don't think you can rollback easily by downgrading the sst package, this would have to be handled on the provider level anyway. 
Users would have to rollback the state manually, probably by restoring a previous version in the state s3 bucket (we should really have a cli command for this)

www/src/content/docs/docs/examples.mdx
jamesgibbons92: Yeh this change is literally from the autogeneration, I ran it as I had originally made some docs changes - I didn't edit this file manually. I guess there were some changes that had not yet had the markdown regenerated for.

cmd/sst/deploy.go
vimtor: when i tried doing:

1. sst refresh
2. sst deploy

to migrate the state i got an error

did it work for your?

pkg/project/preflight.go
vimtor: makes sense

maybe we should be more explicit in the error about which commands have to be run and that rolling back is "impossible"?

vimtor: alternatively we could run the commands for the user and show a nice confirmation UI like this:
https://github.com/anomalyco/sst/pull/5109

but i believe just being more explicit should be enough

cmd/sst/deploy.go
austeane: Refresh and deploy did work for me off this branch

vimtor: got this error when trying the refresh then deploy:

CleanShot 2026-01-22 at 19 03 58@2x

i'm also thinking that if someone runs `sst refresh` and they’re on a new version they might accidentally upgrade their state. seems unlikely though

maybe we should be more explicit - either with a major version that auto-refreshes the state on the first deploy or a specific `sst upgrade` command

i'm not sure tbh, what do you think @austeane @jamesgibbons92?

jamesgibbons92: @vimtor can you DM me your sst.config.ts so I can reproduce the region error - I've not seen it in my testing.

> i'm also thinking that if someone runs sst refresh and they’re on a new version they might accidentally upgrade their state. seems unlikely though

I don't know how we could possibly handle this scenario. We would probably to a sst major version update anyway for the breaking changes, so it's up to the users to check the release notes and upgrade on a lower environment.

> maybe we should be more explicit - either with a major version that auto-refreshes the state on the first deploy or a specific sst upgrade command

Yeh I think major version, i'm not sold on auto refresh, as users need to check their non-sst aws components and make any relevent updates. And I feel like having refresh and deploy as separate steps gives people time to check the refresh or recover if something goes wrong with state migration - they can fix this by rolling back to state with no changes to their infra. Wheras with an auto-refresh before a deploy  it would just continue  on with deploy which no chance to verify the changes between.

pkg/project/preflight.go
jamesgibbons92: I'll update the copy to be more specific. I was expecting the more detailed steps to go in the release notes on github but it doesn't hurt to be specific here as well.

cmd/sst/deploy.go
jamesgibbons92: @vimtor Do you have `region`  set in your aws profile? I can reproduce this error if I remove my region config.
Looks like sst defaults to us-east-1 if you have no default region set. but this doesn't seem to play well with the `--run-program` pulumi arg.

jamesgibbons92: @vimtor this should be resolved now if you can test again.
The issue was --run-program arg makes it such that the refresh command uses the latest provider you have in your program, rather than the one that is stored in state. 
SST does not pass the provider config to pulumi cli via the --config arg when calling the refresh command (and it's not supported in the version of pulumi we're using). So to resolve i've set the AWS_REGION in the environment variables similarly to how sst sets AWS_PROFILE.

samuelzapote:
Really appreciate all the hard work. Just spent the whole day figuring out the "mx-central-1" region isn't supported yet because it needs this provider upgrade. Our non-profit team is really hoping to hit the ground running with sst on our nextjs project. We're rooting for y'all! 🥹🚀

We're currently getting `Type '"mx-central-1"' is not assignable to type 'Input<Region> | undefined'. Did you mean '"me-central-1"'?` when trying to initialize sst and setting the following config,

```
providers: {
  aws: {
    region: "mx-central-1"
  }
}
```

We've found out that the reason is because support for mx-central-1 was added in [pulumi-aws v7.16.0](https://github.com/pulumi/pulumi-aws/releases/tag/v7.16.0). This PR will get us nearly there, hope we can have a quick follow up PR to bump it to 7.16.0 once this one merges. Leaving all the info I've found here in case anyone else runs into this problem, and in case this takes a while longer. (Animo banda, ya mero llega sst a Mexico! 🇲🇽🧑🏽‍💻)

Edit (Feb 14, 2026 💌): We're going to go ahead and work with an npm symlink to update our pulumi-aws to v7.16.0. Hopefully to help prepare any findings we run into for future contributions. See y'all soon! Happy Valentines Day lovely contributors. <3

vimtor:
@samuelzapote viva méxico! a ver si mergeamos esto pronto

kevinnorris:
Really looking forward to this upgrade. Any updates on how close it is to merging?

vimtor:
@kevinnorris we’re looking to merge it this week or beginnings of the next

platform/src/components/component.ts
```diff
@@ -159,7 +159,7 @@ export class Component extends ComponentResource {
               "aws:rds/proxyDefaultTargetGroup:ProxyDefaultTargetGroup",
               "aws:rds/proxyTarget:ProxyTarget",
               "aws:route53/record:Record",
-              "aws:s3/bucketCorsConfigurationV2:BucketCorsConfigurationV2",
+              "aws:s3/bucketCorsConfiguration:BucketCorsConfiguration",
```
vimtor, line 163: any reason we swapped this one but kept the other V2 like `BucketWebsiteConfigurationV2` or `BucketVersioningV2`?

www/src/content/docs/docs/examples.mdx
```diff
@@ -764,8 +764,8 @@ View the [full example](https://github.com/sst/sst/tree/dev/examples/aws-bun-red
 ---
 ## AWS Cluster custom autoscaling
```
vimtor, line 765: some of the examples or jsdocs still reference `getRegionOutput().name` instead of `getRegionOutput().region`

platform/src/components/aws/bucket.ts
```diff
@@ -792,7 +792,7 @@ export interface BucketSubscriberArgs {
 
 interface BucketRef {
   ref: boolean;
-  bucket: s3.BucketV2;
+  bucket: s3.Bucket;
```
vimtor, line 795: we need to migrate the `BucketLifecycleConfigurationV2` too

i probably merged this after your update, so that's why you didn't include it

i'll pause new components until we merge this

vimtor (changes requested):

platform/src/components/component.ts
jamesgibbons92: nope just an oversight, I kept the other v2 in just because technically the components are only deprecated so people could still have a valid reason for using them. there isn't a reason I can think not the support the old v2 until they delete it (presumably in v8)

vimtor (approved):
soon after merging this i'll release a major v4 version

thanks everyone, especially the goat himself @jamesgibbons92

petterw03:
The release doens't contain any notes about breaking changes?

vimtor:
@petterw03 https://sst.dev/docs/migrate-from-v3/

\*technically\* there are no breaking changes in sst, only if you use transforms or the aws provider directly

were you refering to the release in github?

petterw03:
> @petterw03 https://sst.dev/docs/migrate-from-v3/
> 
> *technically* there are no breaking changes in sst, only if you use transforms or the aws provider directly
> 
> were you refering to the release in github?

Yes, I saw that there was a new major release but was a bit mystified as to what made it a major release. I think it would be good to link to the migration manual from the release notes.

vimtor:
@petterw03 done!

PR #6262, opened 2025-12-06
feat: biome suppression for auto-generated ts files

nemeth2300:
Add support for biome error suppressions


vimtor (approved):
approved!

Image

PR #6263, opened 2025-12-06
Fix docs when running locally

vimtor:
when checking #5819 i found the docs can't be tested locally because the JSDoc extraction broke for two files:

- `aws/permission.ts`: needs a global `@packageDocumentation` since we create a page via `useModuleComment`
- `cloudflare/kv.ts`: needs a getter for `namespaceId` since  `getSSTLink` props are iterated

i've also added instructions on how to run the docs locally in the README.md

proof of them working now:

https://github.com/user-attachments/assets/d792f67b-a41e-4551-9fcf-91709a240628

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ go run ../../cmd/sst <command>
 If you want to build the CLI, you can run `go build ./cmd/sst` from the root. This will create a
 `sst` binary that you can use.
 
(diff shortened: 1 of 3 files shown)


Jcgrguez-wq (commented):
Gracias

Issue #5220, opened 2024-12-17
Feat: Add Cognito Domain


manhcuongincusar1:
Hi all,

I am currently migrating my application from SST v2 to v3. My app supports login with Google and Facebook via Cognito. I originally developed the app using the example from the SST guide:
[How to Add Google Login to Your Cognito User Pool](https://guide.sst.dev/examples/how-to-add-google-login-to-your-cognito-user-pool.html).

In v2, I used the following code to add a Cognito domain to the user pool:

```
// Auth Domain
const domain = auth.cdk.userPool.addDomain("AuthDomain", {
  cognitoDomain: {
    domainPrefix: `${app.stage}-auth-domain`,
  },
});
```

However, in SST v3, I couldn’t find any way to add the Cognito domain. Could you please provide guidance or examples of how to achieve this in v3?

Thanks in advance for your help!

manhcuongincusar1:
I found a solution! I used Pulumi AWS to create a Cognito Domain. I hope this helps anyone with the same concern.



```
const domain = new aws.cognito.UserPoolDomain("main", {
    userPoolId: userPool.id,
    domain: `${$app.stage}-auth`,
});
```

In sst.config.ts

```auth_domain: domain.domain.apply(domain => `https://${domain}.auth.${region}.amazoncognito.com`),```

**Comment by paupenin** [comment]
Thank you for your contribution @manhcuongincusar1 !

**Comment by AlessandroVol23** [comment]
I've built it the same way with Pulumis resource for `UserPoolDomain` 

But I think SST would benefit if this would be available like the client as well since this is used a lot IMO.

Similar to the client maybe? 

```ts
const userPoolClient = userPool.addClient('client')

// Maybe like that?
userPool.addDomain('domain')
```

jayair:
Yeah let me put it on the roadmap

jhw:
Any news on this one? I too would like an SST Cognito Domain

onceLearner:
The support for  Domain is really needed especially with the new Managed login

andersonbcdefg:
Would also love this!

PR #6265, opened 2025-12-07
Add custom domain to `CognitoUserPool`

wcvo:
Adds support for configuring Cognito User Pool domains via `domain` prop on `CognitoUserPool`. This addresses the feature request in #5220.

**Features:**
- Cognito prefix domains (e.g., `my-app.auth.us-east-1.amazoncognito.com`)
- Custom domains with automatic ACM certificate creation and DNS configuration
- Multiple DNS providers (Route 53, Cloudflare, Vercel)
- Option to use existing certificates

**Usage:**

   ```ts
   // Prefix domain
   new sst.aws.CognitoUserPool("MyUserPool", {
      domain: {
         prefix: "my-app-dev"
      }
   });

   // Custom domain with Route 53
   new sst.aws.CognitoUserPool("MyUserPool", {
      domain: "auth.example.com"
   });

   // Custom domain with Cloudflare
   new sst.aws.CognitoUserPool("MyUserPool", {
      domain: "auth.example.com"
      dns: sst.cloudflare.dns()
   });
   ```

**Test plan**

   - [x] Tested prefix domain creation
   - [x] Tested custom domain creation
   - [x] Tested domain change (deleteBeforeReplace)


jamesgibbons92:
Good contribution @wcvo ! 
Is there a reason you chose a method addDomain(), over just adding a `domain` prop to the user pool component? I think a prop makes more sense and fits the pattern used in other sst components which support custom domains

wcvo:
@jamesgibbons92 Valid point! I initially went with a method to match the existing `addClient()` and `addIdentityProvider()` patterns in the component, but you're right that a prop makes more sense since a User Pool can only have one domain (vs multiple clients/providers). Will update shortly.

dimitor115:
Nice work! I’ve been looking for this 😄 If I may suggest: since this is such a convenient setup, it’d be great to include the domain usage in the `aws-cognito` example as well. It should render correctly once my other PR is merged.

vimtor:
let's wait on merging #6271 first

vimtor (approved):
thanks for your contribution @wcvo 

<details>

<summary>did a couple of changes but it's working like a charm</summary>

Image

</details>

i've also created #6504 as a separate issue to avoid blocking this one

PR #6269, opened 2025-12-08
Add the ability to pass a role to the EC2 based NAT of the VPC component

mkilp:
Hi all,

I ran into this issue in my organization where role creation is blocked for most accounts. I felt this was probably a common issue where existing roles have to be reused instead of creations.

Hence this PR.

LMK if you need any changes!

P.S.: I know this can also be done with transforms but I figured we want a "SST" way of doing it :)

### DIFF
--- a/platform/src/components/aws/vpc.ts
+++ b/platform/src/components/aws/vpc.ts
@@ -795,12 +795,12 @@ export class Vpc extends Component implements Link.Linkable {
         if (nat) {
           if (nat.ec2 && nat.type === "managed")
             throw new VisibleError(
-              `"nat.type" cannot be "managed" when "nat.ec2" is specified`,
+              `The "nat.type" cannot be "managed" when "nat.ec2" is specified.`,
             );
 
-          if (!nat.type)
+          if (!nat.type && !nat.ec2)
             throw new VisibleError(
-              `Missing "nat.type" for the "${name}" VPC. It is required when "nat.ec2" is not specified`,
+              `Missing "nat.type" for the "${name}" VPC. It is required when "nat.ec2" is not specified.`,
             );
 
           if (nat.ip && nat.ip.length !== zones.length)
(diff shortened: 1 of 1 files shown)


platform/src/components/aws/vpc.ts
```diff
@@ -938,28 +958,35 @@ export class Vpc extends Component implements Link.Linkable {
...
-          {
-            assumeRolePolicy: iam.getPolicyDocumentOutput({
-              statements: [
-                {
-                  actions: ["sts:AssumeRole"],
-                  principals: [
+        const role = nat.ec2.role
+          ? iam.Role.get(
+              `${name}NatInstanceRole`,
+              nat.ec2.role,
```
vimtor, line 1017: looking at the rest of the codebase, `iam.Role.get` accepts the role name as the second parameter

there's a `parseRoleArn` utility for this

vimtor (changes requested):

platform/src/components/aws/vpc.ts
mkilp: I generally agree with you, however there seems to be inconsistency across SST:

Service (-> Fargate.ts) uses `RoleName` as the input and not an ARN.
Function.ts uses RoleArn hence the utility.
Cognito uses RoleArn as well.

While I understand changing function and cognito to Name or Service to ARN would be a breaking change, we do have to pick a lane at this point. Since the native getter function takes Names over ARNs I would just use names going forward.

@vimtor thoughts?

vimtor: interesting @mkilp

let's go with the name like you said and see what @fwang says in the review

fargate is a newer addition, so he probably changed the pattern for a reason

mkilp: I changed the example in my PR to point out we are expecting the role name instead of the ARN now.

vimtor (approved):

<details>

<summary>i had to fix a couple of minor bugs but it is working
</summary>

Image

</details>

PR #6271, opened 2025-12-09
Fix missing AWS Cognito example

dimitor115:
Hi all, I have noticed the AWS Cogito example is not rendered on the [https://sst.dev/docs/examples/](https://sst.dev/docs/examples/) because of a missing JS doc at the top. I have added it and now it will render properly. (I have run `/www generate.ts` and confirmed it's there now). 
image

- Should I include in my PR the generated examples.mdx or do you have another process of generating it? As I see there are more changes than mine in the diff.

### DIFF
--- a/examples/aws-cognito/sst.config.ts
+++ b/examples/aws-cognito/sst.config.ts
@@ -1,8 +1,8 @@
 /// <reference path="./.sst/platform/config.d.ts" />
 /**
- * ## AWS Cogito User Pool
+ * ## AWS Cognito User Pool
  *
- * Create a Cognito User Pool with triggers and identity pool.
+ * Create a Cognito user pool with triggers and identity pool.
  *
  */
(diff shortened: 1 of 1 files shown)


examples/aws-cognito/sst.config.ts
```diff
@@ -1,5 +1,10 @@
 /// <reference path="./.sst/platform/config.d.ts" />
-
+/**
+ * ## AWS Cogito User Pool
+ *
+ * Create a Cognito User Pool with triggers and identity pool.
```
vimtor:
```suggestion
 * ## AWS Cognito User Pool
 *
 * Create a Cognito user pool with triggers and identity pool.
```

vimtor (changes requested):

vimtor (approved):

PR #6272, opened 2025-12-09
Add callbackUrls to aws cognito userpool client args

dimitor115:
Hi all, while implementing an AWS User Pool in AWS Cognito I ran into a small inconvenience with the `addClient` method. AFAIK, the `callbackUrl` is required for the managed login flow to redirect back to your app, but it isn’t exposed as an input parameter, so you currently have to rely on a small `transform` workaround:

My current code:
```TypeScript
    const webClient = userPool.addClient("Web", {
      transform: {
        client: (args: UserPoolClientArgs) => {
          args.callbackUrls = [adminUI.url.apply(url => `${url}/auth/callback`)];
        },
      },
    });
```
While I believe it's quite convenient requirement, so it could look like this:

```TypeScript
    const webClient = userPool.addClient("Web", {
      callbackUrls: [adminUI.url.apply(url => `${url}/auth/callback`)]
    });
``` 

Please let me know if that makes sense for you to expose it.

### DIFF
--- a/platform/src/components/aws/cognito-user-pool-client.ts
+++ b/platform/src/components/aws/cognito-user-pool-client.ts
@@ -63,7 +63,7 @@ export class CognitoUserPoolClient extends Component implements Link.Linkable {
               "openid",
               "aws.cognito.signin.user.admin",
             ],
-            callbackUrls: this.callbackUrls,
+            callbackUrls: callbackUrls,
             supportedIdentityProviders: providers,
           },
           { parent },
--- a/platform/src/components/aws/cognito-user-pool-client.ts
+++ b/platform/src/components/aws/cognito-user-pool-client.ts
@@ -42,8 +42,8 @@ export class CognitoUserPoolClient extends Component implements Link.Linkable {
     }
 
     function normalizeCallbackUrls() {
-      if(!args.callbackUrls) return ["https://example.com"]
-      return output(args.callbackUrls)
+      if (!args.callbackUrls) return ["https://example.com"];
+      return output(args.callbackUrls);
     }
 
     function createClient() {
--- a/platform/src/components/aws/cognito-user-pool-client.ts
+++ b/platform/src/components/aws/cognito-user-pool-client.ts
@@ -32,12 +32,18 @@ export class CognitoUserPoolClient extends Component implements Link.Linkable {
 
     const providers = normalizeProviders();
     const client = createClient();
+    const callbackUrls = normalizeCallbackUrls();
 
     this.client = client;
 
     function normalizeProviders() {
       if (!args.providers) return ["COGNITO"];
       return output(args.providers);
+    }
+
+    function normalizeCallbackUrls() {
+      if(!args.callbackUrls) return ["https://example.com"]
+      return output(args.callbackUrls)
     }
 
     function createClient() {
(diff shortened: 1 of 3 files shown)


vimtor (changes requested):
thanks for the contribution @dimitor115 

problem now is that if the user skips `callbackUrls` you get:

```
| Error Web sst:aws:CognitoUserPoolClient → WebClient aws:cognito:UserPoolClient
creating Cognito User Pool Client (Web): operation error Cognito Identity Provider: CreateUserPoolClient, https response error StatusCode: 400, RequestID: d56889f8-e70d-4983-a441-97ce613d900e, InvalidOAuthFlowException: CallbackUrls can not be empty when code flow or implicit flow is selected
```

i proposed a change to keep it compatible with existing defaults, but yeah that https://example.com looks suspicious

maybe `callbackUrls` should just be required?

dimitor115:
Personally I believe it should be required, but as far as I see it's not required in [Pulumi](https://www.pulumi.com/registry/packages/aws/api-docs/cognito/userpoolclient/) as well. This makes me think, let's keep it backward compatible and fallback to **example.com**

platform/src/components/aws/cognito-user-pool-client.ts
```diff
@@ -57,7 +63,7 @@ export class CognitoUserPoolClient extends Component implements Link.Linkable {
               "openid",
               "aws.cognito.signin.user.admin",
             ],
-            callbackUrls: ["https://example.com"],
+            callbackUrls: this.callbackUrls,
```
vimtor:
```suggestion
            callbackUrls: callbackUrls,
```

vimtor (changes requested):

platform/src/components/aws/cognito-user-pool-client.ts
dimitor115: 🤦 I did not committed this one

platform/src/components/aws/cognito-user-pool-client.ts
```diff
@@ -32,6 +32,7 @@ export class CognitoUserPoolClient extends Component implements Link.Linkable {
 
     const providers = normalizeProviders();
     const client = createClient();
+    const callbackUrls = normalizeCallbackUrls();
```
vimtor: otherwise we get "Cannot access 'callbackUrls' before initialization"

```suggestion
    const callbackUrls = normalizeCallbackUrls();
    const client = createClient();
```

platform/src/components/aws/cognito-user-pool-client.ts
```diff
@@ -40,6 +41,11 @@ export class CognitoUserPoolClient extends Component implements Link.Linkable {
       return output(args.providers);
     }
 
+    function normalizeCallbackUrls() {
+      if(!args.callbackUrls) return ["https://example.com"]
+      return output(args.callbackUrls)
```
vimtor: missing formatting

```suggestion
      if (!args.callbackUrls) return ["https://example.com"];
      return output(args.callbackUrls);
```

vimtor (changes requested):

vimtor (approved):
good work!

```ts
const userPool = new sst.aws.CognitoUserPool("MyUserPool", {
  triggers: {
    preSignUp: {
      handler: "index.handler",
    },
  },
});

const client = userPool.addClient("Web", {
  callbackUrls: ["https://sst.dev/auth/callback"],
});
```

Image

PR #6274, opened 2025-12-10
Fix AWS bus example

### LINKED ISSUE #6270: AWS Bus example is outdated and broken
it uses `bus.handle` instead of `.subscribe` or it passes the wrong parameters to `event.builder`

vimtor:
closes #6270 

image


XX-NF-XX:
These event names probably should be the same:
```
bus.subscribe("Example", "./src/receiver.handler");
```
```
export const MyEvent = defineEvent(
  "app.myevent",
  z.object({
    foo: z.string(),
  }),
);
```

Most likely it even works with different names because event pattern (filter) is not deployed properly by SST.
When any event publish is triggered on a bus all the events of that bus are being triggered because all of them match the empty pattern
```
{
  "source": [{
    "prefix": ""
  }]
}
```
which in this case should be
```
{
  "detail-type": ["Example"]
}
```

vimtor:
@XX-NF-XX `Example` is just the name you give to the Pulumi resource, it has nothing to do with the subscription

XX-NF-XX:
@vimtor you've just hit the reset button in my brain 🙌🏼
Indeed the name is just the name of the subscription.
I've reviewed my whole event bus setup and found the solution for specific event triggering.
It was already there in TSdoc example.

examples/aws-bus/sst.config.ts
```diff
@@ -11,15 +16,12 @@ export default $config({
...
+      handler: "./src/publisher.handler",
       url: true,
       link: [bus],
     });
-    bus.subscribe("./src/receiver.handler");
 
-    return {
-      url: fn.url,
-    };
+    bus.subscribe("Example", "./src/receiver.handler");
```
XX-NF-XX, line 25: It would be nice to add this explicit event pattern match to the example.
```
bus.subscribe("Example", "./src/receiver.handler", {
  pattern: { detailType: ["app.myevent"] }
});
```

vimtor:
yep, you can see an example of how to subscribe to specific events in the docs:
https://sst.dev/docs/component/aws/bus/#subscribe

PR #6276, opened 2025-12-10
feat(vpc): add check if BastionInstanceProfile is provided and skip creating Role

mkilp:
Hi everyone,

similar to #6269 `createRole` is often restricted in bigger organizations but I would still like to use the bastion feature of vpcs.

This PR adds a check if a `iam.InstanceProfile` is provided and uses that instead of creating a new role and profile.

LMK if you need any changes. Appreciate all input.

Thanks,

Marv

### DIFF
--- a/platform/src/components/aws/vpc.ts
+++ b/platform/src/components/aws/vpc.ts
@@ -1239,33 +1239,62 @@ export class Vpc extends Component implements Link.Linkable {
             ),
           );
 
-          const role = new iam.Role(
-            `${name}BastionRole`,
-            {
-              assumeRolePolicy: iam.getPolicyDocumentOutput({
-                statements: [
+          // Check if iamInstanceProfile is provided via transform
+          const bastionTransform = args.transform?.bastionInstance;
+          const hasCustomInstanceProfile =
+            bastionTransform &&
+            typeof bastionTransform === "object" &&
+            "iamInstanceProfile" in bastionTransform;
+
+          const instanceProfile = hasCustomInstanceProfile
+            ? (() => {
+                const profile = (
+                  bastionTransform as {
+                    iamInstanceProfile: Input<string> | iam.InstanceProfile;
+                  }
+                ).iamInstanceProfile;
+                // If it's already an InstanceProfile resource, use it directly
+                if (profile instanceof iam.InstanceProfile) {
+                  return profile;
+                }
+                // Otherwise it's a string name, get the existing instance profile
+                return iam.InstanceProfile.get(
+                  `${name}BastionProfile`,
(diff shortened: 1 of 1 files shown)


mkilp:
@vimtor do you think I could get a review on this? We would love to use the bastion functionality of sst but IAM permissions are blocking us. This PR checks for transform but if you prefer I can also make it a component level setting aka

```ts
new sst.aws.Vpc("MyVPC", {
  bastion: {
    enabled: true,
    instanceProfile: <exampleArn>,
  }
});
```

vimtor:
@mkilp i will try to review both of these today!

vimtor:
@mkilp after reviewing the code i much rather prefer:

```ts
new sst.aws.Vpc("MyVPC", {
  bastion: {
    enabled: true, // probably not needed
    instanceProfile: <exampleArn>,
  }
});
```

mkilp:
@vimtor change is done. Added a normalize function to the bastion to be in line with the rest of SST. I also then defaulted "enabled" to true and modified parts of the app where the bastion component is checked!

LMK if you need changes.

platform/src/components/aws/vpc.ts
```diff
@@ -1239,33 +1270,54 @@ export class Vpc extends Component implements Link.Linkable {
             ),
           );
 
-          const role = new iam.Role(
-            `${name}BastionRole`,
-            {
-              assumeRolePolicy: iam.getPolicyDocumentOutput({
-                statements: [
+          const instanceProfile = bastion.instanceProfileName
```
vimtor: it's better to follow this pattern in pulumi:

```ts
return output(bastions.instanceProfileName).apply(instanceProfileName => {
  if (instanceProfileName) {
    return iam.InstanceProfile.get();
  }
  
  return new iam.InstanceProfile();
});
```

it avoids many bugs

vimtor (changes requested):

platform/src/components/aws/vpc.ts
mkilp: Should we apply this everywhere for iam getters? In fargate.ts for example we do

```ts
if (args.taskRole)
    return iam.Role.get(`${name}TaskRole`, args.taskRole, {}, { parent });
``` 

(https://github.com/anomalyco/sst/blob/dev/platform/src/components/aws/fargate.ts#L871)
which is why i assumed a direct getter would be fine.
Happy to rewrite it to the output pattern tho

vimtor (approved):

<details>

<summary>just applied the minor fixed suggested but everything else is currently working</summary>

Image

Image

</details>

PR #6277, opened 2025-12-10
fix(router): respect path segment boundaries in route matching

azzzy:
## Problem

The Router's CloudFront function uses simple `startsWith()` matching for paths, causing incorrect route matching when paths share a common prefix.

For example:
- Route `/travel-plan` incorrectly matches `/travel-plans`
- Route `/api` incorrectly matches `/api-docs`

## Solution

Updated the path matching logic to check for proper path segment boundaries. A path now only matches if:
- The URI exactly equals the path, OR
- The character after the matched prefix is `/`, OR  
- The route is the root path `/`

## Changes

- Modified `CF_ROUTER_INJECTION` in `platform/src/components/aws/router.ts` (line ~1541)
- Added unit tests in `platform/test/components/router-path-matching.test.ts`

## Testing

The test file duplicates the path matching logic rather than importing it because the actual implementation lives inside a CloudFront function string literal. This allows testing the logic in isolation without complex string parsing.

Verified with live CloudFront deployment:
- `/travel-plans` correctly does NOT match route `/travel-plan`
- `/travel-plan/abc` correctly matches route `/travel-plan`
- All existing routes continue to work as expected

## Breaking Changes

None. This is a bug fix that corrects unexpected behavior.

### DIFF
--- a/platform/src/components/aws/router.ts
+++ b/platform/src/components/aws/router.ts
@@ -1538,7 +1538,7 @@ async function handler(event) {
         || (host.includes("*") && new RegExp(host).test(requestHostRegexPattern));
       if (!hostMatches) return;
 
-      const pathMatches = event.request.uri.startsWith(path);
+      const pathMatches = event.request.uri.startsWith(path) && (event.request.uri === path || event.request.uri[path.length] === '/' || path === '/');
       if (!pathMatches) return;
 
       match = {
--- a/platform/test/components/router-path-matching.test.ts
+++ b/platform/test/components/router-path-matching.test.ts
@@ -1,0 +1,138 @@
+import { describe, it, expect } from "vitest";
+
+/**
+ * Tests for Router path matching logic
+ *
+ * The Router CloudFront function uses path matching to route requests.
+ * This test verifies the path segment boundary logic to ensure routes
+ * only match at proper path boundaries.
+ */
+describe("Router path matching", () => {
+  /**
+   * Simulates the path matching logic from the CloudFront function
+   * This is extracted from platform/src/components/aws/router.ts
+   */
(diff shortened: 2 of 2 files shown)


vimtor:
thanks for the contribution @azzzy 

i'm trying to reproduce this with a nextjs site but i don't see the issue:

CleanShot 2026-01-11 at 17 52 58@2x

CleanShot 2026-01-11 at 17 52 45@2x

is there an specific use case where this fails?

azzzy:
@vimtor , are you sure this isn't just Next.js client routing taking over?
If you do a hard refresh on `/travel-plans` is it working still?

azzzy:
@vimtor , I've created a minimal reproduction repo: https://github.com/azzzy/sst-routing-bug-reproduction
The setup:
- A Lambda function at `/travel-plan`
- A StaticSite (React SPA) on the same domain

To reproduce:
Open the SPA, click on the `Travel plans` link and once on the  `/travel-plans` route  hard refresh (Cmd+Shift+R / Ctrl+Shift+R)
The bug: `/travel-plans` incorrectly routes to the `/travel-plan` Lambda handler instead of the React site.

I have it deployed here - https://sst-routing-pr.sourcelab.dev

vimtor (approved):
approved!

thanks for the reproduction @azzzy 

<details>

spa-home

spa-travel-plan

spa-travel-plans

i also tried different examples like the `aws-nextjs` one to ensure this doesn't break existing routing:

nextjs-home

nextjs-travel-info

nextjs-travel-plan

nextjs-travel

</details>

vimtor:
fixed in [v3.18.0](https://github.com/anomalyco/sst/releases/tag/v3.18.0)

PR #6278, opened 2025-12-11
fix(vpc): add missing parent parameter to EipAssociation in createNat…

desangelo:
When creating a VPC in a secondary region using a custom provider, the `EipAssociation` for NAT instances was not inheriting the provider, causing it to attempt operations in the default region.

eg.

```ts
const secondaryRegionProvider = new aws.Provider("east-2", {
    region: "us-east-2"
})
```

```ts
new sst.aws.Vpc(`secondary-region`, {
    bastion: true,
    nat: {
        type: "ec2"
    },
}, {
    provider: secondaryRegionProvider,
})
```

Changes:

1. Adds { parent: self } to the EipAssociation creation in `vpc.ts` to properly inherit the provider
2. Adds `aws:ec2/eipAssociation:EipAssociation` to the exclusion list in `component.ts`(required for any resource using { parent: self } that doesn't need physical name prefixing)

### DIFF
--- a/platform/src/components/aws/vpc.ts
+++ b/platform/src/components/aws/vpc.ts
@@ -1018,10 +1018,13 @@ export class Vpc extends Component implements Link.Linkable {
               ),
             );
 
-            new ec2.EipAssociation(`${name}NatInstanceEipAssociation${i + 1}`, {
+            new ec2.EipAssociation(`${name}NatInstanceEipAssociation${i + 1}`, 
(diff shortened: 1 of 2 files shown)


vimtor:
duplicate of #6087

desangelo:
I was looking if there was anything in the works but seemed I missed that. I'll close this out then. Thanks

Issue #6282, opened 2025-12-14
Feature request: add nodes for VPC NAT Instance & Bastion Security Groups


danielsharvey:
Further to #5538, request the addition of the VPC NAT Instance & Bastion Security Groups to the output nodes.

This will allow reference to these Security Groups in SecurityGroupRule's (etc).

PR #6283, opened 2025-12-14
feat(aws/vpc): expose security groups for NAT and bastion instances

danielsharvey:
- Add natSecurityGroup and bastionSecurityGroup to Vpc component properties
- Return security group references from createNatInstances() and createBastion()
- Update component state to track and persist security groups
- Add documentation for new security group outputs

This allows users to reference and modify security groups for NAT and bastion instances when needed for custom networking configurations.

Addresses #6282.


danielsharvey:
I omitted reference handling - fixing now.

PR #6243, opened 2025-11-27
feat(aws): add support for Node.js 24.x runtime

casperiv0:
First contribution to SST, so let me know if there should be more added here 🙏.

PR #6288, opened 2025-12-17
add node 24 as a lambda runtime

### LINKED ISSUE #6287: Add nodejs24.x as lambda runtime
Since AWS now supports nodejs 24 lts, we should add it as a runtime target.

I would like to update esbuild to 0.27.2 also because it ads ES2024 as a build target and because this is the node 24 es target we would have full access to all node 24 stuff.

We are running on esbuild 0.21.5 which was released back in 2024
i tried upgrading to the new esbuild and it seems to be working.
But that should probably be a breaking change since it might cause problems.

I will submit PR for this shortly

leon:
Since AWS now supports nodejs 24 lts, we should add it as a runtime target.

I would like to update esbuild to 0.27.2 also because it ads ES2024 as a build target and because this is the node 24 es target we would have full access to all node 24 stuff.

We are running on esbuild 0.21.5 which was released back in 2024
i tried upgrading to the new esbuild and it seems to be working.
But that should probably be a breaking change since it might cause problems.

Fixed #6287

### DIFF
--- a/pkg/runtime/node/build.go
+++ b/pkg/runtime/node/build.go
@@ -23,6 +23,7 @@ var forceExternal = []string{
 }
 
 var targetMap = map[string]esbuild.Target{
+	"nodejs24.x": esbuild.ES2023,
 	"nodejs22.x": esbuild.ES2023,
 	"nodejs20.x": esbuild.ES2023,
 	"nodejs18.x": esbuild.ES2022,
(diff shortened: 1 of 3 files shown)


leon:
@vimtor Do you think we could get this one in the new release also?
it's simple and will not break anything.

vimtor:
duplicate of #6243

PR #6289, opened 2025-12-18
Fix AWS `StaticSite` file upload order

vimtor:
when checking what files to upload for static sites we do it in the appropriate order (html then the rest) to avoid deploying a new html that references a not-yet-uploaded file

however, there's still a problem because all the files are uploaded in parallel, so it might still happen that the html file is uploaded before the rest

this pr implements a fix to ensure that the non-html files are always uploaded first

<details>

<summary>i've tested against some of the AWS `StaticSite` examples and it worked flawlessly</summary>

CleanShot 2025-12-18 at 18 06 44@2x
CleanShot 2025-12-18 at 18 06 51@2x
CleanShot 2025-12-18 at 18 06 47@2x

</details>


pkg/server/resource/aws-bucket-files.go
```diff
@@ -118,6 +119,36 @@ func (r *BucketFiles) upload(client *s3.Client, bucketName string, files []Bucke
...
+	// referencing assets that haven't been uploaded yet.
+	var htmlFiles, nonHtmlFiles []BucketFile
+	for _, file := range files {
+		oldFile, exists := oldFilesMap[file.Key]
+		if exists && oldFile.Hash != nil && *oldFile.Hash == *file.Hash &&
+			oldFile.CacheControl == file.CacheControl &&
+			oldFile.ContentType == file.ContentType {
+			continue
+		}
+		if strings.HasSuffix(file.Key, ".html") {
```
jamesgibbons92, line 133: Would it be better just to break the files into these 2 groups instead of html and nonHtml:
- updatedFiles: Files that have an existing key and have a different hash <--- upload these **last**
- createdFiles: Files that do not have an existing key <---- upload these **first**

This should still solve the race issue, and avoids having to bring in and handle specific file extension logic.

vimtor: i believe this would break for new deployments as there are no existing keys

also, i believe that the JS code treads HTML case specifically too, so it should be fine

PR #6298, opened 2025-12-25
fix: improve import error output with detailed diff information

### LINKED ISSUE #6285: Failing to import a Pulumi resource due to unhelpful output
In the documentation on how to import a Pulumi resource here: https://sst.dev/docs/import-resources/
the documentation says that I should attempt the import and change the fields that the import tells me to change.  

I'm trying to import a Cloudfront Distribution and the very helpful message I get is:
```
inputs to import do not match the existing resource: [comment origins]
```
Since origins is a massive complex JSON as detailed here:
https://www.pulumi.com/registry/packages/aws/api-docs/cloudfront/distribution/#distributionorigin
That comment is worse than useless and I have been fiddling with the values for a while now failing to get anywhere.

Is there any approach that would allow SST to print a better error message or allow dropping into a Pulumi console to allow debugging the error, or maybe doing a manual import that might suggest how to debug the issue?  Or any kind of documentation which might help explain what to do?

Ayush2k02:
## Summary

Fixes #6285

This PR significantly improves the error output when importing Pulumi resources fails due to mismatched inputs.

## Problem

When users attempted to import resources (especially complex ones like CloudFront Distributions), they would get unhelpful error messages like:

```
inputs to import do not match the existing resource: [comment origins]
```

For complex fields like `origins` (which is a massive nested JSON structure), this message provided **zero useful information** about what was actually different. Users couldn't tell:
- What value currently exists in the cloud
- What value they're trying to set
- What specifically is different

## Solution

Enhanced the import diff display in `cmd/sst/mosaic/ui/ui.go` to provide much more helpful information:

### Changes Made

1. **Pretty-print JSON values** using `json.MarshalIndent` instead of compact JSON
2. **Show both values**:
   - The existing value in the cloud resource (`diff.Old`)
   - The value being attempted in the code (`diff.New`)
3. **Handle complex vs simple values differently**:
   - Simple values (strings, numbers): Display on one line
   - Complex objects/arrays: Display with proper indentation AND show what you're trying to set
4. **Add helpful hint** pointing to `.sst/log/pulumi.log` for even more detailed debugging

### Example Output

**Before:**
```
Set the following:
   - `origins: [massive json blob]`
```

**After:**
```
Set the following:
   - `origins: {
       "domainName": "example.com",
       "originId": "S3-example",
       ...
     },`
     Trying to set: {
       "domainName": "different.com",
       "originId": "S3-different",
       ...
     }

   For more details, check: .sst/log/pulumi.log
```

## Impact

Users can now:
- ✅ See exactly what value exists vs what they're trying to set
- ✅ Compare complex JSON structures side-by-side
- ✅ Understand what needs to change in their code
- ✅ Know where to find even more detailed logs

This transforms import errors helpful for debugging.

### DIFF
--- a/cmd/sst/mosaic/ui/ui.go
+++ b/cmd/sst/mosaic/ui/ui.go
@@ -477,19 +477,40 @@ func (u *UI) Event(unknown interface{}) {
 						u.println(TEXT_NORMAL.Render("\n\nSet the following:"))
 					}
 					for _, diff := range importDiffs {
-						value, _ := json.Marshal(diff.Old)
+						oldValue, _ := json.MarshalIndent(diff.Old, "     ", "  ")
+						newValue, _ := json.MarshalIndent(diff.New, "     ", "  ")
 						if diff.Old == nil {
-							value = []byte("undefined")
+							oldValue = []byte("undefined")
 						}
+						if diff.New == nil {
+							newValue = []byte("undefined")
+						}
+
+						// For simple values (not objects/arrays), show them on one line
+						oldStr := string(oldValue)
+						newStr := string(newValue)
+						isComplex := strings.Contains(oldStr, "\n") || strings.Contains(newStr, "\n")
+
 						u.print(TEXT_NORMAL.Render("   - "))
 						if isSSTComponent {
-							u.print(TEXT_INFO.Render("`args." + string(diff.Input) + " = " + string(value) + ";`"))
+							if isComplex {
+								u.println(TEXT_INFO.Render("`args." + string(diff.Input) + " = " + oldStr + ";`"))
+								u.println(TEXT_DIM.Render("     Trying to set: " + newStr))
+							} else {
+								u.println(TEXT_INFO.Render("`args." + string(diff.Input) + " = " + oldStr + ";`"))
+							}
 						}
 						if !isSSTComponent {
-							u.print(TEXT_INFO.Render("`" + string(diff.Input) + ": " + string(value) + ",`"))
+							if isComplex {
+								u.println(TEXT_INFO.Render("`" + string(diff.Input) + ": " + oldStr + ",`"))
+								u.println(TEXT_DIM.Render("     Trying to set: " + newStr))
+							} else {
+								u.println(TEXT_INFO.Render("`" + string(diff.Input) + ": " + oldStr + ",`"))
+							}
 						}
 						u.blank()
 					}
+					u.println(TEXT_DIM.Render("   For more details, check: .sst/log/pulumi.log"))
 				} else {
 					u.blank()
 				}


vimtor:
thanks for your contribution @Ayush2k02

could you provide a sample `sst.config.ts` with the original issue so i can more easily review this pr?

thank you

Ayush2k02:
Thanks for reviewing! Here's a sample `sst.config.ts` that demonstrates the original issue:

## Sample Configuration

```typescript

export default $config({
  app(input) {
    return {
      name: "my-app",
      home: "aws",
    };
  },
  async run() {
    const distribution = new aws.cloudfront.Distribution("MyDistribution", {
      enabled: true,
      comment: "My CloudFront distribution",
      origins: [{
        domainName: "my-bucket.s3.amazonaws.com",
        originId: "S3-my-bucket",
        customOriginConfig: {
          httpPort: 80,
          httpsPort: 443,
          originProtocolPolicy: "https-only",
          originSslProtocols: ["TLSv1.2"],
        }
      }],
      defaultCacheBehavior: {
        targetOriginId: "S3-my-bucket",
        viewerProtocolPolicy: "redirect-to-https",
        allowedMethods: ["GET", "HEAD"],
        cachedMethods: ["GET", "HEAD"],
        forwardedValues: {
          queryString: false,
          cookies: { forward: "none" },
        },
        minTtl: 0,
        defaultTtl: 3600,
        maxTtl: 86400,
      },
      restrictions: {
        geoRestriction: {
          restrictionType: "none",
        },
      },
      viewerCertificate: {
        cloudfrontDefaultCertificate: true,
      },
    }, {
      import: "E2QWRUHAPOMQZL"  // existing CloudFront distribution ID
    });
  },
});
```

## Reproducing the Issue

When you try to import an existing CloudFront distribution whose `origins` configuration doesn't match exactly (e.g., existing resource uses `s3OriginConfig` but code specifies `customOriginConfig`, or has different SSL protocols), you'd see:

**Before this PR:**
```
inputs to import do not match the existing resource: [comment origins]
```

This tells you fields mismatch but gives no details about what values exist versus what you're setting.

**After this PR:**
```
inputs to import do not match the existing resource

Expected (from cloud):
{
  "comment": "Production CloudFront Distribution",
  "origins": [
    {
      "domainName": "my-bucket.s3.amazonaws.com",
      "originId": "S3-my-bucket",
      "s3OriginConfig": {
        "originAccessIdentity": "origin-access-identity/cloudfront/ABCDEFG"
      }
    }
  ]
}

Got (from code):
{
  "comment": "My CloudFront distribution",
  "origins": [
    {
      "domainName": "my-bucket.s3.amazonaws.com",
      "originId": "S3-my-bucket",
      "customOriginConfig": {
        "httpPort": 80,
        "httpsPort": 443,
        "originProtocolPolicy": "https-only",
        "originSslProtocols": ["TLSv1.2"]
      }
    }
  ]
}

For additional details, see .sst/log/pulumi.log
```

Now you can clearly see the mismatch: the existing resource uses `s3OriginConfig` with an origin access identity, but the code specifies `customOriginConfig` with protocol settings.

Let me know if you need any other clarification!

Issue #6305, opened 2026-01-03
Tunnel is configured with incorrect IP address


danielsharvey:
The tunnel is configured with the incorrect IP address.

The value `bastion.publicIp` is referenced rather then the Elastic IP address(es).

The [Pulumi docs](https://www.pulumi.com/registry/packages/aws/api-docs/ec2/instance/) state:
```
Public IP address assigned to the instance, if applicable. 
**NOTE**: If you are using an `aws.ec2.Eip` with your instance, you should 
refer to the EIP's address directly and not use `publicIp` as this field will 
change after the EIP is attached.
```

This results in the tunnel forwarding to the wrong IP address and does not work. 

This seems to be experienced by others e.g. #5657.

The relevant code is:
https://github.com/anomalyco/sst/blob/e4159bbeca0dd23043a70b22c366d930067a15f2/platform/src/components/aws/vpc.ts#L679-L700

danielsharvey:
See https://github.com/anomalyco/sst/issues/6309

GuilhermeCunha:
I'm being affected by this issue. Could you find a solution @danielsharvey ?

danielsharvey:
> I'm being affected by this issue. Could you find a solution @danielsharvey ?

I have proposed a fix in https://github.com/anomalyco/sst/pull/6306 - I have patched my personal install whilst awaiting the PR review.

jamesgibbons92:
This was somewhat resolved in #5820 
However the fix isn't actually available yet in a release version, as the release pipelines are broken and haven't been fixed yet.
I've made further changes in #6341 to resolve the remaining bugs with tunnel.
I will close this issue as it is a duplicate, if you have any further issues not addressed by the PR can you mention them in issue #6309

PR #6306, opened 2026-01-03
fix(vpc): use elastic IP for tunnel connection when available

danielsharvey:
When elastic IPs are configured, use the first elastic IP address for the tunnel connection instead of the bastion instance's public IP. This ensures the tunnel uses the stable NAT instance IP (which gets replaced by the EIP after creation) rather than a potentially changing public IP address.

Addresses #6305.

### DIFF
--- a/platform/src/components/aws/vpc.ts
+++ b/platform/src/components/aws/vpc.ts
@@ -680,14 +680,17 @@ export class Vpc extends Component implements Link.Linkable {
       self.registerOutputs({
         _tunnel: all([
           self.bastionInstance,
+          self.elasticIps,
           self.privateKeyValue,
           self._privateSubnets,
           self._publicSubnets,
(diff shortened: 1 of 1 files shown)


jamesgibbons92:
@danielsharvey 
The bastion is not given an EIP when the nat is managed, so this doesn't work in that case. 
The eip should only be used if nat instances are used, as one of the instances is used as the bastion instead of a standalone ec2.
This was resolved in https://github.com/anomalyco/sst/pull/5820
But you've rightly pointed out some additional issues when importing the VPC in [this issue](https://github.com/anomalyco/sst/issues/6309). I'll fix these in a PR and then  close down some of these duplicate PR / issues. Thank you

jamesgibbons92:
Closing as issues should all be resolved by #6341

PR #6318, opened 2026-01-11
fix: race condition in basic mode server startup

### LINKED ISSUE #3990: sst dev --mode=basic does not work
**Steps to reproduce:**

Follow the steps here:
[https://sst.dev/docs/start/aws/api](https://sst.dev/docs/start/aws/api)

Until you get to the 
`npx sst dev`

Instead run 
`npx sst dev --mode=basic`

**What do you expect happens:**
The api to be run in dev mode

**What happens:**
Nothing. I get a single blank line in my console then nothing until I ctrl-c.

**Version:**
3.1.71

vsonti23:
Race condition causing sst dev --mode=basic to hang on startup

In basic mode, the UI client was attempting to connect to the HTTP server before it was ready to accept connections, resulting in "connection refused" errors and causing the application to hang
indefinitely.

Root cause: server.Start() runs in a goroutine while CmdUI() immediately tries to connect, creating a race between server initialization and client connection.

Solution:

• Added a Ready channel to the Server struct that signals when the server is listening
• Modified basic mode to wait on <-server.Ready before starting the UI
• Close the Ready channel after net.Listen() succeeds (and on error to prevent infinite hangs)

fixes #3990

### DIFF
--- a/cmd/sst/mosaic.go
+++ b/cmd/sst/mosaic.go
@@ -302,6 +302,7 @@ func CmdMosaic(c *cli.Cli) error {
 
 	if mode == "basic" {
 		wg.Go(func() error {
+			<-server.Ready
 			return CmdUI(c)
 		})
 	}
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -23,9 +23,10 @@ import (
 )
 
 type Server struct {
-	Port int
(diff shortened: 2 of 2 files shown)


vimtor (approved):
thanks for your contribution @vsonti23 

<details>

Image

Image

Image

</details>

PR #6324, opened 2026-01-12
Allow esbuild options to be configured in Worker runtime

### LINKED ISSUE #6293: Cloudflare Worker: Allow overriding hardcoded esbuild options
The Worker component allows setting esbuild options via `transform.server.build.esbuild`, but many of them are silently ignored because they're hardcoded in the runtime: https://github.com/sst/sst/blob/dev/pkg/runtime/worker/worker.go#L113
Can we please make these hardcoded options actually configurable via `transform.server.build.esbuild`? @vimtor

lucassarcanjo:
Previously, KeepNames, Target, and Sourcemap options were hardcoded and ignored user configuration. This allows users to override these options via transform.server.build.esbuild to resolve issues like SSR hydration problems in TanStack Start applications.

Note: KeepNames defaults to true. Only override if user explicitly disables it. For Target and Sourcemap, check non-zero values as zero represents "not set" for these enums.

Fixes #6293

### DIFF
--- a/pkg/runtime/worker/worker.go
+++ b/pkg/runtime/worker/worker.go
@@ -138,6 +138,19 @@ func (w *Runtime) Build(ctx context.Context, input *runtime.BuildInput) (*runtim
 		},
 	}
 
+	// Apply user overrides for configurable esbuild options
+	if !build.ESBuild.KeepNames {
+		options.KeepNames = false
+	}
+
+	if build.ESBuild.Target != 0 {
+		options.Target = build.ESBuild.Target
+	}
+
+	if build.ESBuild.Sourcemap != 0 {
+		options.Sourcemap = build.ESBuild.Sourcemap
+	}
+
 	w.lock.RLock()
 	buildContext, ok := w.contexts[input.FunctionID]
 	w.lock.RUnlock()


vimtor:
thanks for your contribution @lucassarcanjo 

do you have any good example/use case for me to try this out?

lucassarcanjo:
I used `examples/cloudflare-worker` and added custom properties like `keepNames` in the build configuration:

<details>
  <summary>Example of sst.config.ts</summary>

  ```ts

export default $config({
  app(input) {
    return {
      name: "cloudflare-worker",
      home: "cloudflare",
    };
  },
  async run() {
    const bucket = new sst.cloudflare.Bucket("MyBucket");
    const worker = new sst.cloudflare.Worker("MyWorker", {
      handler: "./index.ts",
      link: [bucket],
      url: true,

      // configuration to be tested here
      build: {
        esbuild: {
          keepNames: false,
        },
      },
    });

    return {
      api: worker.url,
    };
  },
});

  ```
</details>

Added a log to get all configurations from worker.go and got the correct values:

`slog.Info("building with options", "options", options)`

image

vimtor (changes requested):
thanks for your contribution @lucassarcanjo 

the `keepNames` option works without any issue

however the sourcemap and target options don't seem to update when you set the config probably because of how the json unmarshalling works from the raw string to the go enum

we either fix them or remove their overriding

vimtor:
closed in favor of #6409

PR #6328, opened 2026-01-14
Add `Ai` Cloudflare component

saibayadon:
### Summary

This PR adds the `Ai` component for Cloudflare Workers making it a first-class Component alongside existing ones like Bucket and Worker.

The main driver for this change is that currently, to use AI Bindings you have to manually update the bindings in the transform:

```typescript
const api = new sst.cloudflare.Worker("Example", {
  handler: "./src/index.ts",
  enviroment: {
    SOME_ENV: "value",
  },
  transform: {
    worker: (args) => {
      args.bindings = $output(args.bindings).apply((bindings) => [
        ...(bindings ?? []),
        { name: "AI", type: "ai" },
      ]);
    },
  },
});
```

Which, I mean - works; But doesn't look great and has to be duplicated in every worker configuration and more importantly doesn't allow you to access it from `Resource` and you have to manually ensure it's present on `ENV` instead.

### Changes
- New `Ai` component (`platform/src/components/cloudflare/ai.ts`)
- Binding types (`platform/src/components/cloudflare/binding.ts`): Add AiBinding type definition
- Updated Example (`examples/cloudflare-hono/`)

### Usage

```typescript
const ai = new sst.cloudflare.Ai("ai");
const worker = new sst.cloudflare.Worker("Worker", {
  link: [ai],
  handler: "index.ts",
});
```

Then access via `Resource.ai` in the worker script:

```typescript
const result = await Resource.ai.run("@cf/meta/llama-3-8b-instruct", {
  prompt: "What is the origin of the phrase 'Hello, World'",
});
```

### Notes
I updated the hono example to add this, which necessitated the update to the latest `worker-types` package (`4.20260114.0`). I'm not 100% sure if this is something that needs to be updated someplace else to ensure that SST installs the appropriate version - so let me know if there's something that I need to update for that.

I also wasn't sure if there was any tests I needed to update, so apologies if that's missing as well.


saibayadon:
Screenshot 2026-01-14 at 3 08 06 AM

Screenshot 2026-01-14 at 3 08 03 AM

Attached are screenshots of a local `sst deploy` with this change.

vimtor (approved):
thanks for your contribution @saibayadon 

did some minor change but works like a charm

<details>

Image

</details>

yashwanth2804:
is documentation ready for this , i dont find in sst dev site

vimtor:
@yashwanth2804 you are right!

fixed in https://github.com/anomalyco/sst/pull/6533

Issue #6307, opened 2026-01-05
update bridge-task image URL after GitHub org rename


jgrant-sms:
## Summary

  The hardcoded `bridge-task` container image URL references the old GitHub organization (`sst/sst`), which is no longer accessible after the org was renamed to `anomalyco`.

  ## Problem

  ECS tasks fail in dev mode because the image at `ghcr.io/sst/sst/bridge-task:20241224005724` returns "denied".

  ## Solution

  Update references from `ghcr.io/sst/sst/bridge-task` to `ghcr.io/anomalyco/sst/bridge-task` in:

  - `platform/src/components/aws/task.ts` (runtime image reference)
  - `platform/scripts/build` (build/push target)

  ## Verification

  ```bash
  # New location - accessible
  docker manifest inspect ghcr.io/anomalyco/sst/bridge-task:20241224005724

  # Old location - denied
  docker manifest inspect ghcr.io/sst/sst/bridge-task:20241224005724

PR #6330, opened 2026-01-15
fix: Docker image pull for bridge-task

seannetlife:
It appears that the organisation rename breaks the ghcr.io pulls. :( 

Tasks are failing in development;

image


vimtor:
duplicate #6307

PR #6334, opened 2026-01-16
Bumping default version of OpenNext to 3.9.8

leanderiversen:
The current default version (3.6.6) was released in June, so I think it's time to bump this to the latest version.

### DIFF
--- a/platform/src/components/aws/nextjs.ts
+++ b/platform/src/components/aws/nextjs.ts
@@ -11,7 +11,7 @@ import { isALteB } from "../../util/compare-semver.js";
 import { Plan, SsrSite, SsrSiteArgs } from "./ssr-site.js";
 import { Bucket } from "./bucket.js";
 
-const DEFAULT_OPEN_NEXT_VERSION = "3.6.6";
+const DEFAULT_OPEN_NEXT_VERSION = "3.9.8";
 
 type BaseFunction = {
   handler: string;


vimtor (approved):
thanks for your contribution @leanderiversen 

<details>

<summary>i tried deploying multiple AWS Next.js examples and they seem to be working</summary>

Image
Image
Image
Image
Image

</details>

any specific reason why you wanted the new OpenNext version? i'll try to get this merged but given how popular the Next.js construct is i'll need more testing

leanderiversen:
Thanks for the testing, @vimtor. For my projects, I always manually specify the latest version of OpenNext, but considering someone new to SST, and potentially Next itself, it would be beneficial for them to use the latest version of OpenNext, considering the amount of patching that has been done for recent versions.

vimtor:
closes #6348

onhate:
Hey! Just a heads up - OpenNext 3.9.8 has a build issue that breaks projects using packages with native modules (like `@swc/core` from `next-intl`).

The error looks like this:
```
✘ [ERROR] No loader is configured for ".node" files: @swc/core-darwin-arm64/swc.darwin-arm64.node
✘ [ERROR] Could not resolve "@swc/wasm"
```

The root cause was that the compiled Next.js config patch wasn't properly inlining `require.resolve` calls, which caused esbuild to fail when trying to bundle dependencies with native bindings.

This was fixed in 3.9.9: https://github.com/opennextjs/opennextjs-aws/pull/1077

Would be great to bump the default OpenNext version to 3.9.9+ to avoid this regression affecting other users.

onhate:
As a temporary workaround, you can pin to the latest 3.x in your SST config:

```ts
new sst.aws.Nextjs('Web', {
  buildCommand: 'pnpm dlx @opennextjs/aws@3 build',
  // ...
});
```

This will use 3.9.14 (or latest) instead of the broken 3.9.8.

vimtor:
@onhate what happens if you specify the [`openNextVersion`](https://sst.dev/docs/component/aws/nextjs/#opennextversion)?

onhate:
> @onhate what happens if you specify the [`openNextVersion`](https://sst.dev/docs/component/aws/nextjs/#opennextversion)?

it ends up I set it as `oopenNextVesion: ^3` and it works

vimtor:
@onhate can you check version [v3.18.2](https://github.com/anomalyco/sst/releases/tag/v3.18.2)? 

it should work by default without specifying the `openNextVersion` option

onhate:
> @onhate can you check version [v3.18.2](https://github.com/anomalyco/sst/releases/tag/v3.18.2)?
> 
> it should work by default without specifying the `openNextVersion` option

it worked!

PR #6335, opened 2026-01-17
fix: ensure dev resources reliably appear in multiplexer on startup

### LINKED ISSUE #6011: Inconsistent startup of `dev` resources (`sst.aws.StaticSite`, `sst.aws.Service`, `sst.x.DevCommand`...)

execute008:
- Subscribe to CompleteEvent before deployer starts to avoid missing events
- Use PostEventWait instead of PostEvent to prevent event dropping when queue is full
- Add procChan buffer and ready channel for proper initialization synchronization
- Extract helper functions for processing CompleteEvent in multi/mono modes

Fixes #6011


vimtor:
thanks for your contribution

closed in favour of https://github.com/anomalyco/sst/pull/6402

PR #6344, opened 2026-01-21
feat: add deploy.after hook for post-deployment actions

smorimoto:
## Summary

Add a `deploy.after` hook that runs after `sst deploy` completes. This is useful for:
- Sending notifications (Slack, Discord, etc.)
- Triggering follow-up actions
- Logging deployment results

### Changes

- **New `hooks.go`**: Core hook execution logic with timeout support (default 30s, configurable via `SST_HOOK_TIMEOUT`)
- **UI integration**: Hook start, complete, and error events are displayed in the CLI
- **TypeScript types**: `DeployResult`, `DeployError`, `DeployHooks`, and `Hooks` interfaces
- **Documentation**: JSDoc with examples in `config.ts`

### Hook Result

The `after` hook receives a `DeployResult` object containing:
- `success` — Whether the deployment completed without errors
- `errors` — Array of errors if any occurred  
- `outputs` — The outputs from your `run` function
- `resources` — Number of resources in the stack
- `app.name` — The app name
- `app.stage` — The stage name

### Example

```ts title="sst.config.ts"
hooks: {
  deploy: {
    async after(result) {
      await fetch("https://example.com/webhook", {
        method: "POST",
        body: JSON.stringify({
          text: `Deployed ${result.app.name} to ${result.app.stage}`
        })
      });
    }
  }
}
```

### Notes

- Hook only runs after `sst deploy`, not during `sst dev` or `sst diff`
- Hook errors will cause the deployment to fail
- Hook will not run if deployment is interrupted (Ctrl+C)

## Test plan

- [ ] Deploy with a working hook and verify it runs after deployment
- [ ] Deploy with a hook that throws an error and verify deployment fails
- [ ] Deploy with a hook that uses try-catch and verify deployment succeeds
- [ ] Verify hook does not run during `sst dev` or `sst diff`
- [ ] Test timeout with `SST_HOOK_TIMEOUT=5` and a slow hook

### DIFF
--- a/cmd/sst/mosaic/ui/ui.go
+++ b/cmd/sst/mosaic/ui/ui.go
@@ -505,6 +505,15 @@ func (u *UI) Event(unknown interface{}) {
 			}
 		}
 		u.blank()
+	case *project.HookStartEvent:
+		u.printEvent(TEXT_INFO, "Hook", fmt.Sprintf("Running %s hook...", evt.Hook))
+
+	case *project.HookCompleteEvent:
+		u.printEvent(TEXT_SUCCESS, "Hook", fmt.Sprintf("%s completed", evt.Hook))
+
+	case *project.HookErrorEvent:
+		u.printEvent(TEXT_DANGER, "Hook", fmt.Sprintf("%s failed: %s", evt.Hook, evt.Error))
+
 	case *cloudflare.WorkerBuildEvent:
 		if len(evt.Errors) > 0 {
 			u.printEvent(TEXT_DANGER, "Build Error", u.functionName(evt.WorkerID)+" "+strings.Join(evt.Errors, "\n"))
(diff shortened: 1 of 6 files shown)


michaelrambeau:
A deploy hook would be a great improvement!

Maybe the configuration option to setup the hook should be mentioned in the documentation: https://sst.dev/docs/reference/config

Source: https://github.com/anomalyco/sst/blob/dev/platform/src/config.ts

smorimoto:
The documentation should probably be generated from the JSDoc of that file!

vimtor:
thanks for your contribution @smorimoto

this is a great change, probably it could replace #6177 as a more elegant solution.

a couple of thoughts:
1. have you thought how this relates to the `console.workflow` config? in principle something similar to hooks could be implemented as a workflow. you need to use the sst console, though, so maybe it makes sense to uplift the `workflow` property to work outside the console?
2. i believe hook errors should fail the deployment. someone will probably request it, so maybe failing the deployment is a better default. we can always let the user try-catch if they want that behaviour ([similar to this workflow example](https://sst.dev/docs/reference/config/#pullrequestevent:~:text=const%20%7B%20exitCode%20%7D%20%3D%20await%20%24%60pnpm%20test%60.nothrow()%3B))
3. imagine you want to send a Slack message for the deployment. you'll need some API secret, right? how do you provide it? if i'm not mistaken, with the current setup you'll need to rely on env variables instead of SST secrets

PR #6347, opened 2026-01-21
fix(Vpc): use self.natInstances in registerOutputs to fix Vpc.get() with bastion

### LINKED ISSUE #6346: Vpc.get() fails with "Cannot access 'natInstances' before initialization" (regression in v3.17.26)
## Bug Description

When using `sst.aws.Vpc.get()` to reference an existing VPC that was created with `bastion: true`, deployment fails with:

```
ReferenceError: Cannot access 'natInstances' before initialization
    at file://.../.sst/platform/src/components/aws/vpc.ts:693:19
```

## Root Cause

In `platform/src/components/aws/vpc.ts`, the `registerOutputs()` function at line 693 references the local variable `natInstances`:

```typescript
ip: natInstances.apply((instances) =>
  instances.length ? elasticIps[0]?.publicIp : bastion.publicIp,
),
```

When using `Vpc.get()` (the ref path, lines 370-386), `registerOutputs()` is called at line 385, but the local `const natInstances = createNatInstances();` is only defined at line 401.

## The Fix

Line 693 should use `self.natInstances` instead of `natInstances`, since `this.natInstances` is set at line 380 in the ref path.

## Introduced In

This bug was introduced in **v3.17.26** via PR #5820 ("Export elastic ip for tunnel when present").

A reviewer flagged concerns about this exact issue in the PR.

## Reproduction Steps

1. Create a VPC with bastion enabled in a permanent stage:
   ```typescript
   export const vpc = new sst.aws.Vpc('MyVpc', { bastion: true })
   ```

2. Reference the VPC in a different stage using `Vpc.get()`:
   ```typescript
   export const vpc = sst.aws.Vpc.get('MyVpc', 'vpc-xxxxx')
   ```

3. Deploy the stage that uses `Vpc.get()`

4. Error occurs

## Environment

- SST Version: 3.17.34 (and all versions >= 3.17.26)
- Last working version: 3.17.25

brendan-smp:
## Summary

- Fixes `Vpc.get()` failing with `ReferenceError: Cannot access 'natInstances' before initialization` when referencing a VPC created with `bastion: true`

## Problem

When using `sst.aws.Vpc.get()` to reference an existing VPC that was created with `bastion: true`, deployment fails with:

```
ReferenceError: Cannot access 'natInstances' before initialization
```

## Root Cause

In `platform/src/components/aws/vpc.ts`, the `registerOutputs()` function references the local variable `natInstances` instead of the class property `self.natInstances`.

When using `Vpc.get()` (the ref path, lines 370-386), `registerOutputs()` is called at line 385, but the local `const natInstances = createNatInstances();` is only defined at line 401 (after the ref code path returns).

In the ref path, `this.natInstances` is properly set from `ref.natInstances`, but the `registerOutputs()` function incorrectly references the local variable instead of the class property.

## Fix

Change `natInstances` to `self.natInstances` in the `registerOutputs()` function (line 693).

Fixes #6346

This is a regression from #5820

### DIFF
--- a/platform/src/components/aws/vpc.ts
+++ b/platform/src/components/aws/vpc.ts
@@ -690,7 +690,7 @@ export class Vpc extends Component implements Link.Linkable {
           ([bastion, elasticIps, privateKeyValue, privateSubnets, publicSubnets]) => {
             if (!bastion) return;
             return {
-              ip: natInstances.apply((instances) =>
+              ip: self.natInstances.apply((instances) =>
                 instances.length ? elasticIps[0]?.publicIp : bastion.publicIp,
               ),
               username: "ec2-user",


jamesgibbons92:
I think I resolved this in #6341 already. But I will ping Frank to merge this as it's smaller and fixes the immediate issue.

PR #6358, opened 2026-01-26
fix(aws): add InvokeFunction permission for public Function URLs

### LINKED ISSUE #6198: Lambda Function URL Permissions not AWS-compliant
## Problem

I received a warning from AWS that starting November 1, 2026, all Lambda Function URLs require both permissions:
* `lambda:InvokeFunctionUrl`
* `lambda:InvokeFunction`

Currently, the second permission (lambda:InvokeFunction) is missing from the Next.js Lambda Functions created by SST.

## Affected Resources

When using sst.aws.Nextjs, 2 Lambda Functions with Function URLs are created:
1. Server Function: [...]FrontendServerEucentral1Function-*
2. Image Optimizer Function: [...]FrontendImageOptimizerFunction-*

**Both currently only have the `lambda:InvokeFunctionUrl` permission, but not the `lambda:InvokeFunction` one.**

## My setup so far

```
// sst.config.ts
export default $config({
app(input) {
  return {
    name: 'frontend',
    home: 'aws',
    providers: {
      aws: { region: 'eu-central-1' }
    }
  };
},
async run() {
  const site = new sst.aws.Nextjs('Frontend', {
    warm: 5,
    domain: { name: 'example.com' }
  });
  return { SiteUrl: site.url };
}
});
```

**Versions:**
- SST: 3.17.21
- OpenNext: 3.8.5
- Next.js: 15.3.3
- Node: 20.x

What I Tried

1. Updated dependencies: SST 3.17.5 → 3.17.21, OpenNext 3.6.5 → 3.8.5
2. Removed stack and deployed fresh

Neither works, there is still only one permission set and it still shows as Affected Resource :(

bruno-espino:
Fixes https://github.com/anomalyco/sst/issues/6198

  ## What this PR changes

  - When `sst.aws.Function` is configured with `url.authorization: "none"`, SST now adds an explicit resource-based permission:
    - `action: "lambda:InvokeFunction"`
    - `principal: "*"`
    - `function: <function name>`

  This matches the AWS console guidance for public Function URLs and fixes existing stacks without requiring a manual policy edit.

  ## Notes / follow-ups

  - The AWS console scopes the `InvokeFunction` statement with `lambda:InvokedViaFunctionUrl=true`.
  - This PR adds the minimal required permission via `aws.lambda.Permission` and includes an inline comment to remove it after we migrate to `@pulumi/aws` v7 (where Lambda/provider behavior matches the current
  console policy).

  ## Repro

  ```ts
  new sst.aws.Function("Api", {
    handler: "src/lambda.handler",
    url: { authorization: "none" },
  });

  Deploy and hit the function URL; before this change you'll get 403

### DIFF
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -2509,6 +2509,27 @@ export class Function extends Component implements Link.Linkable {
           },
           { parent },
         );
+
+        // Starting Oct 2025, public (auth NONE) Function URLs require an explicit
+        // lambda:InvokeFunction permission in addition to lambda:InvokeFunctionUrl.
+        //
+        // The lambda:InvokeFunctionUrl statement is added by Lambda or by the
+        // provider, but the lambda:InvokeFunction statement is not (for older
+        // deployments). This can be removed after migrating to @pulumi/aws v7
+        // (where Lambda/provider behavior matches the current AWS console policy).
+        //
+        // Note: The AWS console scopes this with lambda:InvokedViaFunctionUrl=true.
+        if (url.authorization === "none") {
+          new lambda.Permission(
+            `${name}UrlPublicInvokeFunction`,
+            {
+              action: "lambda:InvokeFunction",
+              function: fn.name,
+              principal: "*",
+            },
+            { parent, dependsOn: fnUrl },
+          );
+        }
         if (!url.route) return fnUrl.functionUrl;
 
         // add router route


jakehewitt:
Tested this PR and verified it works

platform/src/components/aws/function.ts
```diff
@@ -2509,6 +2509,27 @@ export class Function extends Component implements Link.Linkable {
...
           { parent },
         );
+
+        // Starting Oct 2025, public (auth NONE) Function URLs require an explicit
+        // lambda:InvokeFunction permission in addition to lambda:InvokeFunctionUrl.
+        //
+        // The lambda:InvokeFunctionUrl statement is added by Lambda or by the
+        // provider, but the lambda:InvokeFunction statement is not (for older
+        // deployments). This can be removed after migrating to @pulumi/aws v7
+        // (where Lambda/provider behavior matches the current AWS console policy).
```
jakehewitt, line 2519: I might not have tested this correctly but I don't think upgrading to @pulumi/aws v7 fixes this automatically. I tested by cloning https://github.com/anomalyco/sst/pull/6259 and upgrading from v7.12.0 → v7.16.0 (which includes terraform-provider-aws v6.28.0 where https://github.com/hashicorp/terraform-provider-aws/issues/44829 was resolved), but the issue persisted without these changes.

vimtor:
let's wait on merging #6259 first to see if the warning disappears

platform/src/components/aws/function.ts
bruno-espino: Interesting, the fix for v7 was an assumption from my side, I didn't test it.

bruno-espino: @vimtor if jake cloned from #6259 looks like this is still an issue

vimtor:
thanks for your contribution @bruno-espino 

closed in favor of #6400 which contains the fix for `protection: "oac"` mode too

> let's wait on merging https://github.com/anomalyco/sst/pull/6259 first to see if the warning disappears

i've changed my mind on this since i don't want new users to wait on the v7 to get merged for this common use-case to work

Issue #6304, opened 2026-01-02
feat: Expose ECR image to use in other services - like AWS Batch


MengLinMaker:
I would be great to abstract away some implementation jank to gain access to ECR images. This would provide a much better default developer experience.

For context, I'm deploying AWS Batch using SST which requires uploading images to ECR - refer to #6244

This is the current solution - requiring a janky workarounds to find the location and name of dockerfile
```TypeScript
import * as pulumi from '@pulumi/pulumi'
import path from 'node:path'

const Repository = new awsx.ecr.Repository('Repository')

// Need to get out of the '.sst/platform' folder to context folder 'src'
const contextPath = '../../src'

// Specify context and dockerfile name to avoid path issues with SST
const Image = new awsx.ecr.Image('Image', {
    repositoryUrl: Repository.url,
    dockerfile: path.join(contextPath, 'dockerfile'),
    context: path.join(contextPath),
})
```

I'm also using the `awsx` provider.

vimtor:
good idea @MengLinMaker 

we should probably expose the internal `imageBuilder` so you can do something like this:

```ts
const image = new sst.aws.Image('Image', {
  context: "./app"
});

const service = new sst.aws.Service('Service', {
  cluster,
  image,
});
```

MengLinMaker:
Yeah, your example is ideal.
___

Had a look at the internal `imageBuilder` at: https://github.com/anomalyco/sst/blob/4a76d47575f5a280c647194784ee362e7ee4ce67/platform/src/components/aws/helpers/container-builder.ts#L9

The builder is very bare bones, judging by how it's used:

https://github.com/anomalyco/sst/blob/4a76d47575f5a280c647194784ee362e7ee4ce67/platform/src/components/aws/fargate.ts#L1027

- Still need to work around the context path jank:
```Typescript
const contextPath = path.join($cli.paths.root, container.image.context);
```
- Still need to specify access to the ECR registry that `sst` uses internally:
```Typescript
ecr
  .getAuthorizationTokenOutput(
    {
      registryId: bootstrapData.assetEcrRegistryId,
    },
    { parent },
  )
  .apply((authToken) => ({
    address: authToken.proxyEndpoint,
    password: secret(authToken.password),
    username: authToken.userName,
  })),
```

---

Currently the components defined in `function.ts`, `service.ts`, `fargate.ts` all have separate implementations of image creation. Creating a flexible `Image` component that could plug straight into the other components could require a bit of refactoring

[...]

PR #6359, opened 2026-01-26
Add `Image` component for sharing Docker images

MengLinMaker:
Adding Image component for #6304:

- [x] Create `image.ts` component:
  - [x] Initial creation
- [x] Refactor `function.ts` and `fargate.ts` to use `image.ts`
  - [x] `function.ts`
  - [x] `fargate.ts`
- [x] Test changes against sst `examples`:
  - [x] `examples/aws-task`
  - [x] `examples/aws-python-container` - only python uses container for Lambda function
- [x] Add docs
  
Although efforts has been made to preserve functionality. Some behaviors may change:
- `fargate.ts` image name is changed. Can be preserved via transform.


MengLinMaker:
@vimtor the test in `platform` fails since `bun` lacks features from `node`, which are required for pulumi. I will use `npm` node for now. Though I believe `pnpm` will offer a better monorepo experience.

```Zsh
# Unhandled error between tests
-------------------------------
26 | // high-quality error messages.
27 | //
28 | // As a side-effect of importing this file, we must enable the --allow-natives-syntax V8 flag. This
29 | // is because we are using V8 intrinsics in order to implement this module.
30 | const v8 = __importStar(require("v8"));
31 | v8.setFlagsFromString("--allow-natives-syntax");
        ^
NotImplementedError: node:v8 setFlagsFromString is not yet implemented in Bun.
 code: "ERR_NOT_IMPLEMENTED"

      at /Users/menglinmaker/Documents/software-projects/open-source/sst/node_modules/@pulumi/pulumi/runtime/closure/v8.js:31:4
      at anonymous (unknown:1:1)
      at /Users/menglinmaker/Documents/software-projects/open-source/sst/node_modules/@pulumi/pulumi/runtime/closure/createClosure.js:29:7
      at anonymous (unknown:1:1)
      at /Users/menglinmaker/Documents/software-projects/open-source/sst/node_modules/@pulumi/pulumi/runtime/closure/serializeClosure.js:61:78
      at anonymous (unknown:1:1)
      at /Users/menglinmaker/Documents/software-projects/open-source/sst/node_modules/@pulumi/pulumi/runtime/index.js:29:1
```

vimtor:
hey @MengLinMaker no need to upgrade `vitest` or write tests for this feature

i'm working on a better testing setup. for the moment you can test the components against one of the local examples as described in the README

MengLinMaker:
AFAIK some `sst` examples are set to use `latest` version from npm. But we want to use the sst currently being changed...

Are you implying the local sst should be installed globally?

MengLinMaker:
Ah thanks its in the root README `#running-locally`

Then `sst diff` against an empty AWS account for snapshots

vimtor:
i usually do:

1. clone the repo
2. `bun install`
3. `go mod tidy`
4. `cd platform && bun run build`
5. `cd examples/<example> && bun install`
6. `go run ../../cmd/sst dev` (or deploy)

if you make changes you should run 4 and 6 again

MengLinMaker:
Not quite ready to merge, but I'd like confirm that current changes are desired.

Refactoring `fargate.ts` and `function.ts` doesn't add anything new, just cleaner abstractions.

Adding ability to use input image directly into `service.ts` should be straight forward.

platform/src/components/aws/image.ts
```diff
@@ -0,0 +1,250 @@
...
+   * SHA256 digest of the ECR image
+   */
+  public get digest() {
+    return this.image.digest;
+  }
+
+  /**
+   * The underlying [resources](/docs/components/#nodes) this component creates.
+   */
+  public get nodes() {
```
MengLinMaker: Not sure if this is needed

platform/src/components/aws/image.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  /**
+   * The uri of the ECR container image.
+   */
+  public get uri() {
+    return this.image.ref.apply((ref) => ref?.replace(":latest", ""));
+  }
+
+  /** @internal */
+  public getSSTLink() {
```
MengLinMaker: Not sure if this is needed

platform/src/components/aws/fargate.ts
```diff
@@ -1002,78 +999,18 @@ export function createTaskDefinition(
...
-            dockerIgnorePath,
-            [...lines, "", "# sst", ".sst"].join("\n"),
-          );
-        }
-
-        // Build image
-        const image = imageBuilder(
-          ...transform(
-            args.transform?.image,
-            `${name}Image${container.name}`,
```
MengLinMaker: The image name will be different with current abstraction

MengLinMaker:
Link issue #6304
Could also investigate #6369 after completion

vimtor:
@MengLinMaker this looks very promising

MengLinMaker:
## Progress update 2026-03-03

The main issue is that the image that lambda uses is not updating, as the api is not updating.

I'm experiencing flaky deployments after refactoring the `function.ts` component to use `image.ts`. Somehow image uri is not available despite explicitly instructing pulumi to wait until image output is available. But AWS console shows that the uri is correct and deployed.

The error:
```Bash
sdk-v2/provider2.go:515:
sdk.helper_schema: updating Lambda Function (aws-python-container-menglinmaker-PythonFnFunction-fxxadnar)
code: operation error Lambda: UpdateFunctionCode,
https response error StatusCode: 400, RequestID: 3a5010bd-2f70-4ef9-bb85-7cc19f62a16c,
InvalidParameterValueException: Source image 585946893245.dkr.ecr.ap-southeast-2.amazonaws.com/sst-asset:PythonFn@sha256:84d06db24db954438ec2befa6aa62bbc2e91ec405db5b4c8411f29421bf647b0 is not valid. Provide a valid source image.: provider=aws@6.66.2
```

### Current _"dodgy"_ fix
- remove digest from uri - so lambda essentially pulls latest container uploaded.
- this is a bad fix since you cannot rollback to older versions and lose track of which specific container build is causing images.

### Ideal solution
- force use of specific container version - previously a digest was used.

Also realised there are language specific logic just for python containers.

MengLinMaker:
Alright fix should be easy: `interpolate`${bootstrapData.assetEcrUrl}@${image.digest}`
This is the uri that should be used

MengLinMaker:
@vimtor 

I think the PR should be done now.

As for passing Image as a prop to Function to support containerised lambdas, there quite a few things to figure out:
1. How would dev work?
  1.1 `isContainer` is true when `args.dev` is true (enabled) and `$dev` (cli dev) is true
2. How would you retain breakpoints feature?
3. How to handle conflicting inputs?
4. Should we pass `Image` into `Function` or `uri`?
  4.1 Accommodate both?
```TypeScript
// image.ts
interface ImageInterface = { uri: string }
export class Image extends Component implements Link.Linkable, ImageInterface {
  ...
}

// function.ts
image: ImageInterface
```
There's a lot of delicate logic.

MengLinMaker:
I'm adding image compression options

MengLinMaker:
To be tested again

vimtor:
@MengLinMaker is it ready then? could you first fix the branch conflicts first so i don't miss anything when reviewing?

MengLinMaker:
@vimtor Yep ready to review

platform/src/components/aws/image.ts
```diff
@@ -0,0 +1,317 @@
...
+  Image as PulumiDockerImage,
+  ImageArgs as PulumiDockerImageArgs,
+} from "@pulumi/docker-build";
+import path from "path";
+
+// Extracted from `@pulumi/docker-build` `Platform` - unhandled by doc generator.
+export type Platform = "darwin/386" | "darwin/amd64" | "darwin/arm" | "darwin/arm64" | "dragonfly/amd64" | "freebsd/386" | "freebsd/amd64" | "freebsd/arm" | "linux/386" | "linux/amd64" | "linux/arm" | "linux/arm64" | "linux/mips64" | "linux/mips64le" | "linux/ppc64le" | "linux/riscv64" | "linux/s390x" | "netbsd/386" | "netbsd/amd64" | "netbsd/arm" | "openbsd/386" | "openbsd/amd64" | "openbsd/arm" | "plan9/386" | "plan9/amd64" | "solaris/amd64" | "windows/386" | "windows/amd64"
+
+// Extracted from `@pulumi/docker-build` `CompressionType` - unhandled by doc generator.
+export type CompressionType = "zstd" | "gzip" | "estargz"
```
MengLinMaker, line 19: `CompressionType` can be exported from `@pulumi/docker-build`. However, `CompressionType` is both an enum (string constant optionals) and a namespace, which the doc generator does not support.

Ideally, doc generator should support this, but would require quite a fair bit of logic to navigate the TypeScript AST.

platform/src/components/aws/image.ts
```diff
@@ -0,0 +1,317 @@
...
+
+  constructor(name: string, args: ImageArgs, opts?: any) {
+    const componentName = `${name}Image`
+    super(__pulumiType, componentName, args, opts);
+
+    const parent = this;
+    const region = getRegionOutput({}, opts).name;
+    const bootstrapData = region.apply((region) => bootstrap.forRegion(region));
+    // Empty uri should fail deployment if not set
+    this._uri = interpolate``
```
vimtor, line 197: defining `_uri` like this feels prone to race conditions. we should find an alternative

platform/src/components/aws/function.ts
```diff
@@ -2234,67 +2234,15 @@ export class Function extends Component implements Link.Linkable {
...
-          const authToken = ecr.getAuthorizationTokenOutput({
-            registryId: bootstrapData.assetEcrRegistryId,
+      return all([isContainer, dev, architecture]).apply(
+        ([isContainer, dev, architecture]) => {
+          if (!isContainer || dev) return;
+          return new Image(name, {
+            context: `.sst/artifacts/${name}-src`,
+            tags: ['latest'],
+            platforms: [architecture === "arm64" ? "linux/arm64" : "linux/amd64"],
           });
```
vimtor, line 2245: we should pass `{ parent }` like we do for other nested components

platform/src/components/aws/function.ts
```diff
@@ -2234,67 +2234,15 @@ export class Function extends Component implements Link.Linkable {
       // The build artifact directory already exists, with all the user code and
       // config files. It also has the dockerfile, we need to now just build and push to
       // the container registry.
-      return all([isContainer, dev, bundle, containerCache]).apply(
-        ([
-          isContainer,
-          dev,
-          bundle, // We need the bundle to be resolved because of implicit dockerfiles even though we don't use it here
-          containerCache,
```
vimtor, line 2344: we are no longer using this variable, perhaps we can remove it?

platform/src/components/aws/fargate.ts
```diff
@@ -1052,86 +1049,20 @@ export function createTaskDefinition(
...
+          dockerfile: container.image.dockerfile,
+          args: container.image.args,
+          secrets: linkEnvs,
+          target: container.image.target,
+          platforms: [container.image.platform],
+          tags: [container.name, ...(container.image.tags ?? [])],
+          transform: {
+            image: args.transform?.image,
+          },
+        });
```
vimtor, line 1063: same `{ parent }` thing

platform/src/components/aws/function.ts
vimtor, line 2245: same `{ parent }` thing

platform/src/components/aws/image.ts
```diff
@@ -0,0 +1,317 @@
...
+     * - `gzip` - best compatibility.
+     * - `estargz` - fastest pull.
+     * 
+     * @default `"zstd"`
+     */
+    type: Input<CompressionType>
+    /**
+     * Compression level. Different limits for each compression algorithm.
+     */
+    level: Input<number>
```
vimtor, line 113: i wonder how necessary `level` is, don't we always want the highest level?

www/generate.ts
```diff
@@ -789,7 +789,7 @@ function renderType(
 
     // @ts-expect-error
     delete type._project;
-    console.log(type);
+    console.log(JSON.stringify(type, null, 2));
```
vimtor: is this change necessary?

examples/aws-task/package.json
```diff
@@ -10,6 +10,6 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "sst": "^4"
+    "sst": "latest"
```
vimtor: can we bring this back as well?

vimtor (changes requested):
i'm going back to this again. sorry for the delay @MengLinMaker

can you update your changes with `origin/dev`? i believe we added secrets support in the deleted file

i've left some comments with issues i found and suggestions

my main concern is that, since we are replacing `imageBuilder` with a custom component, we will force a deletion + creation for existing images. not sure how much of a problem this is in practice, i need to investigate

platform/src/components/aws/image.ts
MengLinMaker: Agreed that highest compression isn't necessary, but having another avenue to reduce cold starts benefits some applications. Eg: on demand ML models.

Maybe we can specify which compression level ranges are recommended instead of a generic `number`?

Or remove and just use default compression level

MengLinMaker:
> my main concern is that, since we are replacing `imageBuilder` with a custom component, we will force a deletion + creation for existing images. not sure how much of a problem this is in practice, i need to investigate

For Fargate tasks, the new image will also create a new Task Definition.

Even currently SST appears to be creating new images and Task Definitions for Fargate.

Not sure about Function, which is only used by python

platform/src/components/aws/image.ts
MengLinMaker: @vimtor Sorry I don't quite understand the issue.

From what I understand, pulumi outputs needs to be "awaited" in Pulumi style: all(...).apply()

Are you saying that _uri may not be defined when get uri() is called?

Once the constructor is called, _uri should be defined before accessing

platform/src/components/aws/function.ts
MengLinMaker: Which variable? bundle is already removed

platform/src/components/aws/image.ts
```diff
@@ -0,0 +1,307 @@
...
+  constructor(name: string, args: ImageArgs, opts?: ComponentResourceOptions) {
+    const componentName = `${name}Image`
+    super(__pulumiType, componentName, args, opts);
+
+    const region = getRegionOutput({}, opts).name;
+    const bootstrapData = region.apply((region) => bootstrap.forRegion(region));
+    // Empty uri should fail deployment if not set
+    this._uri = interpolate``
+
+    all([args, bootstrapData]).apply(
```
MengLinMaker: Apparently creating resources inside all(...).apply() will not log the image uri

Reference: https://www.pulumi.com/docs/iac/concepts/inputs-outputs/apply/?utm_source=chatgpt.com#accessing-single-outputs-with-apply

platform/src/components/aws/image.ts
```diff
@@ -0,0 +1,307 @@
...
+ * ```
+ */
+export class Image extends Component implements Link.Linkable {
+  private _uri: Output<string>;
+
+  constructor(name: string, args: ImageArgs, opts?: ComponentResourceOptions) {
+    const componentName = `${name}Image`
+    super(__pulumiType, componentName, args, opts);
+
+    const region = getRegionOutput({}, opts).name;
```
MengLinMaker: Mark uri as something to be defined later in Pulumi

platform/src/components/aws/image.ts
MengLinMaker: This problem should be solved now.

MengLinMaker:
@vimtor, I've addressed the issues mentioned

straygar:
Great change, looking forward to its release! 🙏 (also for AWS Batch tidiness)

PR #6361, opened 2026-01-27
fix: Component resource sending input args to pulumi breaks some sst components

### LINKED ISSUE #6353: Bug: Component name is not unique with Linkable components
Hi,

I have the following manually wrapped `aws.sqs.Queue` (I know SST has its own queue but I want to have a bit more control here):

```typescript
sst.Linkable.wrap(aws.sqs.Queue, (resource) => ({
  properties: {
    arn: resource.arn,
    name: resource.name,
    url: resource.url,
  },
  include: [
    sst.aws.permission({
      actions: ['sqs:ReceiveMessage', 'sqs:SendMessage', 'sqs:DeleteMessage'],
      resources: [resource.arn],
    }),
  ],
}))

...

export const generalDlq = new aws.sqs.Queue('GeneralDeadLetterQueue', {
  name: `${appName}-general-dlq`,
  fifoQueue: false,
  messageRetentionSeconds: 1209600,
})
```

When deploying this with sst version `3.17.25` everything works fine. When deploying this with the latest version `3.17.38` I receive the following

```
|  Error       
Error: Component name GeneralDeadLetterQueue is not unique.
    at file:///project/.sst/platform/src/components/link.ts:65:17
    at new Resource (/project/.sst/platform/node_modules/@pulumi/resource.ts:469:26)
    at new CustomResource (/project/.sst/platform/node_modules/@pulumi/resource.ts:1292:9)
    at new Queue (/project/.sst/platform/node_modules/@pulumi/sqs/queue.ts:293:9)
    at Object.construct (/project/.sst/platform/node_modules/@pulumi/sqs/index.ts:46:24)
    at deserializeProperty (/project/.sst/platform/node_modules/@pulumi/runtime/rpc.ts:798:51)
    at Object.deserializeProperty (/project/.sst/platform/node_modules/@pulumi/runtime/rpc.ts:721:20)
    at resolveOutputs (/project/.sst/platform/node_modules/@pulumi/runtime/resource.ts:1185:33)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at /project/.sst/platform/node_modules/@pulumi/runtime/resource.ts:768:21
```

When I uncomment my linkable code, it works. I also have this with other linkable commponents.

Regards,

jamesgibbons92:
Closes #6353
 
Pulumi v3.202.0 introduced a change (https://github.com/pulumi/pulumi/pull/20357) that sends component resource inputs to the engine to be saved in state. 

This change causes two issues in SST that we know of:

Issue 1: "Component name is not unique" when using Linkable.wrap()
Issue 2: Step functions Out of memory errors

This is due to pulumi serialising/traversing the input args which has some side affects of causing issues in our transforms where a component is transformed twice, and step function components where there is some circular dependency issues.

Solution

<del>Added
`PULUMI_NODEJS_SKIP_COMPONENT_INPUTS=true` which was added in https://github.com/pulumi/pulumi/pull/20842
Which is a feature flag for disabling this behaviour for backwards compatibility.</del>

Updated SST Component class to not pass any args to pulumi ComponentResource. Effectlvely replicating the behaviour pre-2.202.0

### DIFF
--- a/pkg/project/run.go
+++ b/pkg/project/run.go
@@ -273,7 +273,6 @@ func (p *Project) RunNext(ctx context.Context, input *StackInput) error {
 		// "PULUMI_DISABLE_AUTOMATIC_PLUGIN_ACQUISITION=true",
 		"NODE_OPTIONS=--enable-source-maps --no-deprecation",
 		"PULUMI_HOME="+global.ConfigDir(),
-		"PULUMI_NODEJS_SKIP_COMPONENT_INPUTS=true",
 	)
 	if input.ServerPort != 0 {
 		env = append(env, "SST_SERVER=http://127.0.0.1:"+fmt.Sprint(input.ServerPort))
--- a/pkg/project/run.go
+++ b/pkg/project/run.go
@@ -273,6 +273,7 @@ func (p *Project) RunNext(ctx context.Context, input *StackInput) error {
 		// "PULUMI_DISABLE_AUTOMATIC_PLUGIN_ACQUISITION=true",
 		"NODE_OPTIONS=--enable-source-maps --no-deprecation",
 		"PULUMI_HOME="+global.ConfigDir(),
+		"PULUMI_NODEJS_SKIP_COMPONENT_INPUTS=true",
 	)
 	if input.ServerPort != 0 {
 		env = append(env, "SST_SERVER=http://127.0.0.1:"+fmt.Sprint(input.ServerPort))
--- a/platform/src/components/component.ts
+++ b/platform/src/components/component.ts
@@ -62,7 +62,7 @@ export class Component extends ComponentResource {
     for (const transform of transforms) {
       transform({ name, props: args, opts });
     }
-    super(type, name, args, {
+    super(type, name, {}, {
       transformations: [
         // Ensure logical and physical names are prefixed
         (args) => {


jamesgibbons92:
Using the following to reproduce linkable issue:
```
sst.Linkable.wrap(aws.sqs.Queue, (resource) => ({
      properties: {
        arn: resource.arn,
        name: resource.name,
        url: resource.url,
      },
      include: [
        sst.aws.permission({
          actions: [
            "sqs:ReceiveMessage",
            "sqs:SendMessage",
            "sqs:DeleteMessage",
          ],
          resources: [resource.arn],
        }),
      ],
    }));

    const generalDlq = new aws.sqs.Queue("GeneralDeadLetterQueue", {
      fifoQueue: false,
      messageRetentionSeconds: 1209600,
    });

    new sst.x.DevCommand("Command", {
      link: [generalDlq],
    });
```


Before the fix:

image


After the fix:

image

jamesgibbons92:
ugh.. this causes providers such as awsx to stop working as they rely on these args being sent. 
it seems the flag isn't truely backwards compatible and doesn't restore the previous behaviour completely.

jamesgibbons92:
Updated to pass {} to Pulumi's ComponentResource instead of the actual args. This should be safe because:
1. Before v3.202.0, these args were never sent to the engine anyway. Pulumi was setting the args to {} in the ComponentResource constructor explicitly already https://github.com/pulumi/pulumi/pull/20357/files
2. SST doesn't use Pulumi's initialize() or getData() methods that consume these args
3. The transformations set in the constructor returns early for SST Components on line 95 anyway, the transformations are run on the child resources.

I ran the same tests, both linkables and stepfunctions work now.

PR #6372, opened 2026-01-30
Bumping default version of OpenNext to 3.9.14

leanderiversen:
Bumping the default version of OpenNext to 3.9.14, after an issue reported was here: https://github.com/anomalyco/sst/pull/6334#issuecomment-3823532208

### DIFF
--- a/platform/src/components/aws/nextjs.ts
+++ b/platform/src/components/aws/nextjs.ts
@@ -11,7 +11,7 @@ import { isALteB } from "../../util/compare-semver.js";
 import { Plan, SsrSite, SsrSiteArgs } from "./ssr-site.js";
 import { Bucket } from "./bucket.js";
 
-const DEFAULT_OPEN_NEXT_VERSION = "3.9.8";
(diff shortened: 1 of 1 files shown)


vimtor (changes requested):
@leanderiversen the new version fails for next.js 14:

```
Next.js version : 14.2.15
OpenNext v3.9.14
ERROR Next.js version 14.2.15 is not supported by the Next.js team.
Major versions are supported for 2 years from their release date.
Unsupported releases may contain unpatched security vulnerabilities.
```

unsure about what's the best way to proceed here, we want to provide good defaults but i don't want to break existing deployments on a thing like this

so either we downgrade to the previous version before #6334 or add some message telling the user they need to set the `openNextVersion` to `"3.6.6"`

ideally we would detect this and choose in order:
1. the specified version
2. if next.js version 14 → `3.6.6`
3. if next.js version 15 → `3.9.14`

vimtor:
closing in favor of https://github.com/anomalyco/sst/pull/6406

PR #6379, opened 2026-02-02
Add WAF support to `Router`

### LINKED ISSUE #5846: Feature Request: Enable WAF configuration in the Router
It appears that there was some discussion on this in the pre-router days.

https://github.com/sst/sst/issues?q=is%3Aissue%20state%3Aclosed%20waf

And it appears that much of that discussion was closed in the abandoned (?) move to switch to the sst/v2 repo. 

In any case, Router is an incredible accomplishment and I have found it extremely useful. It is merely missing one very powerful feature of AWS which is the WAF.

Here is some related pulumi documentation that really outlines exactly how to do it

https://www.pulumi.com/answers/aws-waf-and-cloudfront-security-setup/

In the spirit of SST it would be amazing to have it have some reasonable default values with customization if desired, so perhaps something such as this.

```typescript
const router = new sst.aws.Router("MyRouter", {
  domain: {
    name: `...`,
    dns: sst.cloudflare.dns(),
  },
  waf: true
});
```

Or for the customizer
 
```typescript

// Define the AWS WAF Web ACL
const waf = new aws.wafv2.WebAcl("MyWaf", {
  ...
});
const router = new sst.aws.Router("MyRouter", {
  domain: {
    name: `...`,
    dns: sst.cloudflare.dns(),
  },
  transform: {
    webAclId: waf.id
  }
});
```

Related discord
https://discord.com/channels/983865673656705025/1377264052635570206

ekaya97:
## Summary

- Add optional WAF (Web Application Firewall) support to the Router component
- Rate limiting per IP (default: 2000 requests per 5-minute window)
- AWS Managed Rules: Core Rule Set, Known Bad Inputs, SQL Injection
- `SizeRestrictions_BODY` set to COUNT to avoid blocking large request bodies
- WAF automatically created in us-east-1 (required for CloudFront)
- Support for `transform.waf` to customize the WebACL

## Usage

```typescript
// Enable with defaults
new sst.aws.Router("MyRouter", {
  waf: true
});

// Or customize
new sst.aws.Router("MyRouter", {
  waf: {
    rateLimitPerIp: 1000,
    managedRules: {
      coreRuleSet: true,
      knownBadInputs: true,
      sqlInjection: false
    }
  }
});
```

## Test plan

- [x] Deployed Router with `waf: true` - WAF WebACL created successfully
- [x] Verified WAF is created in us-east-1
- [x] Verified CloudFront distribution references the WAF ARN
- [x] TypeScript compiles without errors

Fixes #5846

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/aws-router-waf/api.ts
+++ b/examples/aws-router-waf/api.ts
@@ -1,0 +1,9 @@
+export const handler = async (event: any) => {
+  return {
+    statusCode: 200,
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({ message: "Hello from WAF-protected Router!" }),
+  };
+};
--- a/examples/aws-router-waf/sst.config.ts
+++ b/examples/aws-router-waf/sst.config.ts
@@ -1,0 +1,44 @@
+/// <reference path="./.sst/platform/config.d.ts" />
+
+/**
+ * ## Router with WAF
+ *
+ * Enable WAF (Web Application Firewall) for a Router to protect against common
+ * web exploits and bots.
+ *
+ * WAF includes rate limiting per IP, and AWS managed rules for core rule set,
+ * known bad inputs, and SQL injection protection.
+ */
+export default $config({
+  app(input) {
+    return {
+      name: "aws-router-waf",
+      home: "aws",
+      removal: input?.stage === "production" ? "retain" : "remove",
+    };
+  },
+  async run() {
+    const api = new sst.aws.Function("MyApi", {
+      handler: "api.handler",
+      url: true,
+    });
+
+    const router = new sst.aws.Router("MyRouter", {
+      routes: {
+        "/*": api.url,
+      },
+      waf: {
+        rateLimitPerIp: 1000,
+        managedRules: {
+          coreRuleSet: true,
+          knownBadInputs: true,
+          sqlInjection: true,
+        },
+      },
+    });
+
+    return {
+      url: router.url,
+    };
+  },
+});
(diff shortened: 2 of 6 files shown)


vimtor (approved):
thanks for your contribution @ekaya97 

<details>

running this test script:

Image

Image

i've also tried deploying this without `waf` and it works as usual

</details>

could you please add an example in the `examples/` directory?

ekaya97:
@vimtor done

ekaya97:
PS: I did not add logging configuration because Router ships without logging by default.

It probably makes sense to enable CloudWatch and log anything but ALLOW. Let me know if you think this is necessary and whether it should opt-in or opt-out and the degree of configurability. 

Alternatively, the docs could mention that logging is recommended and provide example with pulumi native aws.cloudwatch.LogGroup + aws.wafv2.WebAclLoggingConfiguration

vimtor:
@ekaya97 yep logging should be really useful

i'm thinking of an api that looks like this:

```ts
logging: boolean | {
  include: "all" | "blocked",
  retention: 30,
  redact: {
    queryString: true,
    uriPath: true,
    method: true,
    headers: ["cookie"],
  },
},
```
it should probably be on by default with sane defaults

what do you think?

ekaya97:
should this apply to WAF only or the entire router? WAF creates a ton of logs, cost overhead shouldn't be underestimated. i would suggest off by default.

vimtor:
@ekaya97 it should only apply to WAF using https://www.pulumi.com/registry/packages/aws/api-docs/wafv2/webaclloggingconfiguration right? if it applied to the router it would be a separate feature

also you're right, we can keep it off by default and maybe even add a `::tip` callout mentioning the pricing as we do elsewhere

ekaya97:
force push because dev branch diverged since last commit.

vimtor (approved):
thanks for your contribution @ekaya97 

<details>

<summary>did some minor changes but everything is working</summary>

Image

</details>

vimtor:
released in [v4.1.0](https://github.com/anomalyco/sst/releases/tag/4.1.0)

PR #6394, opened 2026-02-06
Add `.npmrc` auth support

### LINKED ISSUE #6381: Unable to install pulumi in CI
Our CI environment requires using a local npm mirror for installing packages.  When we try to do `npx sst install` in CI we get this error because access to https://github.com is blocked:

```
06:25:24  sst-dev-common: ✕  Could not install pulumi
06:25:24  sst-dev-common:    Get "https://github.com/pulumi/pulumi/releases/download/v3.210.0/pulumi-v3.210.0-linux-x64.tar.gz": dial tcp 140.82.112.3:443: i/o timeout
```

The code that does tries to do that fetch is here:
https://github.com/anomalyco/sst/blob/1b060504abebfd6425083e24a661c186e8e1ffb9/pkg/global/pulumi.go#L67

Would it be possible to just `npm install` the package instead so that the `.npmrc` settings are respected or something similar?

vimtor:
closes #6381

### DIFF
--- a/pkg/npm/npm.go
+++ b/pkg/npm/npm.go
@@ -23,14 +23,14 @@ var usernameRegex = regexp.MustCompile(`^(//.+?):username\s*=\s*(.+)$`)
 var passwordRegex = regexp.MustCompile(`^(//.+?):_password\s*=\s*(.+)$`)
 var registryRegex = regexp.MustCompile(`^registry\s*=\s*(.+)$`)
 
-type NpmAuth struct {
+type Auth struct {
 	token  string
 	scheme string // "Bearer" or "Basic"
 }
 
 type npmrc struct {
 	registry string
-	auths    map[string]NpmAuth
+	auths    map[string]Auth
 }
 
 func expandEnvVars(s string) string {
@@ -40,7 +40,7 @@ func expandEnvVars(s string) string {
 }
 
 func parseNpmrc(content string) npmrc {
-	result := npmrc{auths: make(map[string]NpmAuth)}
+	result := npmrc{auths: make(map[string]Auth)}
 	usernames := make(map[string]string)
 	passwords := make(map[string]string)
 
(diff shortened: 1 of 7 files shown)


krmorse:
hey, i'm happy to try to test this whenever you think it's ready.  my approach was going to be to pull your branch and build the executable, and then i can just try to execute that executable in my ci environment with a simple example app to see if it's able to correctly work with the `.npmrc` file.

vimtor:
@krmorse feel free to try this pr, i believe it should work

pkg/npm/npm.go
```diff
@@ -1,15 +1,101 @@
...
 	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
 
 	"github.com/sst/sst/v3/internal/fs"
 )
 
+var envVarRegex = regexp.MustCompile(`\$\{([^}]+)\}`)
+var authTokenRegex = regexp.MustCompile(`^//([^/:]+).*:_authToken=(.+)$`)
```
krmorse:
```
var authTokenRegex = regexp.MustCompile(`^//([^/:]+).*:_auth\s*=\s*(.+)$`)
```

i had to update the regex a bit to work with my .npmrc in the following ways:
- add `\s*` surrounding the `=` to account for allowed whitespace
- change `_authToken` to be `_auth`, which is what npm expects i believe

pkg/npm/npm.go
```diff
@@ -21,12 +107,19 @@ type Package struct {
...
+	registry := loadNpmRegistry()
+	pkgURL := fmt.Sprintf("%s/%s/%s", registry.url, name, version)
+
+	req, err := http.NewRequest("GET", pkgURL, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	if registry.token != "" {
+		req.Header.Set("Authorization", "Bearer "+registry.token)
```
krmorse:
```
req.Header.Set("Authorization", "Basic "+registry.token)
```

i had to set this to `Basic`.  this is what npm does as well.

krmorse:
hey @vimtor, i took it for a quick spin this morning and was able to get it working with 2 proposed changes above.  let me know what you think? 

thanks again!

krmorse:
i just looked at the docs for npmrc: https://docs.npmjs.com/cli/v10/configuring-npm/npmrc

it looks like both what you implemented and what i did are supported, so maybe it needs to work with both?
- set `Bearer` header if it finds `_authToken`
- set `Basic` header if it finds `_auth`

vimtor:
thanks for the suggestions @krmorse 

can you try again with the latest changes?

krmorse:
just tested, and it worked for me on my machine.  i ran charles proxy to ensure it was actually fetching from my private registry and everything is looking good.  thanks again!

PR #6395, opened 2026-02-06
Add component for shared load balancers

### LINKED ISSUE #5510: ApiGatewayV2 for Private Resources not working
Hi,

This integration should to work out of the box but somehow not working as expected.

My expected flow:
Api gateway > defaultRoute Integration to LB in same vpc > ECS

```js
 api.routePrivate("$default", service.nodes.loadBalancer.arn, {
      transform: {
        integration: {
          description: `Load balancer integration for ${name}`,
          requestParameters: {
            "append:header.extendedrequestid": "$context.extendedRequestId" //https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-parameter-mapping.html
          }
        }
      }
    });
```

Error
```js
Failed    
   ApiRouteBbovca sst:aws:ApiGatewayV2PrivateRoute → ApiRouteBbovcaIntegration aws:apigatewayv2:Integration
   sdk-v2/provider2.go:515: sdk.helper_schema: creating API Gateway v2 Integration: operation error ApiGatewayV2: CreateIntegration, https response error StatusCode: 400, RequestID: d4639c52-1a73-491d-bad7-204af3dcfe47, BadRequestException: For VpcLink VPC_LINK, integration uri should be a valid ELB listener ARN or a valid Cloud Map service ARN.: provider=aws@6.66.
```

Previous code (Working with CloudMap)
Sometimes getting 503 error on high load, Want to check the same with LB setup.

```js
 api.routePrivate("$default", service.nodes.cloudmapService.arn, {
      transform: {
        integration: {
          description: `SRV record for ${name}`,
          requestParameters: {
            "append:header.extendedrequestid": "$context.extendedRequestId" //https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-parameter-mapping.html
          }
        }
      }
    });

### LINKED ISSUE #6017: Flexible load balancer for Service
I want to create one load balancer in one service and then import this in another service and assign there service, port, conditions etc...

So basically, One load balancer to serve all ecs spread over different repo.

abc.com/a1 - Service1>Container1
abc.com/a2 - Service1>Container2
abc.com/a3 - Service2>Container1
abc.com/a4 - Service3>Container1

Routing can be done via path or header
So only one endpoint will be exposed to serve traffic from all micro-service.

**Tried**
I have tried serviceRegistry for the same but api gateway is throwing random 503 error (INTEGRATION_NETWORK_FAILURE) on high load. 

Reference
https://github.com/sst/sst/issues/4228
https://github.com/sst/sst/issues/5510

### PR DESCRIPTION
**Pull request by ekaya97**
## Summary

Add a standalone `Alb` component that can be shared across multiple `Service` instances. This allows multiple ECS services to share a single Application Load Balancer with path-based routing.

Fixes #6017
Fixes #5510

### New API

```ts
// Create a shared ALB
const alb = new sst.aws.Alb("SharedAlb", {
  vpc,
  listeners: [{ port: 80, protocol: "http" }],
});

// Attach services with routing rules
new sst.aws.Service("Api", {
  cluster,
  loadBalancer: {
    instance: alb,
    rules: [
      {
        listener: "80/http",
        forward: "3000/http",
        conditions: { path: "/api/*" },
        priority: 100,
      },
    ],
  },
});
```

### Features

- **Alb component** — standalone ALB with HTTP/HTTPS listeners, SSL/TLS, and DNS integration
- **Service integration** — `loadBalancer.instance` option for attaching to external ALBs
- **Routing rules** — path, header, and query string condition-based routing with explicit priorities
- **Reference pattern** — `Alb.get(name, arn)` with lazy listener discovery for cross-stack references
- **Linkable** — implements `Link.Linkable`, exposes `url` for SST linking

### Files Changed

| File | Changes |
|------|---------|
| `platform/src/components/aws/alb.ts` | New 705-line ALB component |
| `platform/src/components/aws/service.ts` | Union-typed `loadBalancer` prop, ALB attachment logic |
| `platform/src/components/aws/index.ts` | Export `Alb` |
| `platform/package.json` | Added `@types/node` devDependency |
| `platform/vitest.config.ts` | Vitest config (fixes `@types/node` resolution) |
| `platform/test/components/alb.test.ts` | 10 unit tests |
| `platform/test/components/service-alb.test.ts` | 9 integration tests |
| `examples/aws-shared-alb/` | HTTP example |
| `examples/aws-shared-alb-domain/` | HTTPS with custom domain example |

## Test plan

- [x] Unit tests pass (35 tests total)
- [x] Deployed `aws-shared-alb` example to AWS — 7/7 tests pass
- [x] Deployed `aws-shared-alb-domain` example to AWS — 9/9 tests pass (SSL, HTTP→HTTPS redirect)

### HTTP Example Test Results
Screenshot 2026-02-06 at 16 24 38

### HTTPS Example Test Results  
Screenshot 2026-02-06 at 16 23 17

---

🤖 Generated with [Claude Code](https://claude.ai/code)


nicholasgubbins:
this is super nice @ekaya97

excited to be able to do router.route(`${subdomain}.{domain}`, sharedAlb.url)

ekaya97:
@nicholasgubbins 

Thanks! 

Check out my other PR #6379 - similar kind of extension concerning WAF in Router.

teodorrd27:
This is so valuable I've monkey patched my existing sst deployment to deploy in production. It'll be great to see this merged. Thanks for your effort. 🙏

platform/src/components/aws/alb.ts
```diff
@@ -0,0 +1,705 @@
...
+                  certificateArn: protocol === "HTTPS"
+                    ? certificateArn.apply((arn) => arn!) as Output<string>
+                    : undefined,
+                  defaultActions,
+                },
+                { parent: self },
+              ),
+            );
+
+            result[key] = listener;
```
vimtor: this pattern scares me a bit, i fear that if in some setups `.apply` is applied asynchronously the `result` will be empty. we should probably refactor this to return an `Output` and maybe use the `all([])` utility for resolving all the args

platform/src/components/aws/alb.ts
```diff
@@ -0,0 +1,705 @@
...
+ * });
+ * ```
+ *
+ * #### Reference an existing ALB
+ *
+ * ```ts title="sst.config.ts"
+ * const alb = sst.aws.Alb.get("SharedAlb", "arn:aws:elasticloadbalancing:...");
+ * ```
+ */
+export class Alb extends Component implements Link.Linkable {
```
vimtor, line 218: although the `Alb` name is technically correct i feel like we should name this `LoadBalancer` just to be consistent with the rest of the code, specially considering that the `Nlb` use-case is weaker

no need to change this atm but i wonder what are your thoughts

platform/src/components/aws/service.ts
```diff
@@ -2391,6 +2580,235 @@ export class Service extends Component implements Link.Linkable {
...
+      const raw = args.loadBalancer as any;
+      if (raw && typeof raw === "object" && "instance" in raw) {
+        const instance = raw.instance;
+        if (
+          instance instanceof Alb ||
+          (instance &&
+            typeof instance === "object" &&
+            "_loadBalancer" in instance &&
+            "_listeners" in instance &&
+            "getListener" in instance)
```
vimtor: why do we need all of these checks or say that `args.loadBalancer as any`? shouldn't `instance instanceof Alb` or `'instance' in args.loadBalancer` be enough?

platform/src/components/aws/service.ts
```diff
@@ -221,6 +222,89 @@ interface ServiceRules {
...
+     */
+    header?: Input<{
+      name: Input<string>;
+      values: Input<Input<string>>[];
+    }>;
+  }>;
+  /**
+   * Explicit priority for the listener rule (1–50000).
+   * Must be unique per listener across ALL services sharing the ALB.
+   * Use non-overlapping ranges per service (e.g., Service A: 100-199, Service B: 200-299).
```
vimtor, line 297: i wonder if we could enfore this somehow, like throwing a `VisibleError` is we detect an overlapping range

perhaps doable if the loadbalancer has some kind of registerListener method to keep track

not required, just a thought

platform/src/components/aws/service.ts
```diff
@@ -1679,30 +1818,44 @@ export class Service extends Component implements Link.Linkable {
...
-      : all([self.domain, self.loadBalancer?.dnsName]).apply(
-          ([domain, loadBalancer]) =>
-            domain ? `https://${domain}/` : `http://${loadBalancer}`,
-        );
+    this.domain = albAttachment
+      ? output(undefined)
+      : lbArgs?.domain
+        ? lbArgs.domain.apply((domain) => domain?.name)
+        : output(undefined);
+    this._url = albAttachment
```
vimtor: wondering if the url should include `conditions.path` so when you link it you have the proper url to call

it's probably ideal but too cumbersome to parse the path for wildcards and weird formats

platform/vitest.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  resolve: {
+    alias: {
+      // @types/node has empty "exports": {} which Vite can't resolve.
+      // It's type-only, so alias to the .d.ts file at runtime.
+      "@types/node": "@types/node/index.d.ts",
```
vimtor: thanks for this! i'll add a vitest step to the github action after we merge

platform/src/components/aws/service.ts
```diff
@@ -2339,7 +2528,7 @@ export class Service extends Component implements Link.Linkable {
                 predefinedMetricSpecification: {
                   predefinedMetricType: "ALBRequestCountPerTarget",
                   resourceLabel: all([
-                    loadBalancer?.arn,
+                    loadBalancer?.arn ?? albAttachment?.instance.arn,
```
vimtor: i believe the proper pulumi pattern here is to include both in the `all()` and check inside the `apply` since they are both `Output` this can lead to unexpected results

platform/src/components/aws/alb.ts
```diff
@@ -0,0 +1,705 @@
...
+    );
+
+    this.registerOutputs({ _hint: this._url });
+
+    function reference() {
+      const ref = args as unknown as AlbRef;
+      const loadBalancer = lb.LoadBalancer.get(
+        `${name}LoadBalancer`,
+        output(ref.loadBalancerArn).apply((arn) => {
+          // Extract the ID from the ARN for the Pulumi get call
```
vimtor: not sure i understand this comment given we are not extracting the ID from the ARN and just returning it

just in case, there is an utillity function for extracting the ID from an ARN somwhere

platform/src/components/aws/alb.ts
```diff
@@ -0,0 +1,705 @@
...
+   * ```js
+   * {
+   *   defaultAction: "fixed-403"
+   * }
+   * ```
+   */
+  defaultAction?: Input<
+    | "fixed-404"
+    | "fixed-403"
+    | { redirect: { port: Input<number>; protocol: Input<"http" | "https"> } }
```
vimtor: exposing the whole api here doesn't feel wrong to me.

what about something like:

```
response?: { statusCode: Input<number>; body?: Input<string> };
redirect?: { protocol: Input<"http" | "https">; host?: Input<string>; port: Input<number>; };
```

i don't want to open a new pr to add `"fixed-500"`,`"fixed-503"` etc.

platform/src/components/aws/alb.ts
```diff
@@ -0,0 +1,705 @@
...
+            internal: pub.apply((v) => !v),
+            loadBalancerType: "application",
+            subnets,
+            securityGroups: [securityGroup.id],
+            enableCrossZoneLoadBalancing: true,
+            tags: {
+              "sst:ref:version": Alb._refVersion.toString(),
+              "sst:ref:sg": securityGroup.id,
+              "sst:ref:vpc-id": vpcId,
+              ...(domain.apply((d) =>
```
vimtor: this reads to me like a pulumi bug too since we will be spreading `Output` objects the actual object itself

platform/src/components/aws/service.ts
```diff
@@ -2230,18 +2396,41 @@ export class Service extends Component implements Link.Linkable {
...
-                  ),
+                loadBalancers: albAttachment
+                  ? targetGroups &&
+                    all([albAttachment.rules, targetGroups, containers]).apply(
+                      ([rules, targets, ctrs]) =>
+                        Object.entries(targets).map(([targetId, target]) => {
+                          // targetId format: containerNamePROTOCOLport
+                          // We need to extract containerName from the rules
+                          const matchingRule = rules.find((r) => {
+                            const resolved = r as any;
```
vimtor: we don't need this casting here, i tried locally and the type for `r` is properly resolved

platform/src/components/aws/service.ts
```diff
@@ -221,6 +222,89 @@ interface ServiceRules {
...
+   * The listener to attach to, in `{port}/{protocol}` format. Must match a listener on the ALB.
+   *
+   * @example
+   * ```js
+   * {
+   *   listener: "443/https"
+   * }
+   * ```
+   */
+  listener: Input<AlbPort>;
```
vimtor: for the regular `ServiceRules` this is called `listen` we should probably change this to be consistent

platform/src/components/aws/alb.ts
```diff
@@ -0,0 +1,705 @@
...
+      const securityGroup = ec2.SecurityGroup.get(
+        `${name}SecurityGroup`,
+        tags.sgId,
+        {},
+        { parent: self },
+      );
+
+      // Discover listeners via the load balancer's defaultActions
+      // Note: Listeners are discovered at deploy time via the LB ARN.
+      // For the reference pattern, users will need to access listeners by protocol:port key.
```
vimtor: i might be wrong but i believe this comment might be outdated given the `getListener` does a `lb.Listener.get`

vimtor (changes requested):
hey @ekaya97, thanks for this incredible pull request

i tested your examples and they work like a charm

since this is a complex feature, i took the time to review the code for potential bugs and api improvements

i left some comments about it, nothing major

we should be able to merge this pull request soon

vimtor (commented):
also i forgot to mention if you considered moving this logic to the `Cluster` component itself, otherwise we should probably validate that the cluster vpc is the same as the alb one

curious to hear what you think

platform/src/components/aws/alb.ts
ekaya97: Agree, apply would return empty. Will fix.

platform/src/components/aws/alb.ts
ekaya97: Agree in principle but prefer the descriptive pattern.

platform/src/components/aws/alb.ts
ekaya97: Changing the shape would require cascading updates, will add the additional 500 and 503 shorthands.

platform/src/components/aws/alb.ts
ekaya97: Comment is a leftover from development, will remove.

platform/src/components/aws/alb.ts
ekaya97: Good catch, will change to:

```typescript
tags: all([domain]).apply(([d]) => ({
    "sst:ref:version": Alb._refVersion.toString(),
    "sst:ref:sg": securityGroup.id,
    "sst:ref:vpc-id": vpcId,
    ...(d ? { "sst:ref:domain": d.name } : {}),
  })),
```

platform/src/components/aws/service.ts
ekaya97: Unnecessarily defensive, I agree. will simplify

platform/src/components/aws/service.ts
ekaya97: I thought about this but found implementing a shared state per stage to complicate unnecessarily. Instead we are catching collisions early during deployment. Fail-Fast pattern

platform/src/components/aws/service.ts
ekaya97: Exposing multiple rules with different paths is way too cumbersome. Base URL suffices (same idea in principal as Router.route )

platform/src/components/aws/service.ts
ekaya97: Valid, same issue as the tag. Will fix.

platform/src/components/aws/service.ts
ekaya97: Will fix.

ekaya97:
> also i forgot to mention if you considered moving this logic to the `Cluster` component itself, otherwise we should probably validate that the cluster vpc is the same as the alb one
> 
> curious to hear what you think

Agree on VPC confirmation. Moving to Cluster is an anti-pattern - ALB can be used by Router.route as well.

platform/src/components/aws/service.ts
ekaya97: Will fix

platform/src/components/aws/alb.ts
ekaya97: Leftover comment, will remove

ekaya97:
Sorry for the multiple force push - regenerated bun lock. CI passes.

## Changes made:

#### Pulumi Output/async correctness                                                                                                                              
  - Refactored createListeners to unwrap Input<Input<AlbListenerArgs>[]> via all() instead of populating a dict inside nested .apply() callbacks
  - Tags now use all([domain]).apply() instead of spreading an Output object directly
  - resourceLabel in auto-scaling now resolves both LB ARN sources inside all().apply() instead of using ?? on Output wrappers

 #### API consistency
  - Renamed ServiceAlbRule.listener → listen to match ServiceRules.listen

 #### Type cleanup
  - Removed unnecessary as any casts in loadBalancers mapping
  - Simplified detectAlbAttachment to use instanceof Alb only — removed duck-type fallback

 #### Stale comments
  - Removed misleading "Extract the ID from the ARN" comment and simplified to pass ARN directly
  - Removed outdated listener discovery comment

 #### New
  - Added "fixed-500" and "fixed-503" default action shorthands (refactored buildDefaultActions to use a lookup table)
  - Added VPC validation — throws VisibleError if ALB and cluster VPCs don't match
  - Added test coverage for VPC validation

 #### Not addressed (intentionally)
  - Alb → LoadBalancer rename: prefer keeping Alb as it's descriptive and accurate
  - Cross-service priority overlap enforcement: would require shared state per stage, current fail-fast at deploy time is sufficient
  - URL with conditions.path: multiple rules with different paths makes this impractical, base URL is consistent with Router.route

ekaya97:
@vimtor - any updates on this?

vimtor:
@ekaya97 will try to review it next week!

vimtor:
@ekaya97 i've made the following changes:
-  refactored the service logic a bit since it was hard to follow
-  removed `defaultAction` from the `Alb` args for now
-  removed the domain example
-  made `Alb` appear in the docs as an internal component

for me this is ready to merge, can you verify the changes on your end?

ekaya97:
@vimtor - looks good to me 👍🏻

ekaya97:
docs for router component are not referencing alb for Router.route 
- for your consideration @vimtor

vimtor:
@ekaya97 what do you mean? not sure i understand

ekaya97:
Router.route can target Alb.url as @nicholasgubbins pointed out.

vimtor:
@ekaya97 for the moment, i'm going to keep the documentation as it is, since i believe this is a pretty niche use case and the router component is complex enough. maybe we can add it in the future if this new component gets used more

PR #6415, opened 2026-02-17
Add instructions on how to debug Golang Lambdas

baartch:
It is possible to debug Golang functions with VSCode.
All we have to do is figure out the `processId` and attach our debugger to it.

I have added instructions to the _Live_ documentation on how to do this.


vimtor:
@baartch shouldn't we switch go dev builds to `-gcflags=all=-N -l` for this to work reliably?

baartch:
@vimtor Please excuse me. I’m not yet familiar enough with Go to judge that. Looking at what the flags do, it does seem to make sense. For me, it’s worked perfectly fine without them so far.

PR #6438, opened 2026-02-20
Fix binary request body corruption in React and Remix Lambda handlers

### LINKED ISSUE #6275: Binary request bodies corrupted in sst.aws.React Lambda handler due to UTF-8 string conversion
When receiving webhook requests with binary bodies (e.g., gzip-compressed payloads), the `sst.aws.React` Lambda handler corrupts the body by converting it to a UTF-8 string. Any bytes that aren't valid UTF-8 are replaced with the Unicode replacement character (`U+FFFD`), making it impossible to handle webhooks from services that send compressed or binary payloads.

## Root Cause

The bug is in the `convertLambdaRequestToNode` function in `platform/functions/react-server/server.mjs`:

```javascript
body:
  event.body && event.isBase64Encoded
    ? isFormData
      ? Buffer.from(event.body, "base64")
      : Buffer.from(event.body, "base64").toString()  // ← BUG: corrupts binary data
    : event.body,
```

When `isBase64Encoded` is `true` and the content is not form data, the handler:
1. ✅ Correctly base64-decodes the body with `Buffer.from(event.body, "base64")`
2. ❌ Then calls `.toString()` which defaults to UTF-8 encoding, corrupting any non-UTF-8 bytes

### Example corruption

Gzip data starts with magic bytes `0x1f 0x8b`. The byte `0x8b` is not valid UTF-8 (it's a continuation byte without a lead byte), so it gets replaced with `U+FFFD` (`0xef 0xbf 0xbd`):

- **Sent by webhook:** `1f 8b 08 00 ...` (valid gzip)
- **Received by app:** `1f ef bf bd 08 00 ...` (corrupted)

## Why This Appears Unintentional

1. **The `isFormData` exception proves awareness of binary handling** — form data is kept as a Buffer because file uploads would break otherwise. This logic just wasn't extended to other binary content types.

2. **The Web `Request` API accepts binary bodies** — `ArrayBuffer`, `Uint8Array`, and `Blob` are all valid body types. There's no technical requirement to stringify.

3. **Silent data corruption is never desirable** — if this were intentional, it would be documented or throw an error for unsupported content types.

## Steps to Reproduce

1. Deploy a React Router app using `sst.aws.React`
2. Create an API route that receives webhook POST requests
3. Have an external service send a gzip-compressed JSON payload with `Content-Encoding: gzip`
4. Attempt to read/decompress the body in your route handler

```typescript
// app/routes/api/webhook.ts
import { gunzipSync } from "node:zlib";

export async function action({ request }: { request: Request }) {
  const raw = Buffer.from(await request.arrayBuffer());
  
  console.log({
    contentEncoding: request.headers.get("content-encoding"),
    firstBytesHex: raw.slice(0, 8).toString("hex"),
    // Expected: "1f8b0800..." (valid gzip magic bytes)
    // Actual:   "1fefbfbd..." (corrupted - 0x8b replaced with replacement char)
  });
  
  // This throws: "incorrect header check"
  const decompressed = gunzipSync(raw);
  return new Response("ok");
}
```

## Expected Behavior

Binary request bodies should be passed through unchanged, allowing handlers to decompress or process them correctly.

## Actual Behavior

Binary bytes that aren't valid UTF-8 are replaced with `U+FFFD` (`ef bf bd`), corrupting the payload. This causes:
- `gunzipSync` to fail with "incorrect header check"
- JSON parsing to fail with "Unexpected token"
- Any binary processing to receive garbage data

## Proposed Fix

Remove the `.toString()` call and keep binary data as-is. The `Request` constructor handles `Buffer` just fine:

```javascript
body:
  event.body && event.isBase64Encoded
    ? Buffer.from(event.body, "base64")
    : event.body,
```

Or, for a more conservative fix that preserves string behavior for text content:

```javascript
const decodedBody = event.body && event.isBase64Encoded
  ? Buffer.from(event.body, "base64")
  : event.body;

const contentEncoding = event.headers["content-encoding"];
const contentType = event.headers["content-type"] || "";
const isBinary = contentEncoding || 
  isFormData || 
  contentType.includes("octet-stream") ||
  contentType.includes("protobuf");

// Only stringify if we're confident it's text
body: decodedBody instanceof Buffer && !isBinary
  ? decodedBody.toString("utf-8")
  : decodedBody,
```

## Environment

- SST version: 3.17.25
- Node.js: 22.18.0
- Component: `sst.aws.React`

vimtor:
closes #6275

<details>

### before

CleanShot 2026-02-20 at 14 16 54@2x

### after

CleanShot 2026-02-20 at 14 20 47@2x

</details>

### DIFF
--- a/platform/functions/remix-server/regional-server.mjs
+++ b/platform/functions/remix-server/regional-server.mjs
@@ -3,72 +3,6 @@
 
 import { createRequestHandler as createNodeRequestHandler } from "@remix-run/node";
 
-/**
- * Common binary MIME types
- */
-const binaryTypes = [
-  "application/octet-stream",
-  // Docs
-  "application/epub+zip",
-  "application/msword",
-  "application/pdf",
-  "application/rtf",
-  "application/vnd.amazon.ebook",
-  "application/vnd.ms-excel",
-  "application/vnd.ms-powerpoint",
-  "application/vnd.openxmlformats-officedocument.presentationml.presentation",
-  "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
-  "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
-  // Fonts
-  "font/otf",
-  "font/woff",
-  "font/woff2",
-  // Images
-  "image/bmp",
-  "image/gif",
-  "image/jpeg",
-  "image/png",
-  "image/tiff",
-  "image/vnd.microsoft.icon",
-  "image/webp",
-  // Audio
-  "audio/3gpp",
-  "audio/aac",
-  "audio/basic",
-  "audio/mpeg",
-  "audio/ogg",
-  "audio/wavaudio/webm",
-  "audio/x-aiff",
-  "audio/x-midi",
-  "audio/x-wav",
-  // Video
-  "video/3gpp",
-  "video/mp2t",
-  "video/mpeg",
-  "video/ogg",
-  "video/quicktime",
-  "video/webm",
-  "video/x-msvideo",
-  // Archives
-  "application/java-archive",
-  "application/vnd.apple.installer+xml",
-  "application/x-7z-compressed",
-  "application/x-apple-diskimage",
-  "application/x-bzip",
-  "application/x-bzip2",
-  "application/x-gzip",
-  "application/x-java-archive",
-  "application/x-rar-compressed",
-  "application/x-tar",
-  "application/x-zip",
-  "application/zip",
-];
-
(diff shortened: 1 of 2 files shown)


platform/functions/remix-server/regional-server.mjs
```diff
@@ -3,83 +3,13 @@
 
 import { createRequestHandler as createNodeRequestHandler } from "@remix-run/node";
 
-/**
```
vimtor, line 6: these were not being used

PR #6444, opened 2026-02-23
Remove Opencontrol from JS SDK

### LINKED ISSUE #6190: Version of hono included via downstream dependency opencontrol has multiple vulnerabilities
Note: This duplicates https://github.com/sst/opencontrol/issues/48, posting here as it impacts all recent sst versions? Delete whichever one isn't needed.

The version of `hono` included as a dependency in `opencontrol` has two open vulnerabilities, one medium severity and one high:

- Hono has Body Limit Middleware Bypass - https://github.com/advisories/GHSA-92vj-g62v-jqhh
- Hono Improper Authorization vulnerability - https://github.com/advisories/GHSA-m732-5p4w-x69g

Because SST also includes opencontrol, it seems like this impacts all recent versions of SST.

```
sst@3.17.19
  └─┬ opencontrol@0.0.6
    └── hono@4.7.4
```

### LINKED ISSUE #6424: Migration to AWS SDK v3
> also see #3385

SST still uses AWS SDK v2, which has [reached EOL in September 2025](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-javascript-v2/).

It seems like the SDK is only used in [`opencontrol.ts`](https://github.com/anomalyco/sst/blob/dev/sdk/js/src/opencontrol.ts). I can help with the migration, but I don't really understand which parts of the SDK get invoked here. My suggestion would be to add the SDK v3 packages that can get invoked here (like S3 and EC2) and create a wrapper around them, so that the tool invocation can stay more or less the same.

> **Background**:
> This came up during license checking (see also #6423) - the old AWS SDK depends on `querystring@0.2.0`, which has no license and is marked as deprecated.

jens-ox:
closes #6424
closes #6190

This removes the `opencontrol` experiment and with that the very chunky and deprecated `aws-sdk` dependency (and with that transitively from OpenCode etc).


vimtor:
thanks for your contribution @jens-ox

jens-ox:
Thanks @vimtor! I've opened an Opencode PR to update sst: https://github.com/anomalyco/opencode/pull/15345

PR #6448, opened 2026-02-23
Use actual stage name in ApiGatewayV1 URL

### LINKED ISSUE #6368: ApiGatewayV1: Incorrect url returned when using a non-default stage
I think this code here is wrong:
https://github.com/anomalyco/sst/blob/dev/platform/src/components/aws/apigatewayv1.ts#L831

```
interpolate`https://${this.api.id}.execute-api.${this.region}.amazonaws.com/${$app.stage}/`;
```

specifically the `${app.stage}` part ☝️.

This works fine in the default case, but if I have an api with a different stage name it doesn't work right (even though the api does correctly create the specified stage)
```
const api = new sst.aws.ApiGatewayV1("MyApi", {
    transform: {
        stage: {
            stageName: "v1",
        },
    },
});
api.deploy();

//expected: https://someid.execute-api.us-west-2.amazonaws.com/v1/
//actual: https://someid.execute-api.us-west-2.amazonaws.com/appStage/
const url = api.url; 
```

Can someone confirm this is an issue?  
I think the fix would be pretty easy:
```
interpolate`https://${this.api.id}.execute-api.${this.region}.amazonaws.com/${this.stage?.stageName ?? $app.stage}/`;
```

vimtor:
closes #6368

### DIFF
--- a/platform/src/components/aws/apigatewayv1.ts
+++ b/platform/src/components/aws/apigatewayv1.ts
@@ -828,7 +828,7 @@ export class ApiGatewayV1 extends Component implements Link.Linkable {
         ([domain, key]) =>
           key ? `https://${domain}/${key}/` : `https://${domain}`,
       )
-      : interpolate`https://${this.api.id}.execute-api.${this.region}.amazonaws.com/${$app.stage}/`;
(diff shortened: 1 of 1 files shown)


krmorse:
just was able to validate this, works great, thanks again!

PR #6465, opened 2026-02-24
Reuse bridge code for dev lambdas

MaTTaDox:
In dev mode, every aws.Function uploaded the same bridge zip as its own S3 object. This change deduplicates that upload so functions in the same region reuse one shared Live bridge asset.

  ## Problem

  Each Lambda in dev used identical bridge code (bootstrap from platform/dist/bridge) but still created/uploaded a per-function zip object.
  This caused redundant uploads and long startup time of `sst dev` in bigger projects

## Changes
  - Added an in-memory cache for Live bridge assets keyed by region:bundle.
  - In dev mode (non-container functions), create a deterministic bridge zip, hash it, and upload to:
      - assets/live-bridge-code-{hash}.zip
  - Reuse the same BucketObjectv2 promise for all matching functions.
  - Create the shared S3 object at stack root (pulumi.rootStackResource) with opts?.provider.
  - Use stable logical name: LiveBridgeCode{region}.
 
## Validation 
I currently use the version with this change locally and it works way facter than before.
  
##  Breaking Changes
With this release the s3 key of every lambda in dev mode gets replaced once and old objects getting deleted

@vimtor would be awesome if you can have a look at this. I am also open to alternative suggestions for this problem.
Thank you!


vimtor:
very cool @MaTTaDox

let me take a look once we merge the aws v7 upgrade

platform/src/components/aws/function.ts
```diff
@@ -1713,7 +1718,7 @@ export class Function extends Component implements Link.Linkable {
         : true,
     );
     const partition = getPartitionOutput({}, opts).partition;
-    const region = getRegionOutput({}, opts).region;
+    const region = getRegionOutput({}, opts).name;
```
dsonet: Shouldn't it be `.region`? I see all other places are using `.region` now

MaTTaDox: Youre right, thanks!

platform/src/components/aws/function.ts
```diff
@@ -2819,7 +2897,7 @@ export class Function extends Component implements Link.Linkable {
       {
         functionName: this.name,
         environment,
-        region: getRegionOutput(undefined, { parent: this }).region,
+        region: getRegionOutput(undefined, { parent: this }).name,
```
dsonet: I suppose this should be reverted too.

vimtor: yep, it should be `.region` @MaTTaDox

vimtor:
thanks for your contribution @MaTTaDox 

haven't looked into this too much but from what i can remember i only seen logs like this when creating a new lambda function:

```
|  Created     Example sst:aws:Function → ExampleLogGroup aws:cloudwatch:LogGroup (1.1s)
|  Created     Example sst:aws:Function → ExampleRole aws:iam:Role (1.8s)
|  Created     Example sst:aws:Function → ExampleCode aws:s3:BucketObjectv2 (23.9s)
|  Created     Example sst:aws:Function → ExampleFunction aws:lambda:Function (7.4s)
|  Created     Example sst:aws:Function → ExampleInvokeFunction aws:lambda:Permission
|  Created     Example sst:aws:Function → ExampleUrl aws:lambda:FunctionUrl
|  Created     Example sst:aws:Function (36.7s)
```

in what cases does this pull request help then? when creating new functions? on live reload changes? initial startup times?

MaTTaDox:
> thanks for your contribution @MaTTaDox
> 
> haven't looked into this too much but from what i can remember i only seen logs like this when creating a new lambda function:
> 
> ```
> |  Created     Example sst:aws:Function → ExampleLogGroup aws:cloudwatch:LogGroup (1.1s)
> |  Created     Example sst:aws:Function → ExampleRole aws:iam:Role (1.8s)
> |  Created     Example sst:aws:Function → ExampleCode aws:s3:BucketObjectv2 (23.9s)
> |  Created     Example sst:aws:Function → ExampleFunction aws:lambda:Function (7.4s)
> |  Created     Example sst:aws:Function → ExampleInvokeFunction aws:lambda:Permission
> |  Created     Example sst:aws:Function → ExampleUrl aws:lambda:FunctionUrl
> |  Created     Example sst:aws:Function (36.7s)
> ```
> 
> in what cases does this pull request help then? when creating new functions? on live reload changes? initial startup times?

The bridge code for Lambdas is exactly the same for every Lambda when you run `sst dev`. Right now, this bridge code is uploaded separately for every Lambda in the project, which can take a large amount of time depending on internet speed.

With my change, the bridge code is uploaded only once.

This issue is made worse by the fact that the hash of the bridge binary seems to change regulary, even though the code itself has not changed in over a year. I’m currently looking into that but thats something for a different PR/Issue

Anyway, to me it makes sense to upload it only once if the bridge binary is the same.

vimtor:
@MaTTaDox understood. even if it only happens on lambda creation it still takes too much time, right?

the hash bug you mentioned was spotted here:
https://github.com/anomalyco/sst/pull/6061#discussion_r2878992973

haven't tested it but you can include it in your pr if it improves things

MaTTaDox:
@vimtor depending on project size yes. And its unnecessary, if its the same code anyway.

vimtor:
@MaTTaDox makes sense

is this ready to merge then? the `.region` needs to be fixed

also, maybe you want to take a look at the "utf" thing?

MaTTaDox:
@vimtor yes this is ready to merge, fixed the `.region`

MaTTaDox:
> also, maybe you want to take a look at the "utf" thing?

yes will do that when i research the hash problem

vimtor:
@MaTTaDox i've made some minor changes to avoid code duplication and also fixed the function hash bug

will merge this today

thank you!

PR #6478, opened 2026-02-25
Set $dev during refresh on dev stage

jamesgibbons92:
This fix adds `--dev` arg to the refresh command, for refreshing dev stages.


jamesgibbons92:
Tested by using the following in `sst.config.ts`
```
if ($dev) {
      console.log("DEV MODE");
    } else {
      console.log("NOT DEV MODE");
    }
```

`sst dev`
image

Before fix refresh behaviour (I had to read the pulumi.log file to verify the logs, as sst refresh does not always output them):

`sst refresh` 
image

After fix refresh behaviour:
`sst refresh` (without the --dev arg)
image


`sst refresh --dev`` (with the --dev arg)
image

PR #6482, opened 2026-02-25
Fix refresh when using custom provider config

jamesgibbons92:
When using the `providers: {}` configuration block to configure providers, this is not being sent to pulumi during `refresh` command. Which means the refresh fails for these providers.

### Reproduce:

```
export default $config({
  app(input) {
    return {
      name: "neon",
      home: "aws",
      providers: {
        neon: {
          apiKey:
            "xxxxxxxxxxxxxxxxxxxxxx", // replace with your neon api key
        },
      },
    };
  },
  async run() {
    new neon.Branch("NeonDBBranch", {
      projectId: "my-project-id",
      name: "dev",
    });
  },
});
```

Refresh doesn't work:
image

### Fix:
This fix bumps the pulumi sdk to a version where the --config arg [was added for the refresh command](https://github.com/pulumi/pulumi/pull/21299), and then append this arg to the command when sst calls it during a refresh.

Refresh works:
image


pkg/project/provider/aws.go
```diff
@@ -58,9 +58,6 @@ func (a *AwsProvider) Env() (map[string]string, error) {
 	if a.profile != "" {
 		env["AWS_PROFILE"] = a.profile
 	}
-	if a.config.Region != "" {
```
jamesgibbons92, line 61: reverts [this commit](https://github.com/anomalyco/sst/pull/6259/changes/a3fa6e4ccb5146cd6ebb892178c7769b5f2489ab) as it shouldn't be needed now we're passing the config to pulumi

PR #5969, opened 2025-07-14
feat(aws): Add Dsql component and examples

chrisfowles:
This pull request introduces functionality to create DSQL Clusters, both single and multi-region.

It also adds two new examples for AWS Aurora DSQL clusters: one for single-region clusters and another for multi-region clusters. 

### Platform Updates to Support DSQL
* `platform/src/components/aws/index.ts`: Added export for the new `dsql.js` component to integrate DSQL functionality into the AWS platform.
* `platform/src/components/component.ts`: Updated `Component` class to include support for the new `aws:dsql/cluster:Cluster` type.
* `platform/package.json`: Upgraded `@pulumi/aws` dependency to version `6.83.0` for compatibility with DSQL features.
* `platform/src/components/aws/dsql.ts`: The new component itself.

#### Multi-Region Example:
* [`examples/aws-dsql-multiregion/README.md`](diffhunk://#diff-ae46b92990214f8f53803014208088dcb0d41a74c0628069a633ec599e1c65f9R1-R97): Added documentation for creating and connecting to a multi-region Aurora DSQL cluster, including deployment instructions and code snippets for connecting to both primary and peer clusters.
* [`examples/aws-dsql-multiregion/package.json`](diffhunk://#diff-e05df7eadbe87373db2746c4931d4cb4335e540ad1c7f477a1ca172ff06ecba3R1-R20): Defined dependencies and scripts for the multi-region example, including `@aws-sdk/dsql-signer` and `pg` for database connectivity.
* [`examples/aws-dsql-multiregion/src/lambda.ts`](diffhunk://#diff-27d04563fcc6eb4c6f58556fc6c15a1b2f643d92b660444bf6bd4457ba281b68R1-R59): Implemented Lambda function to connect to both primary and peer clusters using IAM authentication and PostgreSQL drivers.
* [`examples/aws-dsql-multiregion/sst.config.ts`](diffhunk://#diff-bf22a238f538f7184ad70ed995a65587d1ca83ad97b19c237c0dc698cecee006R1-R30): Configured SST to create a multi-region DSQL cluster and link it to a Lambda function.

#### Single-Region Example:
* [`examples/aws-dsql/README.md`](diffhunk://#diff-ac32e561654e498d104ce6804a0145dcaf40e65a8f635fde40ed3395da62ee32R1-R101): Added documentation for creating and connecting to a single-region Aurora DSQL cluster, with deployment instructions and code snippets for IAM authentication and PostgreSQL connection.
[...]

PR #6496, opened 2026-02-27
Add `Dsql` AWS component

DamienPace15:
Built off the back of PR #5969, with testing and additional code suggested by @natac13. Also merged in latest dev with the latest 4.x version.

This adds the `sst.aws.Dsql` component for Amazon Aurora DSQL.

### Single-region

Creates an Aurora DSQL cluster in the current region. Linked functions get `endpoint` and `region` via `Resource`.

### Multi-region

Creates a primary cluster in the current region and a peer cluster in another region. An AWS provider for the peer region is created automatically so you don't need to configure one yourself. The two-way `ClusterPeering` handshake that DSQL requires is also handled automatically. Linked functions get `endpoint` and `region` for the primary cluster and `peer.endpoint` and `peer.region` for the peer. The peer cluster is accessed via its public endpoint which is secured by short-lived IAM auth tokens and TLS.

### VPC endpoints

You can optionally create AWS PrivateLink interface endpoints inside a VPC. Two types are supported: `management` for control plane operations and `connection` for PostgreSQL client connections. Both default to `false` since they cost money and need to be explicitly enabled. A security group is created automatically that allows inbound TCP/5432 from within the VPC CIDR. When a connection endpoint is created, `Resource.MyCluster.endpoint` automatically resolves to the private hostname instead of the public one, so your function code doesn't need to change.

> **Note:** Multi-region VPC endpoint support is not included yet. It has some nuances around cross-region VPC creation and routing that are better handled separately.

### `Dsql.get`

Reference an existing cluster by its identifier, useful for sharing a cluster across stages. Supports both single-region and multi-region clusters via `peerIdentifier` and `peerRegion`.

## Tested

- [x] Single region deployment
- [x] Multi region deployment
- [x] Single region connection within Lambda
- [x] Multi region connection within Lambda connecting to both clusters
- [x] Remove single region cluster
- [x] Remove multi region cluster


platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,352 @@
...
+      properties: {
+        arn: this.arn,
+        identifier: this.identifier,
+        publicEndpoint: this.publicEndpoint,
+        vpcEndpointServiceName: this.vpcEndpointServiceName,
+        region: this.region,
+      },
+      include: [
+        permission({
+          actions: ["dsql:DbConnect", "dsql:DbConnectAdmin", "dsql:GetCluster"],
```
DamienPace15, line 558: This is an interesting one to me. 

This actually allows the resource that's linked with the cluster to have super user credentials meaning that they can do anything. 

I was thinking potentially having a link where you would have to explicitly allow admin connection instead of it being enabled by default.

I am still learning about DSQL so I will have to think about this a bit more. Happy for anyone to chime in on this.

pkg/project/provider/aws.go
```diff
@@ -58,6 +58,9 @@ func (a *AwsProvider) Env() (map[string]string, error) {
 	if a.profile != "" {
 		env["AWS_PROFILE"] = a.profile
 	}
+	if a.config.Region != "" {
+		env["AWS_REGION"] = a.config.Region
+	}
```
vimtor: can you update your branch with the latest `origin/dev`?  

it'll remove the need for this

examples/aws-dsql-multiregion/README.md
```diff
@@ -0,0 +1,103 @@
+# AWS Aurora DSQL Multi-Region Example
```
vimtor: examples are generated from the jsdocs

this should be above the `$config` in `sst.config.ts` like the other examples

examples/aws-dsql/README.md
```diff
@@ -0,0 +1,101 @@
+# AWS Aurora DSQL Example
```
vimtor: examples are generated from the jsdocs

this should be above the `$config` in `sst.config.ts` like the other examples

www/src/content/docs/docs/examples.mdx
```diff
@@ -4139,7 +4139,7 @@ import type { SQSEvent, SQSHandler } from "aws-lambda";
 export const handler: SQSHandler = async (event: SQSEvent) => {
   for (const record of event.Records){
     // Message bodies are always strings
-    console.log(record.body) 
+    console.log(record.body)
```
vimtor: can you undo these changes? the `examples.mdx` file is auto generated

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,352 @@
...
+    });
+  }
+
+  private static regionFromArn(clusterArn: Output<string>) {
+    return clusterArn.apply((arn) => {
+      const parts = arn.split(":");
+      const region = parts[3];
+      return region;
+    });
+  }
```
vimtor: there are several ARN utilities in the codebase. we should check if any of the existing ones cover this use case

if this is sql specific, the existing code is fine

examples/aws-dsql-multiregion/lambda.ts
```diff
@@ -0,0 +1,59 @@
+import { DsqlSigner } from '@aws-sdk/dsql-signer';
```
vimtor: minor style thing: can you lift these lambdas out of the `src/` directory, as the other examples?

platform/src/components/aws/providers/dsql-peering.ts
```diff
@@ -0,0 +1,297 @@
...
+        },
+      };
+    } catch (error: any) {
+      throw new Error(`Failed to create DSQL peering: ${error.message}`);
+    }
+  }
+
+  private async establishPeering(inputs: DsqlPeeringInputs): Promise<void> {
+    // Import AWS SDK inside the method to avoid serialization issues
+    const { DSQLClient, UpdateClusterCommand, GetClusterCommand } =
```
vimtor:
```suggestion
    const { DSQLClient, UpdateClusterCommand } =
```

i believe we are not using the `GetClusterCommand` here

platform/src/components/aws/providers/dsql-peering.ts
```diff
@@ -0,0 +1,297 @@
...
+          throw new Error(
+            `Timeout waiting for cluster ${identifier} to establish peering: ${error.message}`,
+          );
+        }
+        // Continue waiting on transient errors
+        await new Promise((resolve) => setTimeout(resolve, delayMs));
+      }
+    }
+
+    throw new Error(
```
vimtor: errors thrown in components need to be of type `VisibleError` to show up correctly

examples/aws-dsql/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "scripts": {
+    "dev": "sst dev",
+    "deploy": "sst deploy",
+    "remove": "sst remove"
+  },
+  "dependencies": {
+    "@aws-sdk/client-dsql": "^3.700.0",
+    "@aws-sdk/dsql-signer": "^3.844.0",
+    "pg": "^8.16.3",
+    "sst": "^4.1.1"
```
vimtor: should be:

```suggestion
    "sst": "^4"
```

like the other examples

examples/aws-dsql-multiregion/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "scripts": {
+    "dev": "sst dev",
+    "deploy": "sst deploy",
+    "remove": "sst remove"
+  },
+  "dependencies": {
+    "@aws-sdk/client-dsql": "^3.700.0",
+    "@aws-sdk/dsql-signer": "^3.844.0",
+    "pg": "^8.16.3",
+    "sst": "^4.1.1"
```
vimtor: should be:

```suggestion
    "sst": "^4"
```

like the other examples

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,352 @@
...
+   * ```js
+   * {
+   *   tags: {
+   *     Environment: "production",
+   *     Team: "backend"
+   *   }
+   * }
+   * ```
+   */
+  tags?: Input<Record<string, Input<string>>>;
```
vimtor: very few componets expose this. i say we either delete the prop or don't clutter the examples with it

bun.lock
vimtor, line 1: just to confirm. did you generate this file by running the recommended bun command?
https://bun.com/docs/pm/lockfile#text-based-lockfile

examples/aws-dsql-multiregion/README.md
```diff
@@ -0,0 +1,103 @@
...
+);
+
+// Connect to the peer cluster
+const peerTime = await connectToCluster(
+  Resource.MyCluster.peerRegion,
+  Resource.MyCluster.peerPublicEndpoint
+);
+```
+
+## Cost considerations
```
vimtor: both example docs feel way too verbose with too many instructions notes and so on

we should check other example and move necessary things like costs to the main docs

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,352 @@
...
+   *
+   * @default `false`
+   * @example
+   * ```js
+   * {
+   *   deletionProtection: true
+   * }
+   * ```
+   */
+  deletionProtection?: Input<boolean>;
```
vimtor: i believe we don't need to expose the `deletionProtection` prop

we should instead add the resources to the `platform/src/auto/run.ts` list so they follow the `protection` prop in `sst.config.ts`

people can always set this via `transform`

this follows the existing pattern for the `Postgres` or `Aurora` components

examples/aws-dsql/src/lambda.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const handler = async (event: any) => {
+  try {
+    // Use the Resource object to get the region and endpoint
+    const signer = new DsqlSigner({
+      region: Resource.MyCluster.region,
+      hostname: Resource.MyCluster.publicEndpoint,
+    });
+
+    // Generate the token
```
vimtor: let's remove these unnecessary comments from the examples

examples/aws-dsql-multiregion/README.md
```diff
@@ -0,0 +1,103 @@
...
+## Key features demonstrated
+
+### Multi-Region Cluster Definition
+
+In `sst.config.ts`, a multi-region cluster is defined by providing the `multiRegion` property.
+
+```typescript
+// sst.config.ts
+const cluster = new sst.aws.Dsql('MyCluster', {
+  multiRegion: {
```
vimtor: this property doesn't exist in the component

platform/src/components/aws/dsql-peering.ts
```diff
@@ -0,0 +1,227 @@
+import { Region } from "@pulumi/aws";
+import { ComponentResourceOptions, Input, Output, all } from "@pulumi/pulumi";
+import { Component } from "../component";
+import type { Dsql } from "./dsql";
+import { DsqlPeeringProvider } from "./providers/dsql-peering";
+
+export interface DsqlPeeringArgs {
```
vimtor: i believe we no longer need the custom provider since pulumi has a resource for this: https://www.pulumi.com/registry/packages/aws-v6/api-docs/dsql/clusterpeering/

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,352 @@
...
+   *
+   * @default `"AWS_OWNED_KMS_KEY"`
+   * @example
+   * ```js
+   * {
+   *   kmsEncryptionKey: "arn:aws:kms:us-east-1:123456789012:key/12345678-1234-1234-1234-123456789012"
+   * }
+   * ```
+   */
+  kmsEncryptionKey?: Input<string>;
```
vimtor: i believe most aws components use the default encryption so i wouldn't expose this as well

users can always use `transform` if they need this level of detail

vimtor (changes requested):
thanks for your contribution @DamienPace15

i've left a bunch of comments and suggestions

thanks again for taking a look at this

bun.lock
DamienPace15: I just used `bun install` in the root. 

Just to confirm I should run `bun install --save-text-lockfile --frozen-lockfile --lockfile-only` and keep the `bun.lockb` file?

platform/src/components/aws/dsql.ts
DamienPace15: I have removed this and added `aws:dsql/cluster:Cluster` to the `addTransformationToRetainResourcesOnDelete()` function in `platform/src/auto/run.ts`.

Have not added the `aws:dsql/clusterPeering:ClusterPeering` relationship because it didn't really fit.

platform/src/components/aws/dsql.ts
DamienPace15: I can't find any generic region from arn like utility that works with Output<string> from what I can see. The existing utilities are all service specific parsers for example `parseQueueArn`, `parseLambdaEdgeArn` and have some specific logic around that.

platform/src/components/aws/dsql-peering.ts
DamienPace15: Should I be removing the dsql peering component completely? 

Using the native pulumi peering component means that I would have to do something like this. 

` 
   const witnessRegion = 'us-west-2';

    const primary = new sst.aws.Dsql('Primary', {
      witnessRegion: witnessRegion,
    });

    const peerProvider = new aws.Provider('PeerRegion', {
      region: 'us-east-2',
    });

    const peer = new sst.aws.Dsql(
      'Peer',
      { witnessRegion: witnessRegion },
      { provider: peerProvider }
    );

    // Create peering between clusters
    new aws.dsql.ClusterPeering('PrimaryPeering', {
      identifier: primary.identifier,
      clusters: [peer.arn],
      witnessRegion,
    });

    new aws.dsql.ClusterPeering(
      'PeerPeering',
      {
        identifier: peer.identifier,
        clusters: [primary.arn],
        witnessRegion,
      },
      { provider: peerProvider }
    );`

Or should we be keeping the peering component so some of this is abstracted?

platform/src/components/aws/providers/dsql-peering.ts
DamienPace15: Will address once comment around the DSQL peering component is addressed if necessary. Just need clarification if we are keeping this component or not.

platform/src/components/aws/dsql.ts
vimtor: probably it still makes sense to extract service-specific arn functions so we can test them. if you update the branch with origin-dev, you'll see new tests for those functions

you can keep them working with regular strings and just do the apply on the component side.

platform/src/components/aws/dsql-peering.ts
vimtor: i believe we can remove the custom peering component and either provide this as an example or abstract this logic inside the existing DSQL component

an example of the latter would be:

```ts
const primary = new sst.aws.Dsql('Primary', {
  witness: { region: 'us-east-1' },
  peers: [
    { region: 'us-east-2' },
    { region: 'eu-west-1' },
  ],
});
```

i'm not super familiar with dsql but this latter approach feels very clean

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,413 @@
...
+   * ```ts
+   * const cluster = new sst.aws.Dsql("MyCluster", {
+   *   multiRegion: {
+   *     witnessRegion: "us-west-2",
+   *     peer: { region: "us-east-2" }
+   *   }
+   * });
+   * ```
+   */
+  multiRegion?: {
```
DamienPace15: @vimtor I have abstracted away the creation of peering. 

I have moved `witnessRegion` inside of a `multiRegion` object. `witnessRegion` was originally one level up but is useless for the single cluster DSQL component. 

AWS currently only supports 2 regions for the multi region setup so the thought process behind it is the following. 

On the creation of the DSQL component it will create the primary cluster no matter what and only create the peering if the multiRegion field exists.

When the component is declared it looks like so. 

single region

``` typescript 
const cluster = new sst.aws.Dsql('MyCluster', {});
```

multi region
``` typescript
const cluster = new sst.aws.Dsql('MultiRegion', {
      multiRegion: { witnessRegion, peer: { region: 'us-east-2' } },
    });
```

There needs to be 3 unique regions. 
witness region
primary region 
secondary region

Let me know what you think. 

Have tested deploying and removing single and multi clusters on normal deployments and live lambda deployments.

platform/package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "@aws-sdk/client-cloudfront": "3.478.0",
     "@aws-sdk/client-cloudwatch-logs": "3.478.0",
+    "@aws-sdk/client-dsql": "^3.844.0",
```
vimtor: i believe we no longer need this since the removal of the dynamic provider

bun.lock
vimtor: don't worry i will fix these before merging the pr.

if you want to fix it deleting it and running bun install should work (in fact if you remove the unnecessary dep no bun changes should appear)

platform/src/auto/run.ts
```diff
@@ -29,6 +29,7 @@ function addTransformationToRetainResourcesOnDelete() {
         [
           "aws:dynamodb/table:Table",
           "aws:rds/instance:Instance",
+          "aws:dsql/cluster:Cluster",
```
vimtor, line 32: i wonder if we should also add the peering stuff here? can you confirm?

examples/aws-dsql/sst.config.ts
```diff
@@ -0,0 +1,49 @@
...
+
+2. **Test the connection**
+   The Lambda function when invoked will connect to the DSQL cluster and run a test query.
+
+3. **View the results**
+   Check the Lambda function logs to see the successful connection and query results.
+
+## Clean up
+
+```bash
```
vimtor: can we restructure these docs to follow the pattern of the others?

if you don't want to bother i will do it before merging

examples/aws-dsql-multiregion/sst.config.ts
```diff
@@ -0,0 +1,71 @@
+/// <reference path="./.sst/platform/config.d.ts" />
+
+/* # AWS Aurora DSQL Multi-Region Example
```
vimtor: same thing here

platform/src/components/aws/dsql.ts
vimtor: minor style suggestion:

since sst usually prefers single names for variable i would rename the args to look:

```ts
const cluster = new sst.aws.Dsql("MultiRegion", {
  regions: { witness: "us-west-1", peer: "us-east-2" },
});
```

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,413 @@
...
+ *   region: Resource.MyCluster.peerRegion,
+ *   hostname: Resource.MyCluster.peerPublicEndpoint,
+ * });
+ * ```
+ */
+
+export class Dsql extends Component implements Link.Linkable {
+  private cluster: Output<dsql.Cluster>;
+  private peerCluster: Output<dsql.Cluster> | undefined;
+  private peerProvider: aws.Provider | undefined;
```
vimtor: does this need to be a property stored? i belive we are not referencing it elsewhere

platform/src/components/aws/dsql.ts
vimtor: bumping this

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,413 @@
...
+          ...transform(
+            args.transform?.peerCluster,
+            `${name}PeerCluster`,
+            {
+              multiRegionProperties: {
+                witnessRegion: multiRegion!.witnessRegion,
+              },
+              tags: { Name: name },
+            },
+            { parent, provider: peerProvider },
```
vimtor: for the provider stuff we can use the existing `useProvider`  utility

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,413 @@
...
+              tags: { Name: name },
+            },
+            { parent },
+          ),
+        ),
+      );
+    }
+
+    function createPeerCluster() {
+      return output(
```
vimtor: i believe we don't need to wrap this with an output

examples/aws-dsql-multiregion/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "aws-dsql-multiregion-example",
```
vimtor: drop `-example` from the package.json and sst.config.ts

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,413 @@
...
+
+export interface DsqlArgs {
+  /**
+   * Configure multi-region cluster peering.
+   *
+   * When specified, creates a multi-region DSQL setup with a primary cluster in the
+   * current region and a single peer cluster in another region, linked via a shared
+   * witness region. An AWS provider for the peer region is created automatically.
+   *
+   * Aurora DSQL supports exactly two regions in a multi-region cluster (primary + one peer).
```
vimtor: i believe you can't transition from single cluster to multi cluster

we shold probably add a warning callout for it

examples/aws-dsql/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "name": "aws-dsql-example",
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "sst dev",
+    "deploy": "sst deploy",
+    "remove": "sst remove"
+  },
+  "dependencies": {
+    "@aws-sdk/client-dsql": "^3.700.0",
```
vimtor: i believe we are not using this dep in the examples?

examples/aws-dsql-multiregion/sst.config.ts
vimtor, line 1: i believe this example is missing a `.gitignore`

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,413 @@
...
+
+    this.peerProvider = peerProvider;
+
+    const primaryCluster = createPrimaryCluster();
+    this.cluster = primaryCluster;
+
+    const createdPeerCluster = multiRegion ? createPeerCluster() : undefined;
+    this.peerCluster = createdPeerCluster;
+
+    // ── Peering resources ────────────────────────────────────────────────────
```
vimtor: we can remove these comments. the code is self-explanatory

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,413 @@
...
+   * Here `app-dev-mycluster` is the identifier of the DSQL Cluster created in the `dev` stage.
+   * You can find this by outputting the cluster identifier in the `dev` stage.
+   *
+   * ```ts title="sst.config.ts"
+   * return {
+   *   cluster: cluster.identifier
+   * };
+   * ```
+   */
+  public static get(
```
vimtor, line 511: just to confirm: have you tested the `get` path to verify it works? just by looking at the code my intuition tells me the peering thing won't work

this is usually solved by using tags to find the other resources (see postgres for an example)

platform/src/components/aws/dsql.ts
vimtor: i've done some research and for the sst model it'd be very cumbersome because each lambda gets its own role and we'd need to run a postgres statement to grant permissions to the role

i believe it's fine since we only grant permissions when we link the resource. also in practice doesn't feel different than the current postgres defaults

probably worth documenting somewhere

examples/aws-dsql-multiregion/sst.config.ts
```diff
@@ -0,0 +1,71 @@
...
+    */
+    const witnessRegion = 'us-west-2';
+
+    //creates primary and peer cluster
+    const cluster = new sst.aws.Dsql('MultiRegion', {
+      multiRegion: { witnessRegion, peer: { region: 'us-east-2' } },
+    });
+
+    const fn = new sst.aws.Function('MyFunction', {
+      handler: 'lambda.handler',
```
vimtor: regarding vpc support: 

can you try putting one of the lambdas in a vpc and check if it works with the current setup?

if it does we can add vpc support later, otherwise we might need to add it

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,413 @@
...
+      properties: {
+        primaryRegion: this.region,
+        primaryPublicEndpoint: this.publicEndpoint,
+        peerRegion: this.peerCluster
+          ? Dsql.regionFromArn(this.peerCluster.arn)
+          : undefined,
+        peerPublicEndpoint: this.peerCluster
+          ? Dsql.publicEndpointFromArn(this.peerCluster.arn)
+          : undefined,
+      },
```
vimtor, line 555: naming suggestion:


```suggestion
      properties: {
        region: this.region,
        endpoint: this.publicEndpoint,
        peer: { region, endpoint },
      },
```

rationale:
- `endpoint` is shorter (sst prefers short names) and matches https://docs.aws.amazon.com/aurora-dsql/latest/APIReference/API_GetCluster.html#auroradsql-GetCluster-response-endpoint 
- top level is for primary (one prefix less and makes more sense for non multi-region setups)
- group `peer` so both properties can be checked at the same time (i.e. if `peer` exists both props are available)

vimtor (changes requested):

examples/aws-dsql-multiregion/sst.config.ts
DamienPace15: I can add one in. 

The OG PR has the following in the root `.gitignore`


```
# SST Env files from examples
examples/*/sst-env.d.ts
```

platform/src/auto/run.ts
DamienPace15: If you want to keep the peering resource then yes. It doesn't contain any data. 

Currently what it will do in multi region is the following

```
Before Delete
[Cluster 1] -> [Peering relationship] -> [Cluster 2]

After Delete 
[Cluster 1] [Cluster 2]
```

The peering relationship is easy to recreate if needed. 

If we add it then it will keep the multi region relationship.

Is this function strictly for resources that hold data? If so Cluster peering isn't the right spot for this imo.

platform/src/components/aws/dsql.ts
DamienPace15: It's referenced in nodes on line 332. 

`peerProvider: this.peerProvider,`

If it's removed then we can't access underlying resources I do believe.

platform/src/components/aws/dsql.ts
DamienPace15: Have done some quick testing on this. 

Going from multi region to single region cluster: Deletes both clusters
Going from single cluster to multi cluster: Deletes single cluster and recreates it then creates the peering connection. 

Will need to document this somewhere for sure.

examples/aws-dsql-multiregion/sst.config.ts
DamienPace15: It can't connect because the VPC wont have access to the internet unless we use a NAT gateway VPC. 

I will do VPC support now via VPC endpoints.

platform/src/components/aws/dsql.ts
DamienPace15: have reworked this so that it can get peering clusters.

 const cluster = sst.aws.Dsql.get('MyCluster', {
      identifier: 'clusterId',
      peerIdentifier: 'peerClusterId',
      peerRegion: 'us-east-2',
    });

that works and I was able to get it correctly

examples/aws-dsql-multiregion/sst.config.ts
DamienPace15: Able to get VPC support working now for single region @vimtor. 

Happy to work on multi region after some discussions and avoid this PR getting heavy. Rather tackle it in a seperate issue.

platform/src/components/aws/dsql.ts
```diff
@@ -0,0 +1,511 @@
...
+   *     instance: vpc,
+   *     endpoints: {
+   *       management: true,
+   *       connection: true,
+   *     }
+   *   }
+   * });
+   * ```
+   */
+  vpc?:
```
DamienPace15, line 77: @vimtor have normalised the VPC to take vpc shorthanded or with the instance and endpoint configs. 

Also defaults to connection as true.

platform/src/components/aws/dsql.ts
DamienPace15: @vimtor this is not blocking, can probably add a to do to do this. 

I have a pretty busy schedule with life/work the next 1-1.5 weeks. Feel free to jump in and change this if you want but happy to return at a later date to get this pushed.

PR #6500, opened 2026-03-01
Fix `sst diff` always reporting changes to `lambda.Permission` with qualified ARNs

### LINKED ISSUE #6154: Bug: sst diff always reports changes to aws:lambda:Permission when using fully-qualified function ARNs with aliases
Reproduction here: https://github.com/noisyscanner/sst-thing

## Background

We create a `live` alias for each of our deployed Lambda functions.
Any resources that point to the function, for example crons, bucket notifications, or API Gateway routes always point to this alias rather than `$LATEST` or a specific version.
This ensures that newly deployed function versions are smoke tested before they become active.

This is done using `aws.lambda.Alias`, as SST does not appear to provide this natively.

## The problem

Running an `sst diff` always shows some changes to `aws:lambda:Permission`, as it thinks we are setting the `function` to the fully-qualified function ARN with the alias at the end, and have removed `qualifier`.

Perhaps SST resources like `sst.aws.Cron` could either parse out the alias in this case, or alternatively accept a new type for `function` like `{ name: arn; qualifier: string; }`.

## Reproduction steps

1. Clone the repo
2. Run `sst deploy`
3. Run `sst diff`

### Expected behaviour

No changes, as we are running a diff without changing anything.

### Actual behaviour

Diff output as below:

```

SST 3.17.14  ready!

➜  App:        sst-thing
   Stage:      brad

~  Diff

|  Created     MyCron sst:aws:Cron → MyCronPermission aws:lambda:Permission
|  Deleted     MyCron sst:aws:Cron → MyCronPermission aws:lambda:Permission

↗  Permalink   https://sst.dev/u/xxxxxx

✓  Generated

+  MyCron sst:aws:Cron → MyCronPermission aws:lambda:Permission
   * function = arn:aws:lambda:eu-west-2:<account-id>:function:sst-thing-brad-MyFunctionFunction-bamswodc:MyFunctionAlias-c76baba
   - qualifier
```

vimtor:
closes #6154

<details>

| **Before** | **After** |
|------------|-----------|
|  CleanShot 2026-03-01 at 19 22 16@2x         |  CleanShot 2026-03-01 at 19 22 47@2x        |

</details>

when sst components pass a qualified lambda arn (e.g. ...function:myfn:live) to lambda.permission's function field, aws quietly splits it into function=...function:myfn plus qualifier=live. on the next sst diff, pulumi notices the stored state doesn't line up with what you asked for, so it flags a change every single time — even though nothing really moved

the fix is to split the qualified arn before handing it to pulumi, feeding function and qualifier separately so the desired state lines up with what aws keeps

### DIFF
--- a/platform/src/components/aws/apigateway-websocket-route.ts
+++ b/platform/src/components/aws/apigateway-websocket-route.ts
@@ -11,6 +11,7 @@ import { FunctionArgs, FunctionArn } from "./function";
 import { ApiGatewayWebSocketRouteArgs } from "./apigateway-websocket";
 import { apigatewayv2, lambda } from "@pulumi/aws";
 import { FunctionBuilder, functionBuilder } from "./helpers/function-builder";
+import { splitQualifiedFunctionArn } from "./helpers/arn";
 
 export interface Args extends ApiGatewayWebSocketRouteArgs {
   /**
@@ -94,7 +95,8 @@ export class ApiGatewayWebSocketRoute extends Component {
         `${name}Permissions`,
         {
           action: "lambda:InvokeFunction",
-          function: fn.arn,
+          function: fn.arn.apply((arn) => splitQualifiedFunctionArn(arn).unqualifiedArn),
+          qualifier: fn.arn.apply((arn) => splitQualifiedFunctionArn(arn).qualifier),
           principal: "apigateway.amazonaws.com",
           sourceArn: interpolate`${api.executionArn}/*`,
         },
--- a/platform/src/components/aws/apigateway-websocket-route.ts
+++ b/platform/src/components/aws/apigateway-websocket-route.ts
@@ -96,7 +96,7 @@ export class ApiGatewayWebSocketRoute extends Component {
         {
           action: "lambda:InvokeFunction",
           function: fn.arn.apply((arn) => splitQualifiedFunctionArn(arn).unqualifiedArn),
-          qualifier: fn.arn.apply((arn) => splitQualifiedFunctionArn(arn).qualifier),
+          qualifier: fn.arn.apply((arn) => splitQualifiedFunctionArn(arn).qualifier!),
           principal: "apigateway.amazonaws.com",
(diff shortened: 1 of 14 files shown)


platform/src/components/aws/apigateway-websocket-route.ts
```diff
@@ -94,7 +95,8 @@ export class ApiGatewayWebSocketRoute extends Component {
         `${name}Permissions`,
         {
           action: "lambda:InvokeFunction",
-          function: fn.arn,
+          function: fn.arn.apply((arn) => splitQualifiedFunctionArn(arn).unqualifiedArn),
+          qualifier: fn.arn.apply((arn) => splitQualifiedFunctionArn(arn).qualifier!),
```
jamesgibbons92, line 99: To be more type-safe (i.e. remove the non-null assertion), we could initialize the permission component inside an `fn.arn.apply()` block.
```
      fn.arn.apply((arn) => {
        const { qualifier, unqualifiedArn } = splitQualifiedFunctionArn(arn);
        return new lambda.Permission(
          `${name}Permission`,
          {
            action: "lambda:InvokeFunction",
            function: unqualifiedArn,
            qualifier: qualifier,
            principal: "events.amazonaws.com",
            sourceArn: rule.arn,
          },
          { parent },
        );
      });
```
it's a bit more code but is more strictly correct

vimtor: yeah it's a nuisance because pulumi doesn't accept `Output<undefined>` but it accepts `undefined`

i believe however that your suggestion goes against pulumi's best practices: https://github.com/pulumi/agent-skills/blob/main/authoring/skills/pulumi-best-practices/SKILL.md#1-never-create-resources-inside-apply

vimtor: sorry i meant to say "inconvenience" not "nuisance"

my spanish → english failed a bit haha

Issue #6493, opened 2026-02-26
Cloudfront viewer request injection leaves little room for customization


troendheim:
When trying to customize viewer request injection on a router like so:
```ts
new sst.aws.Router("router", {
  edge: {
    viewerRequest: {
      injection: "// write a few lines of code here"
  }
}
```
you will very quickly (I would argue <10 lines of normal code) reach the limit of 10KB function size (https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/cloudfront-limits.html#limits-functions) which leads to cloudfront rejecting deploys. That in turn forces us to write unreadable "minified" one-liners which will still mean trouble if there's ever just one more line added to the function by the framework.

I propose either:
- A cleanup of the code trying to reduce character count
- That the framework code is minified before publishing to AWS.

It seems like everything is generated here: https://github.com/anomalyco/sst/blob/1b060504abebfd6425083e24a661c186e8e1ffb9/platform/src/components/aws/router.ts#L1470

PR #6506, opened 2026-03-02
Minify cf router injection

Hjorthdx:
This PR uses the transform function from esbuild to minifiy the CF_ROUTER_INJECTION, so that users are less likely to hit the 10KB function size limit. This problem is described in issue #6493.


vimtor:
thanks for your contribution @Hjorthdx 

closing in favor of https://github.com/anomalyco/sst/pull/6508

PR #6508, opened 2026-03-02
Minify CloudFront viewer request to free up space for user injection

### LINKED ISSUE #6493: Cloudfront viewer request injection leaves little room for customization

vimtor:
closes #6493

this saves ~30% for each edge function which adds up to ~120 lines of code in total (currently ~15 are supported for the router handler)

we duplicated the shared utils but deduplicating it will bring more complexity that doesn't seem worth it


vimtor:
thinking that it might not be worth breaking the pattern just for this, maybe i try a way simpler approach like stripping comments

troendheim:
> thinking that it might not be worth breaking the pattern just for this, maybe i try a way simpler approach like stripping comments

30% reduction is very nice, so I am a bit sad to see that you're considering a different direction again, where we most likely only will gain a handful of lines (if only comments are removed). I think we'll quickly end up in the same spot again.

Another thing that could also be done to reduce character counts are simplifying deep repeated property read/write like event.request.headers. But then we would be in the process of manually minifying our code in the framework. Are you sure an approach like https://github.com/anomalyco/sst/pull/6506 is not a viable way forward?

vimtor:
i've landed on a very similar approach as #6506

i'm a bit skeptical of running esbuild at deploy time but the other solutions have too much complexity (build steps, variable substitution, etc.)

i'll merge this and see how it goes

Issue #6471, opened 2026-02-25
Feat: Suppport for Lambda Durable Functions, Including Live Mode Support


vitorfigmgs:
# RFC: Support for AWS Lambda Durable Functions in SST

In December 2025, AWS announced one of its most powerful recent features: [Lambda Durable Functions](https://docs.aws.amazon.com/lambda/latest/dg/durable-functions.html). It would be awesome to have native support for this in SST. As a big SST fan, it would be an honor to contribute this as my first open-source contribution.

# Interface

We could enable this via a `durable` property in the `sst.aws.Function` config. It could accept the [durableConfig](https://www.pulumi.com/registry/packages/aws/api-docs/lambda/function/#durableconfig_nodejs) properties from Pulumi or a simple boolean for default configuration:

```typescript
{
  link: [bucket],
  durable: true // or { ...durableConfig }
}
```

vitorfigmgs:
This PR solves that (#6510).

Actually, the checkpointing system from Durable Functions is much more stateless and simpler than I expected. Functions running locally with withDurableExecution can easily manage this checkpoint system on behalf of the bridge function, as long as all events are correctly proxied to them (all logic from withDurableExecution are basically calls to CheckpointDurableExecution and GetDurableExecutionState using the CheckpointToken and the DurableExecutionArn it receives from the Lambda Service as event).

So, the Go bridge in SST handles that. The issue was that durable configurations cannot be used by functions using custom runtimes. The idea was to create a Node.js bridge function for the durable functions that executes the Go Bridge binary.

PR #6510, opened 2026-03-03
Add Durable Functions support

vitorfigmgs:
Solves #6471

# Support for AWS Lambda Durable

- [X] Add durable configuration to the Function component
- [X] Make durable functions work locally
- [X] Add a basic example
- [x] Test and create examples with functions using the Python runtime
- [x] Customize `getSSTLink` to provide callback command permissions and include the version in the invoke command
- [x] Add Docs
- [x] Check Issue with function URL
- [x] Make functions URLS to work in durable Functions.

# The NodeJS bridge
Current, [Durable functions are supported only on the Node.js and Python runtimes and Docker images](https://docs.aws.amazon.com/lambda/latest/dg/durable-supported-runtimes.html). SST deploys functions on the Amazon Linux 2023 runtime in dev mode, which does not support durable functions.

To solve that, in dev mode, when the function is a durable function, we are using the [nodejs-bridge](https://github.com/vitorfigmgs/sst/blob/add-durable-function-support/platform/functions/nodejs-bridge/index.mjs). It's a proxy to the Go bridge, since the function simply runs the Go binary.


vimtor:
@vitorfigmgs looking good 👀

anatolzak:
This is amazing! @vitorfigmgs 
Looking forward for this to get merged!

platform/src/components/aws/function.ts
```diff
@@ -2571,11 +2644,28 @@ export class Function extends Component implements Link.Linkable {
         const isOac = output(url.route?.routerProtection).apply(
           (p) => p?.mode === "oac" || p?.mode === "oac-with-edge-signing",
         );
+        const functionName = all([fn.name, durable]).apply(([name, durable]) =>{
+          const needsQualifiedArn = !!durable;
+          if(needsQualifiedArn){
+            // AWS expects function URLs for durable functions to use qualified ARNs.
```
vitorfigmgs: # Durable Functions + Lambda Function URL Issue

I'm disabling URLs for Durable Functions even though they should work
perfectly with Lambda Function URLs. However, I found some strange
Pulumi issues that prevent us from using URLs with Durable Functions.

If `lambda.FunctionUrl` receives a function name from a Durable Function
**without** the version qualifier, it throws this error:

    Durable sst:aws:Function → DurableUrl aws:lambda:FunctionUrl
    sdk-v2/provider2.go:572: sdk.helper_schema: creating Lambda Function URL (aws-lambda-durable-vitor-DurableFunction-xcuoavxt): operation error
    Lambda: CreateFunctionUrlConfig, https response error StatusCode: 400, RequestID: 613cd9be-6564-4e9f-bf4e-0e1740c80314, InvalidParameterValueException:
    You cannot invoke a durable function using an unqualified ARN.: provider=aws@7.20.0

Durable Functions usually require the version qualifier to be specified.
Nothing too unusual here, but:

This should solve it:

``` ts
const fnUrl = new lambda.FunctionUrl(`${name}Url`, {
  functionName: fn.name,
  qualifier: fn.version, // <---- The qualifier, usually "$LATEST"
});
```

However, we get this strange error:

```
    Durable sst:aws:Function → DurableUrl aws:lambda:FunctionUrl
    sdk-v2/provider2.go:572: sdk.helper_schema: creating Lambda Function URL (aws-lambda-durable-vitor-DurableFunction-xcuoavxt/$LATEST): operation error
    Lambda: CreateFunctionUrlConfig, https response error StatusCode: 400, RequestID: 5c7fbe8d-f2c4-49a9-92ef-14d16cb0a29d, api error ValidationException:
    1 validation error detected: Value '$LATEST' at 'qualifier' failed to satisfy constraint:
    Member must satisfy regular expression pattern: ((?!^\d+$)^[0-9a-zA-Z-_]+$): provider=aws@7.20.0
```

Then I tried using the ARN with the version qualifier.

This should solve it (and would work if we could pass it directly to the
AWS API):

``` ts
const functionName = all([fn.name, fn.arn, fn.version, durable]).apply(
  ([name, arn, version, durable]) => {
    const needsQualifiedArn = !!durable;

    if (needsQualifiedArn) {
      return `${arn}:${version}`;
    }

    return name;
  }
);
```

It should also work since other components, such as the IAM Role, accept
this format. However, we get:

```
Durable sst:aws:Function → DurableUrl aws:lambda:FunctionUrl
sdk-v2/provider2.go:572: sdk.helper_schema: creating Lambda Function URL (arn:aws:lambda:us-east-1:************:function:aws-lambda-durable-v
itor-DurableFunction-mxucsdec:$LATEST): operation error Lambda: CreateFunctionUrlConfig, https response error StatusCode: 400, RequestID: 9ef
1bd51-138e-4947-8712-5fd48977f345, api error ValidationException: 1 validation error detected: Value 'arn:aws:lambda:us-east-1:************:f
unction:aws-lambda-durable-vitor-DurableFunction-mxucsdec:$LATEST' at 'functionName' failed to satisfy constraint: Member must satisfy regula
r expression pattern: (arn:(aws[a-zA-Z-]*)?:lambda:)?((eusc-)?[a-z]{2}((-gov)|(-iso([a-z]?)))?-[a-z]+-\d{1}:)?(\d{12}:)?(function:)?([a-zA-Z0
-9-_]{1,64})(:((?!\d+$)[0-9a-zA-Z-_]+))?: provider=aws@7.20.0
```

vitorfigmgs: Solved! I created an alias pointing to $Latest to bypass this broken regex in Pulumi.

platform/src/components/aws/durable.ts
```diff
@@ -0,0 +1,10 @@
+export const DURABLE_CHECKPOINT_RETENTION = {
```
vimtor: any reason we hardcoded this instead of using the `duration.ts` utility?

platform/src/components/aws/function.ts
```diff
@@ -129,6 +130,17 @@ export type FunctionPermissionArgs = {
...
 
+export type DurableFunctionArgs = {
+  /**
+   * Maximum execution time for the durable function
+   */
+  executionTimeout?: Input<Duration>;
+  /**
+   * Number of days to retain the function's execution state.
+   */
+  retentionPeriod?: Input<keyof typeof DURABLE_CHECKPOINT_RETENTION>;
```
vimtor: sst usually prefers single names whenever possible

i would rename these to `timeout` and `retention`

platform/src/components/aws/function.ts
```diff
@@ -2791,6 +2897,13 @@ export class Function extends Component implements Link.Linkable {
     return this.function.arn;
   }
 
+  /**
+   * Whether the Lambda function is a [Durable Function](https://docs.aws.amazon.com/lambda/latest/dg/durable-functions.html).
+   */
+  public get isDurable() {
+    return this.function.durableConfig !== undefined;
```
vimtor: if i'm not mistaken `this.function.durableConfig` is a Output<outputs.lambda.FunctionDurableConfig | undefined>

comparing to `undefined` will always result to `true`

platform/src/components/aws/function.ts
```diff
@@ -2869,15 +2982,32 @@ export class Function extends Component implements Link.Linkable {
...
+    let arnReference = this.arn;
+    const needsQualifiedArn = this.isDurable;
+
+    if (needsQualifiedArn) {
+      arnReference = interpolate`${this.arn}:*`;
+    }
+
+    const allowedActions = ["lambda:InvokeFunction"];
+
+    if (this.isDurable) {
```
vimtor: related to the previous comment this will break

it's quite hard to fix this properly, since getSSTLink needs to be sync

we might even need an `sst.aws.Durable` or something to fix this properly

alternatively we can always give these permissions, i guess it would be fine since we already give permission to invoke the lambda anyway?

platform/src/components/aws/function.ts
```diff
@@ -2694,6 +2793,13 @@ export class Function extends Component implements Link.Linkable {
       });
     }
 
+    function createAlias(infix = "Latest", version = fn.version) {
```
vimtor: shouldn't this alias be durable-related in naming?

vimtor (changes requested):
thanks for your contribution @vitorfigmgs 

this looks very promising. i left a bunch of minor comments, but the core functionality works

could you also explain a bit about the lambda bridge stuff you changed? why did you find it necessary to do so?

anatolzak:
> could you also explain a bit about the lambda bridge stuff you changed? why did you find it necessary to do so?

@vitorfigmgs correct me if I'm wrong, @vimtor  Durable functions are supported only on the Node.js and Python runtimes and Docker images. Currently, SST deploys functions on the Amazon Linux 2023 runtime in dev mode, which does not support durable functions. To enable proxying durable function events from Lambda to the local machine, you need to use a supported runtime like Node.js. Based on the changes in the PR, we are using the Node.js runtime to allow the AWS Lambda function to proxy a durable event to the local machine. The Node.js handler on Lambda acts as a wrapper around the regular binary we use on AWS lambda in dev mode, allowing us to bypass the runtime constraint for durable functions.

platform/src/components/aws/durable.ts
vitorfigmgs: I did this based on what I saw in [logging.ts](https://github.com/anomalyco/sst/blob/dev/platform/src/components/aws/logging.ts), I thought this was the way SST exposed duration properties with day as units.

I made a commit using the DurationDays and created an toDays. Is this the preferred way for this implementation?

platform/src/components/aws/function.ts
vitorfigmgs: Solved in:

`fix: Improve the naming of durable function configuration properties`

platform/src/components/aws/function.ts
vitorfigmgs: Solved in:

`fix: Make isDurable work correctly`

platform/src/components/aws/function.ts
vitorfigmgs: Fixed. It looks like we'll have to go that route.

We could do:
```typescript
this._isDurable = !!args.durable;
```
However, for some reason, I have an intuition that this could break in some contexts, like references to functions outside the stack, even though I haven't seen any implementations in SST that actually do that. 

Regardless, as you mentioned, that would be purely for philosophical reasons, as we can assume that if a resource has permission to invoke the function, it's fine to allow those Commands related to Durable Functions. When the function is not durable, they simply won't work anyway.


We can address this (possibly fixing using just !!args.durable) in another PR if those unused permissions become a problem in some context.

platform/src/components/aws/function.ts
vitorfigmgs: Yes, they should. Solved in:

`fix: Make alias creation be durable related`

vitorfigmgs:
It is exactly that, @anatolzak.

I will add this explanation to the PR description to make it clearer.

platform/src/components/aws/durable.ts
vitorfigmgs: Well… looking at the codebase, I see that logging uses this object because `aws.cloudwatch.LogGroup` doesn’t accept arbitrary retention-day values, and the other resource with day properties uses the DurationDays type.

So Solved :heavy_check_mark:

vimtor (approved):
thanks for your contribution @vitorfigmgs

i made a couple of minor changes and everything's working like a charm

eventually i will want to add some sdk helpers so please feel free to share how you'll use this new feature

PR #6515, opened 2026-03-03
Add support for setting Cloudflare Worker secrets

Fadi729:
This PR adds typed Cloudflare Worker environment bindings so env vars can be configured as either plain_text or secret_text, while preserving full backward compatibility.

### Example Usage

```ts
const api= new sst.cloudflare.Worker("api", {
  handler: "src/worker.ts",
  environment: {
    ENV_TEXT: "Hello, world!",
    ENV_SECRET: {
      type: "secret_text",
      text: "Can you see this?",
    },
  },
});
```

### DIFF
--- a/platform/src/components/cloudflare/worker.ts
+++ b/platform/src/components/cloudflare/worker.ts
@@ -29,6 +29,12 @@ import { getContentType } from "../base/base-site";
 import { physicalName } from "../naming";
 import { existsAsync } from "../../util/fs";
 
+type WorkerEnvironmentTypedValue = {
+  type: "plain_text" | "secret_text";
+  text: string;
+};
+type WorkerEnvironmentValue = Input<string | WorkerEnvironmentTypedValue>;
+
 export interface WorkerArgs {
   /**
    * Path to the handler file for the worker.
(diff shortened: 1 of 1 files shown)


vimtor:
thanks for your contribution @Fadi729 

do you have a specific use case for this? i'm checking and `sst.Secret` is already passed as an encrypted secret

Fadi729:
Oh I totally missed that. I thought `sst.Secret` injects the value at runtime since the values are (correct me if I'm wrong) encrypted and stored in S3 (or R2). In that case I can do something like the following:
```ts
const secret = sst.Secret("ENV_SECRET", process.env.ENV_SECRET)
const api= new sst.cloudflare.Worker("api", {
  handler: "src/worker.ts",
  link: [secret],
  environment: {
    ENV_TEXT: "Hello, world!"
  }
});
```

The only issue I have with this is that I would need to `sst secrets add` every secret I have in CI (or write them to a file and then use `load`) or just use the fallback value as the actual value since it would just work as is. Anyways this is something for me to figure out 😅.

I'll close the PR.
Thanks!

PR #6519, opened 2026-03-03
Start dev commands immediately using cached deployment state

### LINKED ISSUE #6518: sst dev commands no longer start immediately on startup
## Description

  Since upgrading from v3.16.3, dev commands configured in `sst.config.ts` no longer start immediately when running `sst dev`. Instead, they wait for the first deployment to complete, which can take upwards of two minutes.

  Previously, dev commands would launch right away using the cached state from the previous deployment, meaning frontend apps and other dev processes were available almost instantly.

  ## Steps to reproduce

  1. Configure a project with dev commands (e.g. a Vite frontend)
  2. Run `sst dev`
  3. Observe that dev commands in the sidebar don't start until deployment finishes

  ## Expected behavior

  Dev commands should start immediately when `sst dev` is run, using the cached deployment state from the previous run.

  ## Actual behavior

  Dev commands wait for the first deployment to complete before starting.

  ## Cause

  Introduced in #6402 (commit 2180e81d1). The fix for a race condition also added an `if evt.Old { continue }` check that skips the cached `CompleteEvent`, blocking dev commands from starting eagerly.

anatolzak:
closes https://github.com/anomalyco/sst/issues/6518

## Summary

  - Restore the behavior where dev commands start immediately when running `sst dev`, using the cached deployment state from the previous run
  - This was a regression introduced in #6402 where the `evt.Old` check was added, causing dev commands to wait for the first deployment to complete (~2 minutes)

  ## Context

  Commit 2180e81d1 (#6402) fixed a race condition where the multiplexer could miss events by moving `bus.Subscribe` before `deployer.Start`. However, it also added an `if evt.Old { continue }` check that skips the cached
  `CompleteEvent`, which prevented dev commands from starting eagerly.

  The `evt.Old` check is unnecessary because `AddProcess` is already idempotent by key — if a process is already running, calling `AddProcess` again with the same key is a no-op. This means it's safe to start dev commands on
   the cached event and then receive the new event after deployment without any side effects.

  ## Before & After

  ### Before (depending on the number of resources, it can take up to two minutes for the dev command to start)

https://github.com/user-attachments/assets/5b9b666b-dc27-4dfb-aafa-09ba38ce8dab

  ### After (takes three seconds for the dev command to start)

https://github.com/user-attachments/assets/7c83fb80-eab0-4901-b38c-26e326fbc200

  ## Test plan (tested and verified)

  - Run `sst dev` in a project with dev commands configured
  - Verify dev commands appear in the sidebar and start running immediately, before deployment completes
  - Verify that after deployment completes, dev commands continue running without interruption

### DIFF
--- a/cmd/sst/mosaic.go
+++ b/cmd/sst/mosaic.go
@@ -273,9 +273,6 @@ func CmdMosaic(c *cli.Cli) error {
 				case unknown := <-evts:
 					switch evt := unknown.(type) {
 					case *project.CompleteEvent:
-						if evt.Old {
-							continue
-						}
 						for _, d := range evt.Devs {
 							if d.Command == "" {
 								continue
@@ -338,9 +335,6 @@ func CmdMosaic(c *cli.Cli) error {
 				case unknown := <-evts:
 					switch evt := unknown.(type) {
 					case *project.CompleteEvent:
-						if evt.Old {
-							continue
-						}
 						for _, d := range evt.Devs {
 							if d.Command == "" {
 								continue


vimtor (approved):
@anatolzak thanks for your contribution!

<details>

<summary>this has the side effect of restarting finished dev processes after the first deploy</summary>

https://github.com/user-attachments/assets/3213fd96-247d-4e0f-8851-dbc834af4000

</details>

something that shouldn't happen too often in practice, so the trade-off is worth it. just saying in case you wanna tackle it in a separate pr

PR #6530, opened 2026-03-04
Fix broken platform tests

vimtor:
for the moment, i've decided to remove the pulumi component tests because they're so unreliable, full of weird errors and unhandled promises

let's stick with unit tests for now and find a better solution at some point


vimtor:
closing because gpt-5.4 gave me hope (stay tuned)

PR #6544, opened 2026-03-06
Optimize JSON validation of Pulumi state

jasonrowsell:
Before every s3 upload during a deploy, SST validates the state file by calling `json.Unmarshal`. This builds a complete go object graph in memory only to check syntax, then disposes of it. I checked with a 200-resource stack and this allocates ~862KB on the heap

Replace with `json.Valid` and still call `json.Unmarshal` on failure path to recover the specific parse error

### DIFF
--- a/pkg/project/provider/provider.go
+++ b/pkg/project/provider/provider.go
@@ -191,20 +191,39 @@ func PutSecrets(backend Home, app, stage string, data map[string]string) error {
 
 func PushPartialState(backend Home, updateID, app, stage string, data []byte) error {
 	slog.Info("pushing partial state", "updateID", updateID)
-	err := json.Unmarshal(data, &map[string]interface{}{})
-	if err != nil || len(data) == 0 {
-		return fmt.Errorf("something has corrupted the state file - refusing to upload: %w", err)
+	if err := validateStateJSON(data); err != nil {
+		return err
 	}
 	return backend.putData("app", app, stage, bytes.NewReader(data))
 }
 
 func PushSnapshot(backend Home, updateID, app, stage string, data []byte) error {
 	slog.Info("pushing snapshot", "updateID", updateID)
-	err := json.Unmarshal(data, &map[string]interface{}{})
-	if err != nil {
-		return fmt.Errorf("something has corrupted the state file - refusing to upload: %w", err)
+	if err := validateStateJSON(data); err != nil {
+		return err
 	}
 	return backend.putData("snapshot", app, stage+"/"+updateID, bytes.NewReader(data))
+}
+
+// validateStateJSON checks that data is a non-empty, valid JSON object.
+// Pulumi state files are always JSON objects (VersionedCheckpoint structs);
+// anything else indicates corruption or a truncated write.
+func validateStateJSON(data []byte) error {
+	if len(data) == 0 {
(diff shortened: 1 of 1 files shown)


pkg/project/provider/provider.go
```diff
@@ -191,22 +191,41 @@ func PutSecrets(backend Home, app, stage string, data map[string]string) error {
...
+	if len(data) == 0 {
+		return fmt.Errorf("something has corrupted the state file - refusing to upload: file is empty")
+	}
+	if data[0] != '{' {
+		return fmt.Errorf("something has corrupted the state file - refusing to upload: expected JSON object, got %q", data[0])
+	}
+	if !json.Valid(data) {
+		// Use Unmarshal on the failure path only to recover the parse error detail
+		// (position, character). The hot path uses json.Valid for zero allocations.
+		var m json.RawMessage
```
vimtor, line 215: we changed the type from `&map[string]interface{}{}` to `json.RawMessage`

was that intentional?

pkg/project/provider/provider.go
```diff
@@ -191,22 +191,41 @@ func PutSecrets(backend Home, app, stage string, data map[string]string) error {
...
+// validateStateJSON checks that data is a non-empty, valid JSON object.
+// Pulumi state files are always JSON objects (VersionedCheckpoint structs);
+// anything else indicates corruption or a truncated write.
+func validateStateJSON(data []byte) error {
+	if len(data) == 0 {
+		return fmt.Errorf("something has corrupted the state file - refusing to upload: file is empty")
+	}
+	if data[0] != '{' {
+		return fmt.Errorf("something has corrupted the state file - refusing to upload: expected JSON object, got %q", data[0])
+	}
```
vimtor:
```suggestion
```

i wonder if we still need this given we have the optimized `json.Valid` seems like an over optimization

i would assume that json.Valid will exit early in these types of cases

pkg/project/provider/provider.go
```diff
@@ -191,22 +191,41 @@ func PutSecrets(backend Home, app, stage string, data map[string]string) error {
...
-	err := json.Unmarshal(data, &map[string]interface{}{})
-	if err != nil {
-		return fmt.Errorf("something has corrupted the state file - refusing to upload: %w", err)
+	if err := validateStateJSON(data); err != nil {
+		return err
 	}
 	return backend.putData("snapshot", app, stage+"/"+updateID, bytes.NewReader(data))
 }
 
+// validateStateJSON checks that data is a non-empty, valid JSON object.
```
vimtor: the method is very self explanatory

vimtor (changes requested):

pkg/project/provider/provider.go
```diff
@@ -191,22 +191,35 @@ func PutSecrets(backend Home, app, stage string, data map[string]string) error {
...
+		return err
 	}
 	return backend.putData("snapshot", app, stage+"/"+updateID, bytes.NewReader(data))
 }
 
+func validateStateJSON(data []byte) error {
+	if len(data) == 0 {
+		return fmt.Errorf("something has corrupted the state file - refusing to upload: file is empty")
+	}
+	if !json.Valid(data) {
```
vimtor, line 212: also verifying that its' valid doesn't ensure it follows the `&map[string]interface{}{}` shape

vimtor (changes requested):

PR #6553, opened 2026-03-07
Feat: Bastion component

### LINKED ISSUE #6364: Request: Allow for the creation of a bastion for an existing VPC.
Hello,

I was very excited about using your system for live development but need to work with a MSSQL DB on an existing VPC.  It is only possible to create a bastion on a new VPC.  Can you add the ability to add a bastion to an existing vpc?

jamesgibbons92:
Closes #6364 

# Summary:

- Adds a new component 'Bastion' which is a standalone component for deploying an ec2 bastion acting as a tunnel into VPC protected resources. This is functionally equivalent to the bastion created in the VPC component, but has some additional improvements and can support a non-sst vpc by passing in the required props.
- Adds a version number to `sst tunnel` which prompts the user to run `sudo sst tunnel install` again if the binary code changes, this is just a hardcoded version variable which needs to be bumped when changes are made.
- `sst tunnel` now supports multiple bastions in a stack, this is probably a very small use case - but previously the command would just use the last _tunnel it found in the stack. 

## SSM Support
The tunnel code has been extended to support session-manager port forwarding, this is more secure than ssh as it does not require any public facing server. https://aws.amazon.com/blogs/aws/new-port-forwarding-using-aws-system-manager-sessions-manager/
- By default the bastion will deploy in 'ssh' mode, similar to the existing bastion.
- Opt in to `ssm` by passing an arg into the component.
- The only caveat to this mode is that it requires the aws-cli session-manager-plugin to be installed on the client system.

## Notes:
- The existing bastion behaviour in VPC component is unchanged. I left this untouched to reduce the risk of regression issues, this new component is more useful for people needing to tunnel into a non-sst vpc, or if they want to use the more secure `ssm` tunnel mechanism.

# Test plan:

Vpc with `bastion: true` (regression test, ensure the new tunnel code is backwards compatible):
image

Vpc with Bastion component in SSH mode:
image

Vpc with Bastion component in SSM mode:
image

Vpc with Shared Bastion component:
image

SST Dev:
image

Tunnel needs upgrade:
image

SSM mode but session manager plugin not installed:
image

Todo:
- [ ] Test MacOS

### DIFF
--- a/cmd/sst/tunnel.go
+++ b/cmd/sst/tunnel.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"log/slog"
 	"os"
+	"os/exec"
 	"os/user"
 	"strings"
 
@@ -30,7 +32,7 @@ var CmdTunnel = &cli.Command{
 			"```",
 			"",
 			"If your app has a VPC with `bastion` enabled, you can use this to connect to it.",
-			"This will forward traffic from the following ranges over SSH:",
+			"This will forward traffic from the following ranges using either SSH or SSM, depending on your bastion configuration:",
 			"- `10.0.4.0/22`",
 			"- `10.0.12.0/22`",
 			"- `10.0.0.0/22`",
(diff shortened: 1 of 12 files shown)


jamesgibbons92:
@vimtor circling back to this now, 
apart from the merge conflicts, are there any other changes you think I should make to this pr to get it merged?

PR #6559, opened 2026-03-09
Prevent `sst dev` from running on non-personal stages

vimtor:


nowlena:
Could a flag be added to the CLI ignore this requirement?

This update seemingly causes a few problems for us:

1. our team cannot now run the same application using the same stage (ie. "default")
   - we each have separate AWS accounts so this works and requires no code change per dev to run the application
   - I believe we would have to move to having an ENV variable for SST_STAGE with this change since each of us are required to use our own names for the stage and that's it

3. how can we deploy multiple local versions of an app if the stage must be a pre-defined single value?
   - we would have to change the application name in order to now deploy different versions of the same application?

A flag that bypasses this requirement would be great.

Thanks!

vimtor:
@nowlena to be honest i was unsure about this change

the main thing for me was preventing `sst dev --stage production` which could really mess up your whole environment

i will undo this change today and think more about it. ideas welcomed!

PR #6566, opened 2026-03-10
Fix Cloudflare StaticSite SPA routing and docs

### LINKED ISSUE #5947: can't set Cloudflare static assets "not-found-handling" config parameter
I just saw that the Cloudflare Worker SST component can deploy assets and ran a test with static files and a small hono script. The hono part worked fine, but it always ran and none of the static files were being served. Realized it was the `"not-found-handling": "single-page-application"` parameter usually passed to assets in the wrangler config file, so dug through the Pulumi layers to figure out the transform.

```js
  async run() {
    const worker = new sst.cloudflare.Worker("Workit", {
      handler: "./index.ts",
      url: true,
      assets: {
        directory: "./dist",
      },
      transform: {
        worker: {
          assets: {
            config: {
              notFoundHandling: "single-page-application",
            },
          },
        },
      },
    })
```

>  ✕  Failed                                                                                        
>                                                                                                  
> Workit sst:cloudflare:Worker → WorkitScript cloudflare:index:WorkersScript
> [AttributeName("assets").AttributeName("config")] Value Conversion Error: An unexpected error was encountered trying to convert tftypes.Value into workers_script.WorkersScriptMetadataAssetsConfigModel. This is always an error in the provider. Please report the following to the provider developer:
> error retrieving field names from struct tags: assets.config._headers: invalid tfsdk tag, must only use lowercase letters, underscores, and numbers, and must start with a letter

Seems to be related to the [tf bug](https://github.com/cloudflare/terraform-provider-cloudflare/issues/5616#issuecomment-2928724220) @fwang is aware of, but wanted to get this in here as another case to test when the upstream issue is resolved.

### LINKED ISSUE #6691: sst.cloudflare.StaticSite and sst.cloudflare.x.StaticSite do not work in dev
Still experimental but also observed the same error with the regular `sst.cloudflare.StaticSite`.

1. Create `sst.config.ts` with:

```typescript
export default $config({
  app(input) {
    return {
      name: "repro",
      protect: ["production"].includes(input?.stage),
      home: "cloudflare",
    };
  },
  async run() {
      const web = new sst.cloudflare.x.StaticSite("Web");
  },
});

```

2. Run `sst dev`

3. Observe:

```
Web sst:cloudflare:StaticSite → WebRouterScript cloudflare:index:
WorkersScript
AttributeName("assets").AttributeName("asset_manifest_sha256"): E
rror reading asset files - lstat /Users/ernest/Developer/repro/Users/ernest/Developer/repro/.sst/platform/functions/empty-site: no such file or dir
ectory
```

Issue seems to be that `sst.cloudlare.x.StaticSite` creates an [absolute output path during dev](https://github.com/anomalyco/sst/blob/c36553cf648fa0ec748197bae0ad5035000421a7/platform/src/components/cloudflare/experimental/static-site.ts#L215) which is then passed to the underlying worker as [assets](https://github.com/anomalyco/sst/blob/c36553cf648fa0ec748197bae0ad5035000421a7/platform/src/components/cloudflare/experimental/static-site.ts#L256) where the worker [prepends](https://github.com/anomalyco/sst/blob/c36553cf648fa0ec748197bae0ad5035000421a7/platform/src/components/cloudflare/worker.ts#L510) the project root directory causing the double directory.

Solution appears to be just pass a relative platform path to the worker (or check for absolute paths before prepending), can whip up a PR if needed.

bachiitter:
closes https://github.com/anomalyco/sst/issues/5947
closes https://github.com/anomalyco/sst/issues/6691

## Summary
- Fix Cloudflare StaticSite asset handling for SPA routing
- add a new Cloudflare Vite React SPA example that deploys with `sst.cloudflare.x.StaticSite`
- configure the example for SPA-style routing behavior in Cloudflare static assets
- update experimental Cloudflare StaticSite docs examples to use Cloudflare component references instead of AWS ones


Ernxst:
May fix [6691](https://github.com/anomalyco/sst/issues/6691)

bachiitter:
> May fix [6691](https://github.com/anomalyco/sst/issues/6691)

yes pr has fix for this

platform/src/components/cloudflare/experimental/static-site.ts
```diff
@@ -246,14 +243,16 @@ export class StaticSite extends Component implements Link.Linkable {
           ),
           environment: environment.apply((e) => ({
             ...e,
-            INDEX_PAGE: indexPage,
```
vimtor, line 249: i'm checking the `cf-static-site-router-worker-experimental` function, and it still uses these variables

i guess we should update that router function as well

platform/src/components/cloudflare/experimental/static-site.ts
```diff
@@ -129,7 +139,7 @@ export interface StaticSiteArgs extends BaseStaticSiteArgs {
  * Use [Gatsby](https://www.gatsbyjs.com) to deploy a static site.
  *
  * ```js
- * new sst.aws.StaticSite("MyWeb", {
+ * new sst.cloudflare.x.StaticSite("MyWeb", {
  *   errorPage: "404.html",
```
vimtor: given we are now excluding `errorPage` we should remove it's usage from the jsdocs

vimtor (changes requested):

PR #6569, opened 2026-03-10
Fix/service containers types

JanStevens:
Hi,

I noticed that the types between the Service image and containers where not aligned so this Pr fixes it.

Using Pick so future changes also get automatically added to the containers typings

Regards,


vimtor (approved):
thanks for your contribution @JanStevens 

i brought back the original since the cpu type was different and we lost the container-specific jsdocs

just synced the missing fields

PR #6570, opened 2026-03-10
Undo disallow sst dev on non-personal stages

vimtor:
revert https://github.com/anomalyco/sst/pull/6559 since there are use cases for it (see https://github.com/anomalyco/sst/pull/6559#issuecomment-4031321933)

the main goal was to prevent `sst dev --stage production` from accidentally breaking your state

undoing the change until i come up with a better solution

### DIFF
--- a/cmd/sst/cli/cli.go
+++ b/cmd/sst/cli/cli.go
@@ -352,12 +352,3 @@ func guessStage() string {
 	}
 	return stage
 }
-
-// PersonalStage returns the stored or guessed personal stage for a project.
-func PersonalStage(cfgPath string) string {
-	stage := project.LoadPersonalStage(cfgPath)
-	if stage != "" {
-		return stage
-	}
-	return guessStage()
-}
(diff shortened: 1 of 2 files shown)


kilakewe:
My recommendaton is to go with an opt-in style set up, and warn on most likely scenarios. 

So setting a env &/or configurable sst setting that defines 1 or more stage names that should prevent running sst dev makes sense. 

You could include a console warning that is stage = production then warn the user that they are running what looks like a production env in development. Setting a protected stages value of `[]` would disable this warning. 

The way I'm personally solving right now this is pretty similar to the above. I have a function at the top of my sst config that checks $dev and current stage naem against a list of pre-approved stage names. It's also how I ensure my team are using consistenant personal environments too. ie dev stage must have pattern dev-{developerID} else throw error.

vimtor:
@kilakewe love this idea. what about reusing the existing `protection` field in the `sst.config.ts`?

it should overlap with the stages you don't want to run `sst dev` in 99% of cases

PR #6576, opened 2026-03-11
Compress AWS state files on S3 upload

Animii:
## Summary

Compress state file uploads before storing them in S3.

This change:
- compresses state uploads when `SST_STATE_COMPRESS` is enabled
- keeps reads backward compatible with both plain JSON and already-compressed state objects
## Motivation

Large state files can generate significant S3 upload traffic during deploys because SST writes the state multiple times over the course of a run.

Compressing the state object reduces transfer size substantially while keeping the existing S3 key layout and migration path simple.

## Behavior

- `SST_STATE_COMPRESS` defaults to enabled
- if unset, SST writes compressed state objects with `Content-Encoding: gzip`
- if `SST_STATE_COMPRESS=false` or `0`, SST writes plain JSON state objects
- reads support both formats regardless of the current flag value

This means existing deployments can migrate in place on the next write, and users can safely toggle compression on or off.
I can remove the flag if is not needed or disable it by default.

## Implementation

In `pkg/project/provider/aws.go`:
- `getData()` now detects `Content-Encoding: gzip` and decompresses on read
- `putData()` now prepares either compressed or plain upload bodies depending on `SST_STATE_COMPRESS`
- helper functions were extracted to make the behavior easier to test

## Tests

Added unit tests covering:
- compressed upload preparation
- uncompressed upload preparation
- backward-compatible decoding of both formats
- invalid gzip input rejection

## Benchmark

I benchmarked it on our company project on an already deployed app with ~6900 resources and a ~64 mb state file with a 120 Mbit/s connection.
### Results

**Uncompressed**:
Duration: 4m21s
Total uploads: 14
Toal MB uploaded: 882.08 MB

**Compressed:**
Duration: 4m26s
Total uploads: 50
Total MB uploaded(compressed): 165.18 MB
Total MB uploaded if uncompressed: 3149.43 MB
Compression ratio is ~5%(64 MB -> 3.4 MB) for the state file

We save a lot of bandwidth and because uploads don't take as much time,  SST can upload the state file more frequently.


vimtor:
thanks for your contribution @Animii

have you tried with `PULUMI_DIY_BACKEND_GZIP`? not sure how compatible it is with the current sst setup but it should probably work

Animii:
Yes. With the `PULUMI_DIY_BACKEND_GZIP` the local files are compressed too, but SST assumes non compressed files.
For example:
https://github.com/anomalyco/sst/blob/a856be17c091aa4f7848206861b94f09ec6fe5a1/pkg/project/workdir.go#L119
With the `PULUMI_DIY_BACKEND_GZIP` flag enabled SST can't find these files and crashes.

It would take more invasive work on how SST handles state files to make it compatible(fully delegate state file handling to Pulumi?!).
I think the current approach has a cleaner separation since it only cares about S3 uploads.

Animii:
@vimtor Even if we make SST compatible with the files that Pulumi creates, that would not compress files that SST creates and uploads.

vimtor:
@Animii makes sense, i still need to figure out if we should be using the s3 option pulumi provides for diy backends

not sure that was available when this feature was built into sst

vimtor:
@Animii i'm checking your benchmarks and compressing the file takes longer. why do you need this then? is it to increase the number of uploads after this pr gets merged? to save storage costs?

Animii:
@vimtor I would say it takes more or less the same amount of time. Since SST uploads the files in a separate thread, it does not increase/decrease the deploy time.

I would say the biggest win would be the bandwidth/size reduction and increased number of uploads.
The bandwidth/size reduction would also mean a cost reduction.
image

Its going up and up :) 
💵 💵 💵

Animii:
https://github.com/anomalyco/sst/pull/6669

I guess this would solve at least the cost part

vimtor:
i will hold this pr until https://github.com/anomalyco/sst/pull/6669 gets merged

as described in https://github.com/anomalyco/sst/pull/6669#issuecomment-4178490137, there's an opportutnity to expose a new `state` config like this:

```ts
export default $config({
  app(input) {
    return {
      name: "my-sst-app",
      home: "aws",
      state: {
        retention: 30,
        compress: true
      }
    };
  },
});
```

with both of these we will be able to save a lot of $

vimtor:
going back to this soon

vimtor:
i've made some final changes

@Animii can you verify on your end?

Animii:
> i've made some final changes
> 
> @Animii can you verify on your end?

@vimtor  Will do ASAP

Animii:
@vimtor works as intended :) 
Thank you!

PR #6580, opened 2026-03-13
Update migration notes for v3 by removing Auth

janvdhorst:
Removed the Auth construct from the migration table. Support for this was added a while back (see https://github.com/anomalyco/sst/issues/4893#issuecomment-2867859906)


vimtor:
@janvdhorst i guess we should also write docs on how to migrate the component right? did you update the component in your migration?

PR #6582, opened 2026-03-13
Add `--json` flag to `sst diff` command

### LINKED ISSUE #6547: Add `--json` option to sst commands where applicable
https://www.pulumi.com/docs/iac/cli/commands/pulumi_preview/ for example has a `--json` option

brunorodmoreira:
## Summary

- Adds a `--json` flag to `sst diff` for structured output
- When `--json` is used, the diff result is printed as a JSON array of Pulumi native `StepEventMetadata` to stdout, making it suitable for CI pipelines, scripting, and tooling integration
- UI output is suppressed in JSON mode to keep stdout clean
- Errors from `p.Run()` are still propagated as exit codes, so CI can detect partial failures
- The default text output remains unchanged
- Includes unit tests for JSON rendering

### Example output

```json
[
  {
    "op": "update",
    "urn": "urn:pulumi:dev::app::aws:ecs/service:Service::MyService",
    "type": "aws:ecs/service:Service",
    "old": { "...": "..." },
    "new": { "...": "..." },
    "detailedDiff": {
      "healthCheckGracePeriodSeconds": { "diffKind": "update", "inputDiff": false }
    },
    "provider": "..."
  }
]
```

### Usage

```bash
sst diff --json
sst diff --stage production --json
sst diff --json | jq '[.[] | select(.op == "delete")]'
sst diff --json | jq '[.[] | {urn: .urn, type: .type, op: .op}]'
```

### Local Test
image

## Test plan

- [x] Unit tests for `renderDiffJSON` (empty and with changes)
- [x] `go vet` passes
- [x] Manual e2e test with a deployed stage

Closes #6547


cmd/sst/diff.go
```diff
@@ -90,6 +145,21 @@ var CmdDiff = &cli.Command{
 				Long:  "Run policy pack validation against the preview changes.",
 			},
 		},
+		{
+			Name: "output",
```
vimtor: can we make the flag `--json` instead?

cmd/sst/diff.go
```diff
@@ -3,19 +3,74 @@ package main
...
+}
+
+type DiffChange struct {
+	URN        string               `json:"urn"`
+	Type       string               `json:"type"`
+	Operation  string               `json:"operation"`
+	Properties []DiffPropertyChange `json:"properties,omitempty"`
+}
+
+type DiffPropertyChange struct {
```
vimtor: are these types necessary? i believe the pulumi types already have these annotations

also `pulumi preview` has support for a `--json` flag already. maybe we can just use that

vimtor (changes requested):
thanks for your contribution @brunorodmoreira 

i've requested some changes

cmd/sst/diff.go
brunorodmoreira: I was looking for something more generic to allow for more output types, just like the AWS CLI does (JSON, table and text).

Ref: https://docs.aws.amazon.com/cli/v1/userguide/cli-usage-output-format.html

brunorodmoreira: what do you think @vimtor ?

vimtor: i understand the concern, but table and text would get messy for this data, and the current output is very human-friendly

so the only sensible format left is json, which is perfect for programmatic use as you said, especially now with ai

brunorodmoreira: agreed

brunorodmoreira:
hey @vimtor , could you take a look again 🙏?

vimtor (approved):

PR #6593, opened 2026-03-16
Remove state and encryption key on `sst remove`

### LINKED ISSUE #5617: Support Expiry/Deletion of `sst/passphrase/*` Parameters on Stack Removal
_Yes, an LLM wrote this. You can see the original discussion [here](https://discord.com/channels/983865673656705025/983865673656705028/1342626448183267339)_

**Description:**

In certain environments (e.g. ephemeral or PR deployments), the accumulation of `sst/passphrase/*` parameters becomes unnecessary after stack removal. These parameters are used to encrypt state, but once the stack is removed, retaining them isn't always needed—especially if we don’t intend to recover the state.

**Feature request:**

Provide a way to clean up `sst/passphrase/*` parameters on stack removal. Ideally, this could be done via an opt-out flag in the `home` config, e.g.:

```ts
home: {
  provider: "aws",
  deletePassphraseOnRemove: input.stage !== "prod", // default false
}
```

**Alternative ideas discussed:**

 - A dedicated CLI command (e.g. sst purge --stage pr-1234) to fully clean up a stage, including parameter store entries and files in S3.
 - A simple script to delete the key after a remove, as a manual workaround.

Concerns:
Some caution is required, as automatic deletion may be risky if misconfigured. A user noted the potential danger of accidentally deleting passphrases for the wrong environment.

jamesgibbons92:
Closes #5617 

This PR adds a state config prop  `purgeOnRemove` to the sst config. 
When true, the state file, encryption key and secrets related to the stage are deleted from s3 and ssm.
This replicates the same behaviour as `pulumi destroy --remove` and cleans up unused passphrases and objects in s3. 
This is default false to maintain backwards compatibility.

```
  app(input) {
    return {
      name: "sst-debug-test",
      home: "aws",
      providers: {
        aws: {
          region: "eu-west-1",
        },
      },
      state: {
        purgeOnRemove: true,
      },
    };
  },

```


vimtor:
@jamesgibbons92 is this related to https://github.com/anomalyco/sst/pull/6097?

i like the `--purge` flag approach a lot better

jamesgibbons92:
@vimtor Yeh similar, as part of this PR I am also remove the `app/stage.json` state file.

jamesgibbons92:
There is 1 minor issue I need to look into this PR still, if you run `sst remove --purge` it correctly removes everything (state files, ssm parameter).
However if I run `sst remove` again, I get "Stage not found" which is correct and expected - however this command is creating an SSM passphrase for the stage, even though we aren't deploying a new stage. There seems to be a getOrSet type method being called which is always initialising the passphrase if it doesn't exist, probably in the boostrap process. Need to check if we can change this so it only creates the passphrase if we're deploying a new stack.

vimtor:
related to https://github.com/anomalyco/sst/pull/6669#issuecomment-4178490137 and https://github.com/anomalyco/sst/pull/6576#issuecomment-4178504079

do you think it could make sense to expose this as a config like this:

```ts
export default $config({
  app(input) {
    return {
      name: "my-sst-app",
      home: "aws",
      state: {
        retention: 30,
        compress: true,
        purgeable: true // or a better name 
      }
    };
  },
});
```

jamesgibbons92:
makes sense if were exposing more configuration options for the state management this way, I was going to implement it this way initially but opted for arg as it was easier at the time. happy to update this when that state config handling lands

jamesgibbons92:
@vimtor I added the config for this, it was fairly straightforward. 

one thing this PR does not do is clean up the object versions, maybe this would be covered by #6669 ?

jamesgibbons92:
@vimtor just wondering actually, should we put this config in the `home` prop, kinda makes more sense as this is related to where the state is stored. it would be a union type so you could have either

```
   home: "aws"
```
or
```
  home: {
    "provider": "aws",
    "purgeOnRemove": true
  }
```

vimtor:
i'm thinking maybe we do this:

```
state: {
  store: "aws",
  purgeOnRemove: false,
  // etc...
}
```

and deprecate `home` in the next major

vimtor (approved):

platform/src/config.ts
```diff
@@ -306,6 +306,50 @@ export interface App {
    */
   watch?: string[] | AppWatch;
 
+  /**
+   * Configure how your app's state is managed.
+   *
+   * The state keeps track of all your resources, secrets, and the encryption key. By default,
+   * `sst remove` only deletes the live state files and leaves the noncurrent S3 versions in
+   * place so you can recover the stage if needed. Set `purge: true` to drop everything for
+   * the stage when it is removed.
```
vimtor:
```suggestion
```

PR #6602, opened 2026-03-17
Update VPC component to better accommodate end-user route and rule changes

### LINKED ISSUE #6597: Use ec2.Route instead of inline routes for VPC component
There is a [known limitation](https://github.com/pulumi/pulumi-aws/issues/2561) of the AWS provider that prevents inline routes and `ec2.Route` constructs from working reliably inside a `RouteTable` definition.

This prevents consumers of the `VPC` component from adding routes to the `VPC`'s route tables. We'd like to do this to establish a peering connection between accounts for cross-account databases, where the peering connection allows for ingestion of our application database tables into our data warehouse.

Here's a dummy example:
```typescript
const vpc = new sst.aws.Vpc('VPC', {
  bastion: true,
  nat: 'ec2',
});

vpc.nodes.privateRouteTables.apply((routeTables) => {
  routeTables.map((table, idx) => {
    new aws.ec2.Route(`PeeringRoute${idx}`, {
      routeTableId: table.id,
      vpcPeeringConnectionId: 'pcx-foo',
      destinationCidrBlock: '9.9.9.9/16',
    });
  });
});
```

While this works some of the time, the limitation leads to state sync errors that bring the peering connection offline and force us to manually rectify.

## Request
Use `ec2.Route` instead of inline `routes` in `createPublicSubnets` and `createPrivateSubnets` in the `vpc` component.

weston-harper-rox:
Resolves #6597 by creating routes using the `ec2.Route` construct rather than inline routes for the `VPC` component's public and private subnet creation.

### DIFF
--- a/platform/src/components/aws/vpc.ts
+++ b/platform/src/components/aws/vpc.ts
@@ -1170,20 +1170,17 @@ export class Vpc extends Component implements Link.Linkable {
               `${name}PublicRouteTable${i + 1}`,
               {
                 vpcId: vpc.id,
+                routes: [],
               },
               { parent: self },
             ),
           );
 
-          new ec2.Route(
-            `${name}PublicRoute${i + 1}`,
-            {
-              routeTableId: routeTable.id,
-              destinationCidrBlock: "0.0.0.0/0",
-              gatewayId: internetGateway.id,
-            },
-            { parent: self, deleteBeforeReplace: true },
-          );
+          new ec2.Route(`${name}PublicRoute${i + 1}`, {
+            routeTableId: routeTable.id,
+            destinationCidrBlock: "0.0.0.0/0",
+            gatewayId: internetGateway.id,
+          });
 
           new ec2.RouteTableAssociation(
             `${name}PublicRouteTableAssociation${i + 1}`,
@@ -1226,6 +1223,7 @@ export class Vpc extends Component implements Link.Linkable {
               `${name}PrivateRouteTable${i + 1}`,
               {
                 vpcId: vpc.id,
+                routes: [],
               },
               { parent: self },
             ),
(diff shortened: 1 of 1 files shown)


weston-harper-rox:
Still need to validate deployed component in a real environment.

Edit: Validated that the expected routes are created for the base VPC with an ec2 NAT instance. Still need to create a peering connection or some other route and try to validate that the state sync issue is resolved. Unfortunately, I don't have a reliable repro for that scenario yet. Will start there.

weston-harper-rox:
Attempted to fix #6604, but it seems like state issues persist when working with the default sg, reverting to tighten the scope of this PR

weston-harper-rox:
The code works for new VPCs, but fails for existing ones due to attempts to recreate existing routes. Need to investigate how to reliably replace those routes without drama

weston-harper-rox:
One major issue that still needs resolution:

While this works flawlessly for new VPCs, there needs to be a solution for existing VPCs. Existing VPCs will see a `Route already exists` error where the inline routes are in conflict with the new construct routes. That can be cleared by setting the inline routes to empty arrays. However, since the inline properties are still set, state issues continue. Once the inline routes have been cleared after deploying with an empty array, the inline route property can be effectively removed entirely and everything will work.

Basically, we need a good way to clear existing routes for existing VPCs so that the new routes can take over.

@vimtor, any chance you've dealt with something like this before and can point me in the right direction?

I believe the same issue applies to my attempted resolution of #6604, so we should be able to use whatever technique is applied for the resolution here in the resolution for the other issue.

weston-harper-rox:
From the [docs](https://www.pulumi.com/guides/how-to/aws-ec2-routetable/):

> **How do I remove all managed routes from a route table?**
> Set routes to an empty array (routes: []). Note that omitting the routes argument entirely means ignoring existing routes, not removing them.

It seems like this may need to be a two-step process...

weston-harper-rox:
An alternative may be to allow for altering of the inline routes directly, rather than moving to `ec2.Route`

weston-harper-rox:
Initially, was trying to work around the problem discussed in #6597, but having some success in our own infra using Pulumi transforms:
```typescript
new sst.aws.Vpc(
  'VPC',
  {
    bastion: true,
    nat: 'ec2',
  },
  {
    transforms: [
      (args) => {
        if (args.type === 'aws:ec2/routeTable:RouteTable' && args.name.includes('Private')) {
          return {
            ...args,
            props: {
              ...args.props,
              routes: [
                ...args.props.routes,
                {
                  vpcPeeringConnectionId,
                  cidrBlock,
                },
              ],
            },
          };
        }
      },
    ],
  },
)
```

Given the constraints of migration steps, I think it's unlikely this PR will make sense to merge. Ultimately, I think it's a good idea to move away from the inline definitions where possible, but it's likely going to need to be a `Vpc.V3` sort of change, rather than a change to the existing component. Will close this.

PR #6607, opened 2026-03-18
Disable VCS metadata in Go build

MaTTaDox:
Add -buildvcs=false to the go build invocation in platform/scripts/build for the bridge binary. This prevents embedding VCS metadata into the output,  improving reproducibility of the built artifact.

## Problem
This is another issue related to sst dev.
At the moment, the hash of the build bridge bootstrap binary changes with every commit. As a result, all Lambda functions are treated as changed after every SST release.
https://github.com/anomalyco/sst/pull/6465 already reduced this to a single upload, but the S3 key for every Lambda is still changing, which can still slow deployments down.

## Changes 
- remove vcs information from go build

## Validation
I tested this by switching between commits, with this change the hash stays stable, without it changed every time. 

@vimtor this is the other problem i mentioned in my #6465.


vimtor (approved):
goat @MaTTaDox 

merging this. could you check other places in the codebase where this flag is worth adding too? maybe for the bridge task or go lambdas?

MaTTaDox:
@vimtor better late than never 😅 sorry for the delay — finally checked. The go lambdas and the bridge task were both still stamped: #6976

PR #6624, opened 2026-03-20
feat: Default Bucket Encryption

mkilp:
Added a flag to enable the default server side encryption for the `Bucket` component.

usage: 
```ts
export const encBucket = new sst.aws.Bucket(`encryptionTest`, {
  defaultServerSideEncryption: true,
});
```

Tested working:
CleanShot 2026-03-20 at 10 44 33@2x

### DIFF
--- a/platform/src/components/aws/bucket.ts
+++ b/platform/src/components/aws/bucket.ts
@@ -506,7 +506,7 @@ export interface BucketArgs {
    * @example
    * ```js
    * {
-   *   defaultServerSideEncryption: true
+   *   encryption: true
    * }
    * ```
    */
@@ -900,7 +900,7 @@ export class Bucket extends Component implements Link.Linkable {
 
     const bucket = createBucket();
     createVersioning();
-    createDefaultServerSideEncryption();
+    createEncryption();
     const publicAccessBlock = createPublicAccess();
     const policy = createBucketPolicy();
     createCorsRule();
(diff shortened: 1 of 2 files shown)


mkilp:
For some reason it changed formatting through prettier on the components file, will roll that back and recommit.

platform/src/components/aws/bucket.ts
```diff
@@ -497,6 +497,20 @@ export interface BucketArgs {
...
+   *
+   * @default `false`
+   * @example
+   * ```js
+   * {
+   *   defaultServerSideEncryption: true
+   * }
+   * ```
+   */
+  defaultServerSideEncryption?: Input<boolean>;
```
vimtor:
```suggestion
  encryption?: boolean;
```

sst often prefers elegant single names

also, let's unwrap the `Input` to avoid creating resources inside `.apply`, as a good practice that we'll try to follow onwards: https://github.com/pulumi/agent-skills/blob/main/authoring/skills/pulumi-best-practices/SKILL.md#1-never-create-resources-inside-apply

vimtor (changes requested):

i left a minor refactor ask

it should be good to merge after that

platform/src/components/aws/bucket.ts
mkilp: done.

vimtor:
closing this for now since it's already the default for new s3 uploads. also, i don't want to mislead users into thinking their buckets are unencrypted

maybe i'll come back to this if we release a `Kms` component at some point

PR #6652, opened 2026-03-25
Replace deprecated module syntax with namespace

ikeyan:
[Deprecated: legacy module Syntax for namespaces](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#deprecated:-legacy-module-syntax-for-namespaces)


vimtor (changes requested):
nice catch!

we should update it on the sdk/js as well

ikeyan:
Done.

PR #6653, opened 2026-03-26
Expose route options in Function's `url.router` config

### LINKED ISSUE #5892: can't configure timeout and cors on a lambda route for router
the default timeout on the router is 20s, i tried to increase the timeout via the `router.route` config, like so:

```typescript
const API_TIMEOUT_SECONDS = 120

const foo = new sst.aws.Function("Foo", {
  handler: "pkgs/functions/src/foo/handler.handler",
  timeout: `${API_TIMEOUT_SECONDS} seconds`,
  url: {
    cors: API_CORS_CONFIG, // ℹ️ this gets ignored when using router.route
  }
})

// ℹ️ add this config, with the `readTimeout` set, which works
router.route("/foo", foo.url, {
  readTimeout: `${API_TIMEOUT_SECONDS} seconds`, 
})
```

the issue is, when i do this, the `cors` config is completely ignored and i just get cors errors.

trying to configure the lambda like this, with the router config embedded does not expose any override on the cloudfront timeout:

```typescript
const API_TIMEOUT_SECONDS = 120

const foo = new sst.aws.Function("Foo", {
  handler: "pkgs/functions/src/foo/handler.handler",
  timeout: `${API_TIMEOUT_SECONDS} seconds`,
  url: {
    router: {
      instance: router,
      path: "/foo",
      // ℹ️ can't override readTimeout
    },
    cors: API_CORS_CONFIG,
  }
})
```

julen:
When using `url.router` on `sst.aws.Function`, the component writes CloudFront KV route metadata directly instead of calling `Router.route()`. This meant none of the CloudFront origin options (`readTimeout`, `keepAliveTimeout`, `connectionTimeout`, `connectionAttempts`, `rewrite`) were available through the inline `url.router` interface, forcing users to call `router.route()` separately, which in turn caused CORS settings on the Function URL to be ignored.

This adds these additional options to `RouterRouteArgs` s.t. they are then written into the KV metadata in the same structure that `RouterUrlRoute` produces.

Closes #5892.

### DIFF
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -2805,24 +2805,39 @@ export class Function extends Component implements Link.Linkable {
           {
             store: url.route.routerKvStoreArn,
             namespace: routeNamespace,
-            entries: all([fnUrl.functionUrl, isOac]).apply(
-              ([fnUrlValue, oac]) => ({
-                metadata: JSON.stringify({
-                  host: new URL(fnUrlValue).host,
-                  ...(oac
-                    ? {
-                        origin: {
-                          originAccessControlConfig: {
-                            enabled: true,
-                            signingBehavior: "always",
-                            signingProtocol: "sigv4",
-                            originType: "lambda",
-                          },
-                        },
-                      }
-                    : {}),
-                }),
-              }),
+            entries: all([fnUrl.functionUrl, isOac, url.route]).apply(
+              ([fnUrlValue, oac, route]) => {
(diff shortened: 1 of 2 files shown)


julen:
Full disclosure: I built and tested this against v3.19.0, where I needed the `readTimeout` value to take effect in a function exposed at a router endpoint. I added the rest of the options for completeness, understanding they all work the same way. Then I ported my changes to the dev branch.

Issue #6645, opened 2026-03-24
Add backups support for DSQL


vimtor:

DamienPace15:
Hey @vimtor, I have done some research on this, it's not as straight forward as RDS. 

The pattern is slightly different being a serverless database though. 

Backups for RDS are tightly coupled to the database meaning that it's all wrapped in one, set up a `retentionPeriod` with a few more things and it just works together. 

DSQL you need to create 3 seperate components that are seperate to the actual `DSQL` component. 

1) `Backup vaults` 
2)  `Plan` (basically just a schedule saying back this up every x amount of time) 
3) `Selection` which you map to your cluster/s

You also need to either create this in the other region or do it in the main region and copy across the backup to the region which would get into the territory of having to run some scripts either generated by us or put the onus on the developer OR recreate the above in both regions. 

I would consider wrapping the above 3 as a seperate component and attaching the cluster to that possibly. As they are loosely coupled from DSQL by design.

```
const cluster = new sst.aws.Dsql("MyCluster", {});
const clusterBackup = new sst.aws.DsqlBackup("backups", {lifecycle: *schedule*, link[cluster] })
```

Something along those lines. I need to look into it a bit more but not as straight forward as I thought it was going to be. LMK your thoughts but there are some things to think about. 

Otherwise just put it in the component and replicate across multiple region if required. Probably the path of least resistance tbh as we already have the multi region deployment in there with the witness regions etc 

https://docs.aws.amazon.com/aws-backup/latest/devguide/backup-aurora.html
https://docs.aws.amazon.com/aurora-dsql/latest/userguide/backup-aurora-dsql.html
https://aws.amazon.com/blogs/storage/protect-amazon-aurora-dsql-clusters-using-aws-backup/

vimtor:
we should be more opinionated for this case. you can pass something like `backup: "x days"` and all the resources get created inside. then we just expose transforms for those 3 components you mentioned

would that work? @DamienPace15

DamienPace15:
@vimtor first pass in #6659. 

The days don't really work as it only accepts a cron schedule or a few presets. Putting in a cron was the easiest and also timezone support was added, simple and easy addition that's a nice quality of life. 

Also was pretty straight forward lmaooo

PR #6659, opened 2026-03-28
Add backups support for DSQL 

DamienPace15:
Adds optional backup configuration to `sst.aws.Dsql`. When enabled, creates an AWS Backup vault, plan, selection, and IAM role for the cluster. For multi-region clusters, backup resources are mirrored in the peer region.

```ts
new sst.aws.Dsql("MyCluster", {
  backup: true
});
```

```ts
new sst.aws.Dsql("MyCluster", {
  backup: {
    schedule: "cron(0 2 ? * * *)",
    retention: "90 days"
  }
});
```

Backups are explicitly opt-in. `backup: true` enables the default plan: daily at 5 AM UTC with 7 day retention. Or pass an object to customize the schedule and retention.

Backup resources support transforms via `backupVault`, `backupPlan`, and `backupSelection`.

Also adds `aws:backup/vault:Vault`, `aws:backup/plan:Plan`, and `aws:backup/selection:Selection` to `namingRules` in `component.ts` (50 char limit) so SST auto-naming handles these resource types.

Design reasons
Explicit opt-in made the most sense since AWS Backup is a separate service.

For multi-region clusters, mirroring the backup plan in each region is simpler than copying backups across regions and avoids cross-region transfer costs.

Using AWS Backup cron expressions keeps the API close to what the service natively accepts.

Related to issue #6645

### DIFF
--- a/examples/aws-dsql/sst.config.ts
+++ b/examples/aws-dsql/sst.config.ts
@@ -40,16 +40,16 @@
 export default $config({
   app(input) {
     return {
-      name: 'aws-dsql',
-      removal: input?.stage === 'production' ? 'retain' : 'remove',
-      home: 'aws',
+      name: "aws-dsql",
+      removal: input?.stage === "production" ? "retain" : "remove",
+      home: "aws",
     };
   },
   async run() {
-    const cluster = new sst.aws.Dsql('MyCluster', {});
+    const cluster = new sst.aws.Dsql("MyCluster", {});
 
-    const fn = new sst.aws.Function('MyFunction', {
-      handler: 'lambda.handler',
+    const fn = new sst.aws.Function("MyFunction", {
+      handler: "lambda.handler",
       link: [cluster],
       url: true,
     });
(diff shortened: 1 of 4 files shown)


examples/aws-dsql/sst.config.ts
```diff
@@ -59,4 +59,4 @@ export default $config({
       region: cluster.region,
     };
   },
-});
+});
```
DamienPace15, line 62: Think these are my linters

platform/src/components/aws/dsql.ts
```diff
@@ -119,6 +181,18 @@ export interface DsqlArgs {
      * Transform the EC2 VPC endpoint resource for DSQL connections.
      */
     connectionEndpoint?: Transform<ec2.VpcEndpointArgs>;
+    /**
+     * Transform the AWS Backup vault resource.
+     */
+    backupVault?: Transform<aws.backup.VaultArgs>;
```
DamienPace15: I'm not entirely sure exposing these have real value outside of default settings. No harm in keeping the flexibility but just might be rare someone needs to change them.

PR #6662, opened 2026-03-29
Fix nodejs.install package.json lookup

### LINKED ISSUE #6661: nodejs.install resolves versions from build artifacts instead of the source package.json
When nodejs.install is used, SST resolves package versions from the nearest package.json relative to the build artifact path, not the source handler path.
This can make SST read .sst/artifacts/.../package.json instead of the app's real manifest. If that artifact manifest already contains "*" entries from a previous build, SST reuses those values and installs the wrong package versions.
Why this is a bug
- version resolution should be based on source input, not generated output
- rebuilds can reuse stale artifact package.json values
- this is especially problematic for prerelease/native packages where "*" is not safe
Observed case
Using:
nodejs: {
  install: ["secure-exec", "@secure-exec/v8"],
}
With the app manifest pinning:
{
  "dependencies": {
    "secure-exec": "0.2.0-rc.1"
  }
}
SST generated:
.sst/artifacts/AppletsApi-src/package.json
{
  "dependencies": {
    "secure-exec": "*",
    "@secure-exec/v8": "*"
  }
}
On the next build, SST used that artifact-side package.json as the lookup source, so "*" was preserved instead of reading the pinned source dependency.
Expected
- resolve versions from the nearest source package.json relative to the handler entry file
- do not use .sst/artifacts/.../package.json as the lookup source for nodejs.install
Root cause
In pkg/runtime/node/build.go, the lookup is anchored from target:
fs.FindUp(filepath.Dir(target), "package.json")
That allows the search to walk up from the artifact directory.
It should be anchored from the source file path instead:
fs.FindUp(filepath.Dir(file), "package.json")

bjorntechTobbe:
## Summary
- resolve the `nodejs.install` package manifest lookup from the source handler path instead of the build artifact path
- avoid reading stale `.sst/artifacts/.../package.json` files during rebuilds
- fixes #6661

### DIFF
--- a/pkg/runtime/node/build.go
+++ b/pkg/runtime/node/build.go
@@ -256,7 +256,7 @@ func (r *Runtime) Build(ctx context.Context, input *runtime.BuildInput) (*runtim
 
 		if len(installPackages) > 0 {
 			log.Info("installing", "packages", installPackages)
-			src, err := fs.FindUp(filepath.Dir(target), "package.json")
+			src, err := fs.FindUp(filepath.Dir(file), "package.json")
 			if err != nil {
 				return nil, err
 			}


vimtor:
duplicate of #5988

PR #6665, opened 2026-03-30
Allow overriding SST Bun path

### LINKED ISSUE #4857: Allow override of path to third-party executables such as bun and pulumi
The current implementation of `sst` is installing the `bun` executable in a predefined path and is ignoring executables available in the `$PATH`, which in turn makes `sst` commands that depend on `bun` fail on distributions such as NixOS due to it looking for dynamically linked libraries.

The implemented `NO_BUN` environment variable currently resolves that issue. A better long-term solution would be checking for existing binaries for things such as `bun` on the system before deciding whether to install them through the `sst` tool.

Relevant:
https://github.com/NixOS/nixpkgs/pull/300478

### LINKED ISSUE #6425: bug: SST installs/uses bundled Bun even when Bun is already installed
SST appears to always install/use its own Bun binary, even when Bun is already available on the system.

This causes build commands (for example in `sst.aws.StaticSite`) to run with SST’s bundled Bun (`1.2.x`) instead of the system/project Bun (`1.3.x`), which breaks repos that require Bun 1.3+.

### Regression window
- Works: `v3.17.25`
- Broken: `v3.17.34` (and newer in this range)
- Likely first bad release: `v3.17.30` (based on release/commit history around runtime PATH/tool resolution changes)
- maybe introduced by 4a76d47575f5a280c647194784ee362e7ee4ce67

### Expected behavior
If Bun is already installed and available, SST should use that Bun (or at least not force a downgraded bundled Bun for project build scripts).

### Actual behavior
SST installs/uses its bundled Bun and build scripts run with Bun `1.2.x`.

### Impact
- Build commands fail or behave incorrectly in projects requiring Bun `1.3.x`, e.g. for the `StaticSite` construct
- Tooling version in CI/local is unexpectedly overridden by SST.

### Minimal repro
1. Install Bun `1.3.x`.
2. Use SST (`>=3.17.34`) with a `StaticSite` build command using `bun`.
3. During SST build, observe Bun version resolves to SST-installed `1.2.x` rather than system `1.3.x`.

### Environment
- OS: Linux (CI + local)
- SST: `3.17.34`
- Bun required by project: `1.3.x`

Bug was not present in `3.17.25` and was probably introduces by `3.17.30`, however npm publishing was broken between .25 and .34 so i can't verify this for sure

vimtor:
closes #6425 
closes https://github.com/anomalyco/sst/issues/4857

by allowing users to specify `SST_BUN_PATH` similar to the existing `SST_PULUMI_PATH`


AlvaroParker:
Thanks! :)

PR #6669, opened 2026-03-31
Add configurable S3 lifecycle rules for state bucket

### LINKED ISSUE #4521: Deleting old deployment assets
Hi, I have been building a few things with ion this past week and it has been a wonderful experience. 

Something I noticed looking through my S3 buckets was that there seems to be an `sst-asset-[randomID]` bucket which contains the code uploads for my deployments. However, it looks to me as though it is never deleting previous versions as I can see code archives all the way back to my first test, including many that have been fully overwritten. While the size of this bucket is still relatively manageable, I can see it starting to balloon as I continue to deploy if another 5-10MB is added there with every single update. 

Is there a reason old instances aren't deleted, and is there a CLI command or project setting that could enable automatic deletions of old archived code assets? Would setting up my own lifecycle deletion policy be a good option?

Thanks!

### LINKED ISSUE #6074: Automatically cleanup the state bucket
Hi,

I noticed that the state bucket can easily grow to a couple of GB, especially if you are deploying often and large applications.

Currently what is growing is:

- `/eventlog/<app>/<stage>`
- `/snapshot/<app>/<stage>`
- `/update/<app>/<stage>`

After a couple of weeks with a large app we easily hit 10GB. In comparison with terraform where the state bucket is max a couple of mb.

## Proposals

Following list of possible solutions

1. Let the CLI cleanup automatically so we only keep x amount of past eventlog / snapshot, maybe this is configurable or maybe a sane default of 14 items is more than enough. This can happen after every successful deploy.
2. Add a new CLI command to perform the cleanup. This would then only keep x amount of eventlog / snapshots. Allowing people to take control over it.
3. Add lifecycle rules to the s3 bucket so the above prefixes are cleaned up. The downside here is that if you don't deploy your app in the last 14 days everything will be removed from eventlog / snapshot / update (not sure how mission critical that is)

IMO the most automatic way would be v1. Possible we an env variable or configuration property one would control the x amount of items to keep.

## Workarounds

I've added 2 lifecycle rules for `eventlog` and `snapshot` folders to delete everything after 7 days. Not ideal since I'm not 100% sure if everything would work if I didn't deploy our stack for a week.

Regards,

FelixHeinstadt:
Problem: Over time the s3 storage usage exceeds multiple terabytes which causes increased costs and unnecessary storage usage. There should be an option to set sensible lifecycle rules for the sst state bucket.

## Summary
- Adds env-var-driven S3 lifecycle rules for the SST state bucket
- Master toggle `SST_S3_LIFECYCLE=true` (disabled by default)
- Per-prefix config for expiration days (`SST_S3_LIFECYCLE_{PREFIX}_DAYS`) and noncurrent version retention (`SST_S3_LIFECYCLE_{PREFIX}_VERSIONS`)
- Supports `app/`, `lock/`, `eventlog/`, `snapshot/`, `update/` prefixes
- Runs on every init (not a versioned bootstrap step), so config changes take effect immediately
- Non-fatal: failure logs a warning but doesn't block deploys

## Usage
```bash
export SST_S3_LIFECYCLE=true
export SST_S3_LIFECYCLE_APP_DAYS=7
export SST_S3_LIFECYCLE_APP_VERSIONS=10
export SST_S3_LIFECYCLE_LOCK_DAYS=7
export SST_S3_LIFECYCLE_EVENTLOG_DAYS=7
export SST_S3_LIFECYCLE_SNAPSHOT_DAYS=30
export SST_S3_LIFECYCLE_UPDATE_DAYS=30
```

## Issues

Closes https://github.com/anomalyco/sst/issues/6074
Closes https://github.com/anomalyco/sst/issues/4521

### DIFF
--- a/pkg/project/provider/aws.go
+++ b/pkg/project/provider/aws.go
@@ -9,6 +9,7 @@ import (
 	"log/slog"
 	"os"
 	"path"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -128,9 +129,12 @@ func (a *AwsProvider) Init(app string, stage string, args map[string]interface{}
 	if args["region"] == nil {
 		args["region"] = cfg.Region
 	}
-	_, err = a.Bootstrap(cfg.Region)
+	bootstrap, err := a.Bootstrap(cfg.Region)
 	if err != nil {
 		return err
+	}
+	if err := a.applyLifecycleRules(ctx, cfg, bootstrap); err != nil {
+		slog.Warn("failed to apply lifecycle rules", "error", err)
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


vimtor:
thanks for your contribution @FelixHeinstadt 

this looks very promising

i believe we should surface this as a top level config in `sst.config.ts` instead of env vars:

```ts
export default $config({
  app(input) {
    return {
      name: "my-sst-app",
      home: "aws",
      state: {
        retention: 30, // documented as keep X number of versions
      }
    };
  },
});
```

i've collapsed everything into `x number of versions` since time-based expiration seems scary, and having different options for each part doesn't seem worth it

we'll also need to implement it for the cloudflare home provider. it should be doable since it's s3 compatible

what do you think? do you want to give it a go? i can do the config part if you want to just leave one environment variable for the retention

FelixHeinstadt:
@vimtor If I saw it correctly only /app actually creates new object versions. The others unfortunately just create new objects, so only saying keep x versions would not help... 
But 100% agree using versioned objects and keeping x versions would be a lot better.

FelixHeinstadt:
@vimtor let me know your thoughts on how to proceed.

vimtor:
sorry i need to learn more about this to give you a proper response

my initial intution goes towards making everything a versioned object of finding a way to "when X state version is expired, delete all of the events, snapshots, updates, etc. before it"

vimtor:
i'm merging some other state-related PRs. i'll get back to this soon. my current idea is that we should move to keeping all the state files as versioned files. i need to investigate this more, but i feel like it's worth doing anyway

then we can just add a new `retention` field as I put on my first comment

outaTiME:
While `state.retention` is being implemented, is there any official guidance for users on v4.13.1+ (with `state.compress` now active) on how to safely clean up older noncurrent versions in the state bucket?

Specifically, would applying an S3 `NoncurrentVersionExpiration` lifecycle rule on the state bucket interfere with Pulumi's history or rollback mechanism?

Asking because in our case, snapshot files dropped from ~2.4MB to ~210KB after enabling compress, but the noncurrent uncompressed versions remain and there's no clear interim path to reclaim that storage without risking state integrity.

outaTiME:
Hi @vimtor, following up on this — any updates on the versioned-files + `retention` direction? Happy to test once there's a branch.

In the meantime, I'd really appreciate some interim guidance on safely reclaiming storage in both buckets, since usage keeps growing on every deploy:

**`sst-state` bucket**
- After enabling `state.compress`, current snapshots dropped significantly, but the noncurrent uncompressed versions remain.
- Is it safe to apply an S3 `NoncurrentVersionExpiration` lifecycle rule (e.g. keep last N noncurrent versions of `app/`, `snapshot/`, `update/`, `eventlog/`), or does Pulumi rely on older noncurrent versions for recovery/rollback?
- If not lifecycle-safe, is there a recommended manual cleanup path (CLI command, prefixes safe to prune, etc.)?

**`sst-asset` bucket**
- Each deploy uploads new Lambda zip keys, and old ones accumulate indefinitely.
- For users not relying on Pulumi rollbacks to older state versions, would it be safe to apply a time-based expiration (e.g. objects older than 60–90 days) on this bucket?
- Or are there object prefixes / tag conventions that distinguish "still referenced by current stack" from "orphaned" so we can prune more confidently?

vimtor:
Superseded by #6925, which implements count-based state history retention across AWS, Cloudflare, and local homes.

PR #6689, opened 2026-04-03
Support hyperdrive and worker version metadata bindings

### LINKED ISSUE #6690: Hyperdrive and WorkerVersionMetadata produce invalid declaration files.
Ignoring the type error, Using the following binding on a resource correctly binds it on Cloudflare, but breaks type generation:

```typescript
sst.cloudflare.binding({
    type: "cloudflare.binding",
    binding: "versionMetadata",
    properties: {
      type: "version_metadata", // Hack! The sdk spreads properties onto the final binding so this gets used instead of the one above which doesn't map to anything in SST.
      name: this.#resourceName,
    },
  })
```

Produces (included supported resources for comparison):

```typescript
// cloudflare 
import * as cloudflare from "@cloudflare/workers-types";
declare module "sst" {
  export interface Resource {
    "Api": cloudflare.Service
    "Auth": cloudflare.Service
    "Hyperdrive": cloudflare.
    "MailJobDLQ": cloudflare.Queue
    "MailJobQueue": cloudflare.Queue
    "VersionMetadata": cloudflare.
  }
}
```

Ernxst:
This PR adds hyperdrive and worker version metadata to the supported bindings mappings. (Closes #6690)

I tried to work around this with this wrapper component that I thought was clever:

```typescript
import type { WorkerVersionMetadata } from "@cloudflare/workers-types";

export class VersionMetadata extends $util.ComponentResource {
  #resourceName: string;

  constructor(name: string) {
    super(__pulumiType, name);
    this.#resourceName = name;
  }

  getSSTLink() {
    return {
      // Placeholders for type generation
      properties: {
        id: "",
        tag: "",
        timestamp: "",
      } satisfies WorkerVersionMetadata,
      include: [
        {
          type: "cloudflare.binding",
          binding: "versionMetadata",
          properties: {
            type: "version_metadata",
            name: this.#resourceName,
          },
        },
      ],
    };
  }
}

const __pulumiType = "sst:cloudflare:VersionMetadata";
// @ts-expect-error
VersionMetadata.__pulumiType = __pulumiType;
```

It worked as expected, binding the worker version metadata to the worker, but the type generation produced:

```typescript
// cloudflare 
import * as cloudflare from "@cloudflare/workers-types";
declare module "sst" {
  export interface Resource {
    "Api": cloudflare.Service
    "Hyperdrive": cloudflare.
    "MailJobDLQ": cloudflare.Queue
    "MailJobQueue": cloudflare.Queue
    "VersionMetadata": cloudflare.
  }
}
```

Where the two bindings not supported are generated incomplete as they're not in the mapping. Naming in the mapping can be changed if needed.

### DIFF
--- a/pkg/types/typescript/typescript.go
+++ b/pkg/types/typescript/typescript.go
@@ -13,12 +13,14 @@ import (
 )
 
 var mapping = map[string]string{
-	"aiBindings":          "Ai",
-	"r2BucketBindings":    "R2Bucket",
-	"d1DatabaseBindings":  "D1Database",
-	"kvNamespaceBindings": "KVNamespace",
-	"queueBindings":       "Queue",
-	"serviceBindings":     "Service",
+	"aiBindings":               "Ai",
+	"r2BucketBindings":         "R2Bucket",
+	"d1DatabaseBindings":       "D1Database",
+	"kvNamespaceBindings":      "KVNamespace",
+	"queueBindings":            "Queue",
+	"serviceBindings":          "Service",
+	"hyperdriveBindings":       "Hyperdrive",
+	"versionMetadataBindings":  "WorkerVersionMetadata",
 }
 
 func Generate(root string, links common.Links) error {
--- a/platform/src/components/cloudflare/binding.ts
+++ b/platform/src/components/cloudflare/binding.ts
@@ -67,6 +67,18 @@ export interface D1DatabaseBinding {
   };
 }
 
+export interface HyperdriveBinding {
+  type: "hyperdriveBindings";
+  properties: {
+    id: Input<string>;
+  };
+}
+
+export interface VersionMetadataBinding {
+  type: "versionMetadataBindings";
+  properties: Record<string, never>;
+}
+
 export type Binding =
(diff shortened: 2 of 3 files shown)


bachiitter:
exclude hyperdrive as it requires additional config it better to create a separate component for hyperdrive with config and binding like queue

https://github.com/anomalyco/sst/issues/6683

Ernxst:
I think they can both be components just wrapping pulumi equivalents, here's the same in [alchemy](https://alchemy.run/providers/cloudflare/version-metadata/) - then `sst.cloudflare.binding` can be for referencing resources provisioned outside SST. Might be better to track both hyperdrive and version metadata in that PR you linked @bachiitter

bachiitter:
> I think they can both be components just wrapping pulumi equivalents, here's the same in [alchemy](https://alchemy.run/providers/cloudflare/version-metadata/) - then `sst.cloudflare.binding` can be for referencing resources provisioned outside SST. Might be better to track both hyperdrive and version metadata in that PR you linked @bachiitter 

good point, i should have thought of it

PR #6707, opened 2026-04-09
Fix bridge task in dev mode when `entrypoint/command` is defined

achadha235:
**Problem:** Currently, if you define a task with either `command` and / or `entrypoint` and run `sst dev`, the bridge task definition is created incorrectly with the custom command. This prevents the bridge `main` from being executed and the task does not work. 

Screenshot 2026-04-10 at 12 39 11 AM

**Expected Behavior:** When a bridge task is created in `dev`, we should always get `/app/main` as the command in the task definition so that the tunneling / forwarding functionality works.

**Fix:** The cause was the spread operator in `v[0]` in `/aws/task.ts`. Added code to ensure `command` and `entrypoint` are not passed on in dev mode.

### DIFF
--- a/platform/src/components/aws/task.ts
+++ b/platform/src/components/aws/task.ts
@@ -345,6 +345,8 @@ export class Task extends Component implements Link.Linkable {
             return [
               {
                 ...v[0],
+                entrypoint: undefined,
+                command: undefined,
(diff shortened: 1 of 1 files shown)


vimtor (approved):
thanks for your contribution @achadha235

PR #6721, opened 2026-04-11
Add Lambda rollout with CodeDeploy traffic shifting, smoke tests, and alarm-based rollbacks

### LINKED ISSUE #6719: feat(aws-lambda): CodeDeploy-managed traffic shifting with smoke tests, canary/linear deployments, and alarm-based rollbacks
## Summary

Add support for controlled traffic shifting when deploying Lambda functions using AWS CodeDeploy. This gives developers the ability to:

- **Validate new versions before any traffic reaches them** — run a before-traffic hook that invokes the new version and verifies it works before CodeDeploy shifts traffic
- **Gradually shift traffic** — canary (send a percentage for a duration, then 100%) or linear (increment by a percentage at regular intervals)
- **Automatically roll back on failure** — attach CloudWatch alarms that trigger a rollback if error rates or latency spike during the rollout

## API

```ts
const fn = new sst.aws.Function("Api", {
  handler: "src/api.handler",
  url: true,
  rollout: {
    type: "canary",       // or "linear" or "all-at-once"
    percentage: 10,
    duration: "10 minutes",
    alarms: [errorAlarm.name],
    beforeTraffic: "src/before-traffic.handler",
  },
});
```

The before-traffic hook uses a new SDK:

```ts
import { rollout } from "sst/aws/rollout";

export const handler = rollout.handler(async (event) => {
  const resp = await fetch(Resource.Function.latestUrl);
  await rollout.report(event, resp.ok ? "Succeeded" : "Failed");
});
```

## Features

- **Three deployment strategies**: `all-at-once`, `canary`, `linear`
- **Before/after traffic hooks**: validate new versions before shifting traffic, run post-deployment checks after
- **CloudWatch alarm integration**: automatic rollback if alarms fire during deployment
- **SNS notifications**: subscribe to deployment lifecycle events (start, success, failure, rollback)
- **Conflict handling**: configure what happens when a new deploy starts while a rollout is in progress (`cancel`, `rollback`, `fail`)
- **Function URL support**: `url` points to the stable alias, `latestUrl` exposes the latest version for testing
- **Works with API Gateway, Router, and other event sources**: pass the function directly or use `fn.targetArn` to route to the stable version
- **SST SDK for lifecycle hooks**: `rollout.handler()` for typed events, `rollout.report()` to report status — no need to use the CodeDeploy SDK directly
- **`addRollout()` for deferred configuration**: create the alias upfront, configure the deployment strategy later when you need to reference the function in hook functions

### LINKED ISSUE #6720: Function URL should point to the deployed version when versioning is enabled
## Problem

When `versioning: true` is set on a function, the function URL still points to `$LATEST`. This means every invocation via the URL runs whatever code `$LATEST` currently holds.

## Expected behavior

At minimum, the function URL should be pinned to a specific published version.

Ideally, the function URL should point to an alias. This setup allows you to roll back the version on the alias without modifying the Function URL configuration or the Lambda permission configuration. If the URL points to a specific version, you must update it to make the older version the primary target. With an alias, this is not an issue. You only need to update the version that the alias points to, and requests will start reaching the older version.

## Related

This is part of a broader issue: when versioning is enabled, all event sources (not just function URLs) should invoke through an alias rather than `$LATEST` or a bare version number. See https://github.com/anomalyco/sst/issues/6722

### LINKED ISSUE #6722: Event sources should invoke Lambda through an alias when versioning is enabled
## Problem

When `versioning: true` is set, event sources (API Gateway, DynamoDB streams, SQS, S3, SNS, etc.) are configured to invoke a specific published version, and the `lambda.Permission` is scoped to that exact version. This creates several issues:

### Rolling back is a mess

The primary point of versioning is to enable rollback. But rolling back to a previous version currently requires updating every event source's target version *and* every `lambda.Permission` to allow the old version. There's no simple "point everything back to v5" operation.

### Propagation delays cause downtime

Resources like API Gateway have their own propagation times. When a new version is deployed, updating the Lambda version in the API Gateway integration takes time to propagate. With an alias, API Gateway doesn't need to change at all — it always points to the alias, and the moment the alias is updated, traffic shifts instantly.

### Permission gaps during deploys

The `lambda.Permission` is scoped to a specific version number. If the permission update and the event source update don't happen at exactly the same time, there's a window where the event source is trying to invoke a version it doesn't have permission for, or the permission is granted for a version that isn't wired up yet. This can cause brief downtime on every deploy.

## Solution

When `versioning: true`, create an alias (e.g. `latest`) that always points to the most recently published version. All event sources should:

1. **Invoke the alias** instead of a hard-coded version number
2. **Scope `lambda.Permission` to the alias** instead of a specific version

This way, deploying a new version only requires updating the alias pointer. Event sources don't change, permissions don't change, and rollback is a single alias update.

## Related

- #6720 — Function URL should point to the deployed version/alias when versioning is enabled

---

anatolzak:
closes https://github.com/anomalyco/sst/issues/6719
closes https://github.com/anomalyco/sst/issues/6720
closes https://github.com/anomalyco/sst/issues/6722

## Summary

Adds `rollout` support to `sst.aws.Function` using AWS CodeDeploy for controlled traffic shifting on deploy. This enables smoke testing new versions before traffic reaches them, gradual canary/linear deployments, and automatic rollback on CloudWatch alarm triggers.

## Demos

- **Smoke Test — Function URL** — 

https://github.com/user-attachments/assets/906efd2d-2e4b-45cb-9e46-9802f51bc4a2

- **Smoke Test — HTTP API** — 

https://github.com/user-attachments/assets/21ba5e60-957f-4ae2-9686-be49c13021bd

- **Smoke Test — Router** — 

https://github.com/user-attachments/assets/91d50739-2a96-4588-884e-5dfbf37c1cf7

- **Smoke Test — Direct Lambda Invoke Calls** — 

https://github.com/user-attachments/assets/d61ff73f-8386-4426-b642-80aa56c3fae9

- **Lambda Rollout** (canary with alarms) — 

https://github.com/user-attachments/assets/f8c6a137-c5ab-402f-9bc0-8321974b3e0d

### New components and SDK
- **`FunctionRollout`** component (`function-rollout.ts`) — encapsulates all CodeDeploy resources (application, deployment group, deployment config, IAM role, deployment, waiter)
- **`CodeDeployLambdaDeployment`** dynamic provider (`codedeploy-lambda-deployment.ts`) — creates CodeDeploy deployments via RPC
- **`CodeDeployDeploymentWaiter`** dynamic provider (`codedeploy-deployment-waiter.ts`) — polls deployment status
- **`rollout` SDK** (`sdk/js/src/aws/rollout.ts`) — `rollout.handler()` for typed lifecycle hook events, `rollout.report()` to report status back to CodeDeploy without needing the AWS SDK

### Function component changes (`function.ts`)
- **New `rollout` property** on `FunctionArgs` — supports `all-at-once`, `canary`, and `linear` strategies with before/after traffic hooks, alarms, SNS notifications, and conflict handling
- **`addRollout()` method** — deferred rollout configuration for when hook functions need to reference the parent function
- **`latestUrl`** — optional function URL pointing to the latest published version (via alias), separate from the stable `url`
- **`targetArn` / `latestTargetArn`** — getters that resolve to the correct alias ARN for event source subscriptions
- **`latestQualifier`** — exposed in `getSSTLink` for invoking the latest version via the Lambda SDK
- **Rollout transforms** moved to `FunctionArgs.transform.rollout` (not inside `rollout.transform`)
- **`latestAlias`** exposed in `nodes` getter

### Event source compatibility
- **All subscriber/route types** (`Queue`, `Bucket`, `SnsTopic`, `Dynamo`, `KinesisStream`, `ApiGatewayV2`, `ApiGatewayV1`, `Realtime`, `Cron`, `CognitoUserPool`, etc.) now accept `Function` instances in their type signatures. By default the components will use the stable alias by using the `targetArn`.
- **`lambda.Permission`** — added `qualifier` to all permission resources across the codebase, including `ssr-site.ts`
- **`Realtime`** — migrated from `Function.fromDefinition` to `functionBuilder`, uses `targetArn` for IoT authorizer

### Go changes
- **`aws-codedeploy.go`** — shared utilities with struct-based inputs: `handleDeploymentConflict`, `stopDeployment`, `findActiveDeployment` (with pagination), `createDeployment`, `buildAppSpec`
- **`aws-codedeploy-lambda-deployment.go`** — Lambda-specific deployment handler that builds AppSpec and manages version diffing

### Examples
- **`aws-lambda-rollout`** — full canary rollout with CloudWatch alarms and SNS notifications
- **`aws-lambda-smoke-test`** — before-traffic hook using Lambda SDK invoke
- **`aws-lambda-smoke-test-function-url`** — before-traffic hook using function URL
- **`aws-lambda-smoke-test-http-api`** — smoke test with API Gateway HTTP API
- **`aws-lambda-smoke-test-router`** — smoke test with SST Router

## Testing locally

The `sst/aws/rollout` SDK is new and not yet published. To test examples locally:

```bash
bun run setup
bun run build:platform
cd sdk/js && bun run build

# In each example directory:
bun install
rm -rf node_modules/sst && ln -s ../../../sdk/js node_modules/sst
go run ../../cmd/sst install
go run ../../cmd/sst deploy
```

## Notes

- Rollout only triggers when function code changes between deploys. During `sst dev`, the deployed code is a stub that never changes, so rollout is effectively a no-op. Examples set `dev: false` to demonstrate this.
- When connecting a function with rollout to event sources, pass the function directly or use `fn.targetArn`. Using `fn.arn` bypasses rollout entirely.

## Note on failed deployments

If a deployment fails (e.g., the before-traffic hook reports failure or an alarm triggers a rollback assuming `wait: true`), the `sst deploy` process will error out with a message pointing to the CodeDeploy deployment in the AWS console.

If you re-run `sst deploy` without changing any function code, the deployment waiter will detect the deployment ID hasn't changed and skip the wait — so the deploy will succeed cleanly without re-triggering the failed deployment. Once you push a code fix and deploy again, a new CodeDeploy deployment is created and validated normally.

### DIFF
--- a/examples/aws-lambda-rollout/infra/alarms.ts
+++ b/examples/aws-lambda-rollout/infra/alarms.ts
@@ -1,0 +1,79 @@
+export function createCanaryAlarms(
+  name: string,
+  opts: {
+    fn: sst.aws.Function;
+    alertsTopic: sst.aws.SnsTopic;
+  },
+) {
+  const { fn, alertsTopic } = opts;
+
+  // These alarms track a specific Lambda version via the ExecutedVersion
+  // dimension. When we deploy a new version, we replace the alarm so it starts
+  // fresh — no leftover state from the previous version. The old alarm is
+  // deleted after the new one is created so the deployment is not interrupted
+  // by the removal of the alarm mid-deploy.
+  const pulumiOpts: $util.CustomResourceOptions = {
+    replaceOnChanges: ["dimensions"],
+    deleteBeforeReplace: false,
+  };
+
+  // Triggers if any errors occur in the deployed version within a 5-minute window.
+  const errorAlarm = new aws.cloudwatch.MetricAlarm(
+    `${name}ErrorAlarm`,
+    {
+      alarmActions: [alertsTopic.arn],
+      namespace: "AWS/Lambda",
+      metricName: "Errors",
+      dimensions: {
+        FunctionName: fn.name,
+        Resource: getFunctionResource(fn.targetArn),
+        ExecutedVersion: fn.nodes.function.version,
+      },
+      statistic: "Sum",
+      period: 300,
+      evaluationPeriods: 1,
+      threshold: 1,
+      comparisonOperator: "GreaterThanOrEqualToThreshold",
+      treatMissingData: "notBreaching",
+    },
+    pulumiOpts,
+  );
+
+  // Triggers if average latency exceeds 2 seconds in a 5-minute window.
+  const latencyAlarm = new aws.cloudwatch.MetricAlarm(
+    `${name}LatencyAlarm`,
+    {
+      alarmActions: [alertsTopic.arn],
+      namespace: "AWS/Lambda",
+      metricName: "Duration",
+      dimensions: {
+        FunctionName: fn.name,
+        Resource: getFunctionResource(fn.targetArn),
+        ExecutedVersion: fn.nodes.function.version,
+      },
+      statistic: "Average",
+      period: 300,
+      evaluationPeriods: 1,
+      threshold: 2000,
+      comparisonOperator: "GreaterThanOrEqualToThreshold",
+      treatMissingData: "notBreaching",
+    },
+    pulumiOpts,
+  );
+
+  return { errorAlarm, latencyAlarm };
+}
+
+/**
+ * Extracts the `FunctionName:Alias` resource identifier from a Lambda alias ARN.
+ * CloudWatch metrics use this format for the `Resource` dimension when tracking
+ * a specific alias.
+ *
+ * For example, given `arn:aws:lambda:us-east-1:123456789:function:my-fn:live`,
+ * this returns `my-fn:live`.
+ */
+function getFunctionResource(targetArn: $util.Input<string>) {
+  return aws
+    .getArnOutput({ arn: targetArn })
+    .resource.apply((r) => r.split(":").slice(1).join(":"));
+}
--- a/examples/aws-lambda-rollout/infra/topics.ts
+++ b/examples/aws-lambda-rollout/infra/topics.ts
@@ -1,0 +1,36 @@
+export function createTopics(opts?: { email?: string }) {
+  const alertsTopic = new sst.aws.SnsTopic("Alerts");
+
+  allowCloudWatchPublish("Alerts", alertsTopic);
+
+  if (opts?.email) {
+    subscribeEmail("Alerts", alertsTopic, opts.email);
+  }
+
+  return { alertsTopic };
+}
+
+function subscribeEmail(name: string, topic: sst.aws.SnsTopic, email: string) {
+  new aws.sns.TopicSubscription(`${name}Email`, {
+    topic: topic.arn,
+    protocol: "email",
+    endpoint: email,
+  });
+}
+
+function allowCloudWatchPublish(name: string, topic: sst.aws.SnsTopic) {
+  new aws.sns.TopicPolicy(`${name}CloudWatchPolicy`, {
+    arn: topic.arn,
+    policy: aws.iam.getPolicyDocumentOutput({
+      statements: [
+        {
+          actions: ["sns:Publish"],
+          principals: [
+            { type: "Service", identifiers: ["cloudwatch.amazonaws.com"] },
+          ],
+          resources: [topic.arn],
+        },
+      ],
+    }).json,
+  });
+}
--- a/examples/aws-lambda-rollout/package.json
+++ b/examples/aws-lambda-rollout/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "aws-lambda-rollout",
+  "private": true,
+  "type": "module"
+}
--- a/examples/aws-lambda-rollout/scripts/test.ts
+++ b/examples/aws-lambda-rollout/scripts/test.ts
@@ -1,0 +1,64 @@
+interface Result {
+  status: number;
+  latency: number;
+  version: string;
+  error: boolean;
+}
+
+export async function loadTest(url: string, count = 50) {
+  console.log(`URL: ${url}\n`);
+  console.log(`Sending ${count} concurrent requests...\n`);
+
+  const results: Result[] = await Promise.all(
+    Array.from({ length: count }, async () => {
+      const start = Date.now();
+      try {
+        const res = await fetch(url);
+        const body = (await res.json()) as { version?: string };
+        return {
+          status: res.status,
+          latency: Date.now() - start,
+          version: body.version ?? "unknown",
+          error: false,
+        };
+      } catch {
+        return {
+          status: 0,
+          latency: Date.now() - start,
+          version: "unknown",
+          error: true,
+        };
+      }
+    }),
+  );
+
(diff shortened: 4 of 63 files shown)


anatolzak:
Hey @vimtor! I noticed you recently added the `targetArn` getter to the Function component so all event sources use the function version directly, that was a huge help for the rollout work I just put up.

The PR adds full CodeDeploy-managed rollout support to `sst.aws.Function`, canary, linear, and all-at-once deployment strategies with before/after traffic hooks for smoke testing, CloudWatch alarm-based automatic rollbacks, SNS notifications, and a new `sst/aws/rollout` for the before/after traffic hook status reporting SDK allows developers to avoid any interaction with the CodeDeploy SDK.

With rollout enabled, `targetArn` resolves to the stable alias ARN managed by CodeDeploy, so all event sources automatically invoke the validated version. 

In the future, we could also support ECS services through CodeDeploy as well.

vimtor:
thanks for your contribution @anatolzak 

this looks fantastic, your pull requests are always top tier

i'm a bit worried about the scope for this one. the function component it's already very complex

i'm wondering what the common use case is here. probably doing a gradual rollout of the new function code? i'm not convinced the other things (sdk, custom alarms or function hooks)

i'm not super familiar with codedeploy but my guess is that in 99% of cases people just want to ensure that the new lamba doesn't error

what do you think?

anatolzak:
Hey @vimtor, thanks for the kind words, and you were totally right about the scope concern, I completely get it. Trying to land everything in one PR made it hard to evaluate the pieces on their own merit.

Zooming out, what I'm really trying to do is make SST more production-ready around versioning/rollouts/rollbacks. I think that splits cleanly into two PRs:

**1. Make `versioning: true` actually do the right thing.** Right now, when versioning is enabled, SST publishes immutable versions, but event sources and their Lambda permissions are wired to a specific version ARN. That means every time you publish a new version, you have to update every event source to point at the new version and update every permission alongside it. Rollback has the same problem in reverse: find every event source, repoint it at the previous version, update the permission, redeploy. It's exactly the operational pain that versioning is supposed to solve. I created an issue for this https://github.com/anomalyco/sst/issues/6722

The fix is to always create a stable alias when versioning is enabled, point the function URL at the alias, and have event sources/permissions use the alias's ARN and qualifier instead of a specific version. Once that's in place, rollback is trivial, open the Lambda console, flip the alias to the previous version, and every event source picks it up instantly. This is standard AWS best practice for versioned functions, and it's valuable on its own for anyone using `versioning: true` today, independent of any rollout machinery.

**2. Then revisit rollout.** Once versioning is solid, we have a clean foundation to talk about what a rollout feature should actually look like. Happy to discuss that separately once the first piece is merged.

I've just opened https://github.com/anomalyco/sst/pull/6767 for the first part. It's much smaller and self-contained. What do you think about this framing, and does the PR itself look like the right direction before I rework the rollout one on top?

israellev:
It so crazy feature!! I love it!
This will allow us to actually guarantee our clients 99.999 uptime without any wrong release that can destroy everything

Thanks @anatolzak for that

kevbot-git:
Grateful this PR exists! Any update on this?
Agree it would make rollouts much less painful

PR #6731, opened 2026-04-14
Respect autostart in `sst dev --mode=mono`

OliverGilan:
Keep mono-mode dev processes registered without starting them when autostart is disabled, and only start them once a later update enables autostart. Add monoplexer coverage for disabled startup, later enablement, and config change handling.

---
## Problem
We have multiple components in our monorepo that SST manages and deploys but we do not need to run many of them in dev mode. Luckily sst already has a property called `autostart` for many components and `DevCommand` which prevents these components from running in dev mode unless manually triggered.

One of our engineers runs `--mode=mono` on `sst dev` which currently doesn't respect these `autostart` configs. This PR updates the dev command so it does respect `autostart` and doesn't run components in dev mode when unnecessary.

### DIFF
--- a/cmd/sst/mosaic.go
+++ b/cmd/sst/mosaic.go
@@ -430,8 +430,8 @@ func CmdMosaic(c *cli.Cli) error {
 			fnTitle = "Worker"
 			fnFilter = "worker"
 		}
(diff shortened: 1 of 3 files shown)


vimtor (approved):
thanks for your contribution @OliverGilan

PR #6732, opened 2026-04-14
Allow skipping directories during type generation

OliverGilan:
Add an app-level typeIgnore config and thread it into SST's project discovery so JavaScript and Python projects can opt out of generated type files.

--
## Problem
In a monorepo with many packages, a lot of them have nothing to do with SST yet SST insists on generating its types file in every directory where it finds a package.json or pyproject file. 

## Solution
This PR adds a simple config to ignore certain directories for type generation files

### DIFF
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -43,7 +43,22 @@ func IsGitSubmodule(dir string) bool {
 }
 
 func FindDown(dir, filename string) []string {
+	return FindDownWithIgnore(dir, filename, nil)
+}
+
+func FindDownWithIgnore(dir, filename string, ignore []string) []string {
 	var result []string
+	ignored := make([]string, 0, len(ignore))
+	for _, item := range ignore {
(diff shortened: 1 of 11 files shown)


vimtor:
thanks for your contribution @OliverGilan 

could you take a look at https://github.com/anomalyco/sst/pull/6675? maybe we could reuse its existing `watcher.ignore` (although watcher makes less sense in this context)

OliverGilan:
@vimtor I thought about doing that, it would definitely be cleaner than introducing a whole new config setting but I believe what I ran into is that the watcher doesn't run in all contexts where typegen happens. For example sst deploy will generate types but I don't believe there's any watcher in that context. 

I think for our purposes that's not a huge deal because we only deploy in CI so we could go that route if you prefer

vimtor:
closing in favor of https://github.com/anomalyco/sst/pull/6796 (couldn't push to your fork)

PR #6737, opened 2026-04-14
ApiGateway: allow naming routes via `name` arg

### LINKED ISSUE #6736: Allow naming routes added via `api.route()`
## Problem

Routes added via `api.route()` on `ApiGatewayV2`, `ApiGatewayV1`, and
`ApiGatewayWebSocket` get auto-generated component names like
`MyApiRouteAbc123Handler`, where the suffix is a 6-char hash of the
route path. That makes the resulting Lambda function, log group, and
Pulumi resource opaque when:

- Reading CloudWatch logs (which log group belongs to which route?)
- Scanning `pulumi state` or `sst diff` output
- Watching the `sst dev` TUI (the function ID shown for each invoke is
  the hash, not the route)

## Today's workarounds (and why they fall short)

1. **Define a standalone `sst.aws.Function` and pass its ARN**:

   ```ts
   const fn = new sst.aws.Function("GetUser", { handler: "..." });
   api.route("GET /users/{id}", fn.arn);
   ```

   Loses the inline ergonomics: `handlerLink` from the API doesn't
   merge in, `handlerTransform` doesn't apply, the auto-generated
   `description` interpolation is lost, and you have to manually wire
   the function's `link` to include any API-level resources.

2. **`transform.function.name`**: only renames the *physical* AWS Lambda;
   the Pulumi URN, log group, and TUI ID still show the hash.

## Proposed solution

Add an optional `name?: string` to the route args interface for each
API Gateway component. When set, it replaces the hashed suffix used to
derive the route's Pulumi component name:

```ts
api.route("GET /users/{id}", "src/users.get", { name: "GetUser" });
// → resource name becomes MyApiRouteGetUserHandler
```

This mirrors the existing `args.name` pattern used by
`ApiGatewayV1.addAuthorizer()` (apigatewayv1.ts:1181), so the
convention already exists in the codebase.

Backward compatible — routes without `name` keep the existing
hash-based naming.

## Notes

- `name` is a logical (Pulumi) name; the physical AWS Lambda name still
  goes through `physicalName()` and gets the `<app>-<stage>-` prefix, so
  the same `name` is safe across stages.
- Two routes on the same API in the same stage with the same `name`
  collide with a Pulumi duplicate-URN error, like any other resource.

Happy to send a PR — drafted one across all three API Gateway components
already.

the-human-mustachio:
Closes #6736

## Motivation

Today, routes added via `api.route()` get auto-generated names like
`MyApiRouteAbc123Handler` (a 6-char hash of the route path). That suffix
is opaque when reading CloudWatch logs, scanning `pulumi state`, or
watching the `sst dev` TUI — you can't tell which route a function
belongs to without looking it up.

The only existing escape is to define a separate `sst.aws.Function` and
pass its ARN to `api.route()`, which loses the inline ergonomics
(`handlerLink`, `handlerTransform`, the auto `description` interpolation).
`transform.function.name` only renames the physical AWS Lambda — the
Pulumi URN, log group, and TUI ID still show the hash.

## Change

Add an optional `name` field to the route args of all three API Gateway
components (HTTP, REST, WebSocket). When set, it replaces the hashed
suffix used for the route's Pulumi component name:

```ts
api.route("GET /users/{id}", "src/users.get", { name: "GetUser" });
// → MyApiRouteGetUserHandler
```

Mirrors the existing `args.name` pattern used by
`ApiGatewayV1.addAuthorizer()` (apigatewayv1.ts:1181). Backward
compatible — routes without `name` keep the existing hash-based naming.

## Files

- `platform/src/components/aws/apigatewayv2.ts`
- `platform/src/components/aws/apigatewayv1.ts`
- `platform/src/components/aws/apigateway-websocket.ts`

## Notes

- `name` is a logical (Pulumi) name — the physical AWS Lambda name still
  goes through `physicalName()` and gets the standard `<app>-<stage>-`
  prefix, so the same `name` is safe across stages.
- Two routes on the same API in the same stage with the same `name`
  collide with a Pulumi duplicate-URN error, like any other resource.

## Test plan

- [x] `bun run typecheck` clean
- [ ] Maintainer to verify deploy of an example app with `name` set

### DIFF
--- a/platform/src/components/aws/apigateway-websocket.ts
+++ b/platform/src/components/aws/apigateway-websocket.ts
@@ -343,13 +343,12 @@ export interface ApiGatewayWebSocketRouteArgs {
     }
   >;
   /**
-   * The name of the route. By default, SST generates a unique suffix from a hash of the
-   * route name (except for the special `$connect`, `$disconnect`, and `$default` routes,
-   * which always use their literal name). Setting `name` replaces the hashed suffix so
-   * the function becomes `<name>Handler`, giving you a stable, meaningful identifier in
-   * Pulumi state, the CloudWatch log group, and the `sst dev` TUI.
+   * The name of the route.
    *
-   * Must be unique across all routes on the same API.
+   * By default, SST generates a unique suffix from the route path. Setting `name` gives
+   * you a stable, human-readable name like `MyApiRouteSendMessageHandler`.
+   *
+   * Must be unique across all routes.
    *
    * @example
    * ```js
--- a/platform/src/components/aws/apigateway-websocket.ts
+++ b/platform/src/components/aws/apigateway-websocket.ts
@@ -343,6 +343,23 @@ export interface ApiGatewayWebSocketRouteArgs {
     }
   >;
   /**
+   * The name of the route. By default, SST generates a unique suffix from a hash of the
+   * route name (except for the special `$connect`, `$disconnect`, and `$default` routes,
+   * which always use their literal name). Setting `name` replaces the hashed suffix so
+   * the function becomes `<name>Handler`, giving you a stable, meaningful identifier in
+   * Pulumi state, the CloudWatch log group, and the `sst dev` TUI.
+   *
+   * Must be unique across all routes on the same API.
+   *
+   * @example
+   * ```js
+   * {
+   *   name: "SendMessage"
+   * }
+   * ```
+   */
+  name?: string;
+  /**
    * [Transform](/docs/components#transform) how this component creates its underlying
    * resources.
    */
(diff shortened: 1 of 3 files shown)


vimtor (approved):
thanks for your contribution @the-human-mustachio

outaTiME:
Hey! Just tried out the new `name` property on an existing API with ~15 routes and ran into some friction during deploy.

When adding `name` to existing routes, Pulumi tries to create the new named route before deleting the old hashed one. Since API Gateway doesn't allow duplicate route keys (e.g. `GET /fetch`), it fails with `ConflictException: Route with key GET /fetch already exists for this API`.

I tried setting `deleteBeforeReplace: true` via `transform.route.args` but it doesn't propagate to the underlying `aws.apigatewayv2.Route` resource. Ended up having to manually delete the conflicting routes via AWS CLI to unblock the deploy.

Sharing this in case anyone else runs into the same issue, and also curious if there's a better way to handle the migration. Would be great if this could be handled internally — maybe setting `deleteBeforeReplace` on the Route resource when a `name` is provided, so existing APIs can adopt route naming without manual intervention.

Thanks for the feature though, much nicer Lambda names in CloudWatch now! 🙌

the-human-mustachio:
@outaTiME, thanks for that feedback, let me give it a look and see what fits.

PR #6743, opened 2026-04-15
fix(router): remove originAccessControlConfig for HTTP origins to prevent 502

### LINKED ISSUE #6742: Router: router.route() with http:// URL origin returns 502 (Failed to contact the origin)
## Bug

`router.route("/*", "http://<alb-dns>")` results in a **502 "Failed to contact the origin"** because CloudFront attempts HTTPS:443 instead of HTTP:80.

## Reproduction

```ts
const router = new sst.aws.Router("MyRouter", { domain: "example.com" });
const service = new sst.aws.Service("MyService", {
  cluster,
  loadBalancer: {
    rules: [{ listen: "80/http", forward: "8080/http" }],
  },
});

// Pass the ALB's raw DNS as an http:// URL
const albUrl = service.nodes.loadBalancer.dnsName.apply(
  (dns) => `http://${dns}`
);
router.route("/*", albUrl);
```

The ALB only listens on port 80 (HTTP). CloudFront returns 502 because it tries to connect on HTTPS:443.

## Root cause

In the generated CloudFront Function, `setUrlOrigin` builds the origin object like this:

```javascript
function setUrlOrigin(urlHost, override) {
  const origin = {
    domainName: urlHost,
    customOriginConfig: { port: 443, protocol: "https", sslProtocols: ["TLSv1.2"] },
    originAccessControlConfig: { enabled: false },
  };
  override = override ?? {};
  if (override.protocol === "http") {
    delete origin.customOriginConfig;
  }
  // ...
  cf.updateRequestOrigin(origin);
}
```

When `protocol === "http"`, `customOriginConfig` is deleted. The expectation is that `cf.updateRequestOrigin` inherits the placeholder origin's `customOriginConfig` (which is `http-only` port 80). However, the presence of `originAccessControlConfig: { enabled: false }` appears to prevent this inheritance — CloudFront defaults to HTTPS:443 instead.

The KV store metadata confirms the route is correctly stored:

```json
{
  "host": "my-alb-1234567890.us-east-2.elb.amazonaws.com",
  "origin": { "protocol": "http" }
}
```

And the placeholder origin is correctly configured:

```json
{
  "CustomOriginConfig": {
    "HTTPPort": 80,
    "OriginProtocolPolicy": "http-only"
  }
}
```

But the ALB gets an HTTPS connection attempt on 443 → connection refused → 502.

## Suggested fix

Don't include `originAccessControlConfig` for non-OAC origins. For HTTP custom origins it serves no purpose and interferes with `customOriginConfig` inheritance:

```javascript
function setUrlOrigin(urlHost, override) {
  const origin = { domainName: urlHost };

  if (!override || override.protocol !== "http") {
    origin.customOriginConfig = { port: 443, protocol: "https", sslProtocols: ["TLSv1.2"] };
  }

  if (override?.connectionAttempts) origin.connectionAttempts = override.connectionAttempts;
  if (override?.timeouts) origin.timeouts = override.timeouts;
  if (override?.originAccessControlConfig) {
    origin.originAccessControlConfig = override.originAccessControlConfig;
  }

  cf.updateRequestOrigin(origin);
}
```

## Workaround

Monkeypatch `cf.updateRequestOrigin` via `edge.viewerRequest.injection` to strip the empty `originAccessControlConfig` before it reaches CloudFront:

```ts
new sst.aws.Router("MyRouter", {
  domain: "example.com",
  edge: {
    viewerRequest: {
      injection: `
const _update = cf.updateRequestOrigin.bind(cf);
cf.updateRequestOrigin = (o) => {
  if (o.originAccessControlConfig && !o.originAccessControlConfig.enabled) delete o.originAccessControlConfig;
  _update(o);
};`,
    },
  },
});
```

## Environment

- SST: v4.7.3
- AWS provider: 7.20.0
- Region: us-east-2
- CloudFront Function runtime: cloudfront-js-2.0

turisanapo:
## Problem

`router.route("/", "http://<alb-dns>")` returns **502 "Failed to contact the origin"** because CloudFront tries HTTPS:443 instead of HTTP:80.

## Root cause

In `setUrlOrigin`, when `override.protocol === "http"`, only `customOriginConfig` is deleted — the expectation being that `cf.updateRequestOrigin` inherits the placeholder origin's `http-only` config.

However:
1. The remaining `originAccessControlConfig: { enabled: false }` prevents this inheritance
2. Even deleting **both** properties still fails — `cf.updateRequestOrigin` does not reliably inherit `customOriginConfig` from the placeholder origin when the `domainName` changes dynamically

## Fix

Explicitly set `customOriginConfig` to HTTP port 80 for HTTP origins, and remove the unused `originAccessControlConfig`:

```diff
  if (override.protocol === "http") {
-   delete origin.customOriginConfig;
+   origin.customOriginConfig = { port: 80, protocol: "http", sslProtocols: ["TLSv1.2"] };
+   delete origin.originAccessControlConfig;
  }
```

The `override.originAccessControlConfig` check below still allows OAC-enabled origins (`protection: "oac"`) to re-add it explicitly, so existing behavior is preserved.

## Validation

Tested by deploying a CloudFront Function that calls `cf.updateRequestOrigin` with explicit HTTP `customOriginConfig` — confirmed 200 responses through CloudFront to an HTTP-only ALB. Verified the `cf` module is completely frozen in CloudFront Functions runtime (`writable: false, configurable: false`), so runtime workarounds via injection are not possible.

Fixes #6742

### DIFF
--- a/platform/src/components/aws/router.ts
+++ b/platform/src/components/aws/router.ts
@@ -2965,7 +2965,8 @@ function setUrlOrigin(urlHost, override) {
   };
   override = override ?? {};
   if (override.protocol === "http") {
-    delete origin.customOriginConfig;
+    origin.customOriginConfig = { port: 80, protocol: "http", sslProtocols: ["TLSv1.2"] };
+    delete origin.originAccessControlConfig;
   }
   if (override.connectionAttempts) {
     origin.connectionAttempts = override.connectionAttempts;


turisanapo:
This fix is insufficient — we validated that just deleting both `customOriginConfig` and `originAccessControlConfig` still results in 502. CloudFront doesn't reliably inherit `customOriginConfig` from the placeholder origin when the domain changes dynamically via `cf.updateRequestOrigin`.

The correct fix is to **replace** `customOriginConfig` with explicit HTTP settings:

```diff
  if (override.protocol === "http") {
-   delete origin.customOriginConfig;
+   origin.customOriginConfig = { port: 80, protocol: "http", sslProtocols: ["TLSv1.2"] };
+   delete origin.originAccessControlConfig;
  }
```

See the updated analysis in #6742.

PR #6744, opened 2026-04-15
Add `routes` option to `cloudflare.Worker`

Makisuo:
## What this adds

A new `routes` field on `sst.cloudflare.Worker` that attaches the Worker to one or more Cloudflare [route patterns](https://developers.cloudflare.com/workers/configuration/routing/routes/) as an alternative to a full custom domain.

```ts
new sst.cloudflare.Worker("Api", {
  handler: "src/worker.ts",
  routes: ["example.com/api/*"],
});
```

Each pattern becomes a `cloudflare.WorkersRoute` resource, with the zone looked up from the pattern's hostname via the existing `ZoneLookup` provider — same machinery `domain:` already uses.

## Why

Today `cloudflare.Worker` only exposes `domain: "..."`, which maps to `WorkersCustomDomain` — i.e. the Worker owns the whole hostname. That's the right answer when a Worker is the only thing serving a subdomain, but it's not the only shape CF supports.

Routes are the other half of the CF dashboard's "Domains & Routes" panel. They're pattern-based, so multiple Workers (or a Pages project) can share a hostname and unmatched paths fall through to the origin. Concrete cases I keep hitting:

- One Worker fronting `example.com/api/*` while the rest of the site is served by Pages or an existing origin
- Migrating off an old origin incrementally, pattern by pattern
- Running several Workers on the same apex domain

Without a native option you end up reaching into `@pulumi/cloudflare` directly and hand-rolling `WorkersRoute` + `ZoneLookup` next to the SST component. That works but it's friction for something that's a one-liner in wrangler.

### DIFF
--- a/platform/src/components/cloudflare/worker.ts
+++ b/platform/src/components/cloudflare/worker.ts
@@ -104,7 +104,7 @@ export interface WorkerArgs {
    * }
    * ```
    */
-  routes?: Input<Input<string>[]>;
+  routes?: Input<string>[];
   /**
    * Configure how your function is bundled.
    *
--- a/platform/src/components/cloudflare/worker.ts
+++ b/platform/src/components/cloudflare/worker.ts
@@ -3,6 +3,7 @@ import path from "path";
 import crypto from "crypto";
 import {
   ComponentResourceOptions,
+  Output,
   output,
   all,
   jsonStringify,
@@ -66,6 +67,44 @@ export interface WorkerArgs {
    * ```
    */
   domain?: Input<string>;
+  /**
+   * Attach the Worker to one or more [route patterns](https://developers.cloudflare.com/workers/configuration/routing/routes/)
+   * instead of a dedicated custom domain.
+   *
+   * Unlike `domain`, which gives the Worker a full hostname, routes are
+   * pattern-based. This lets multiple Workers (or a Pages project) share a
+   * hostname and fall through to the origin for unmatched paths. Use routes
+   * when you want to put a Worker in front of only part of a site, e.g.
+   * `example.com/api/*`.
+   *
+   * Each pattern must start with a concrete hostname that belongs to a
+   * Cloudflare zone on the same account. Wildcard hostnames like `*.example.com`
+   * are not supported here.
+   *
+   * A proxied DNS record for the hostname must already exist in the zone —
+   * routes do not auto-manage DNS, unlike custom domains. You can create
+   * one in the Cloudflare dashboard or alongside the Worker with a
+   * `cloudflare.DnsRecord` resource.
+   *
+   * Cannot be used together with `domain`.
+   *
+   * @example
+   *
+   * ```js
+   * {
+   *   routes: ["example.com/api/*"]
+   * }
+   * ```
+   *
+   * Multiple patterns on the same Worker:
+   *
+   * ```js
+   * {
+   *   routes: ["example.com/api/*", "api.example.com/*"]
+   * }
+   * ```
+   */
+  routes?: Input<Input<string>[]>;
   /**
    * Configure how your function is bundled.
    *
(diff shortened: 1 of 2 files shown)


platform/src/components/cloudflare/worker.ts
```diff
@@ -66,6 +67,44 @@ export interface WorkerArgs {
...
+   *
+   * Multiple patterns on the same Worker:
+   *
+   * ```js
+   * {
+   *   routes: ["example.com/api/*", "api.example.com/*"]
+   * }
+   * ```
+   */
+  routes?: Input<Input<string>[]>;
```
vimtor: refactor to [avoid pulumi leaks](https://github.com/pulumi/agent-skills/blob/main/authoring/skills/pulumi-best-practices/SKILL.md#1-never-create-resources-inside-apply):

```suggestion
  routes?: Input<string>[];
```

vimtor (changes requested):
thanks for your contribution @Makisuo 

i left a minor comment, everything else looks very promising

may i ask what's your specific use case behind this feature?

platform/src/components/cloudflare/worker.ts
```diff
@@ -619,6 +661,43 @@ export class Worker extends Component implements Link.Linkable {
...
     }
+
+    function createWorkersRoutes() {
+      if (!args.routes) return;
+      if (args.domain)
+        throw new VisibleError(
+          `Cannot set both "domain" and "routes" on the "${name}" Worker. Use "domain" for a dedicated hostname, or "routes" for pattern-based routing.`,
+        );
+
+      return output(args.routes).apply((patterns) =>
```
vimtor: the other requested change was so you can do a foreach here instead of `output(args.router)`

vimtor (changes requested):

Makisuo:
I use Cloudflare Routes quite a lot for wildcard domains. 
I know it's possible to just do this with ZoneRules instead, but then it won't be automatically linked and shown in Cloudflare making it harder for me and other team members to work with.

platform/src/components/cloudflare/worker.ts
Makisuo: Ah makes sense will fix in a bit after lunch

platform/src/components/cloudflare/worker.ts
```diff
@@ -66,6 +66,44 @@ export interface WorkerArgs {
...
+   * pattern-based. This lets multiple Workers (or a Pages project) share a
+   * hostname and fall through to the origin for unmatched paths. Use routes
+   * when you want to put a Worker in front of only part of a site, e.g.
+   * `example.com/api/*`.
+   *
+   * Each pattern must start with a concrete hostname that belongs to a
+   * Cloudflare zone on the same account. Wildcard hostnames like `*.example.com`
+   * are not supported here.
+   *
+   * A proxied DNS record for the hostname must already exist in the zone —
```
vimtor, line 83: as i'm learning more about this routes feature, i'm thinking whether it's a good idea to automatically create the DNS record for the user

if we don't, probably we should add an example of how to do it. ideally, we would create it, but i guess if you have multiple workers, that defeats the purpose

maybe you're more familiar and have a clear idea of how to deal with this

vimtor: any ideas @Makisuo?

for the recent `redirects` features we are creating the record

Makisuo: Sorry completely missed the previous comments, Github Notifications are a bit broken for me currently. 
Think it makes sense if we automatically create a DNS record if not already existing. 

Have often seen people be confused why routes weren't working since this pattern isn't quite clear anyways, so this makes a lot of sense

vimtor: great! if you add that, i will merge the pr

ideally, also attach some screenshots/video of it working

PR #6749, opened 2026-04-15
ensure apigatewayv2 cors configuration is not set when input is 'false'

### LINKED ISSUE #6748: APIGatewayV2 construct doesn't actually clear the CORS config when `cors: false` is set
Hello,

I am using the `APIGatewayV2` construct to deploy API gateway. I set `cors: false` as I manage the CORS at my app level (I need wildcard domains), but the deployed API Gateway still seems to have a CORS config. When I describe it using: `aws apigatewayv2 get-api --api-id <id> --query "CorsConfiguration" --region us-east-2`, it returns `{}`, which is resulting in a default CORS configuration with an empty header list and empty allowed origin list. When I click the "Clear" button on the API Gateway's CORS configuration page in the AWS Console, and then describe the CORS configuration again using the same command, it returns `null`, and my requests start working.

Here is my code:

```
export const api = new sst.aws.ApiGatewayV2("API", {
  domain,
  cors: false,
  vpc: {
    securityGroups: [apigatewaySecurityGroup.id],
    subnets: privateSubnetIds,
  },
  transform: {
    api: {
      name: `sst-${$app.stage}-APIApi`,
    },
    logGroup: {
      name: `/aws/vendedlogs/apis/sst-${$app.stage}-API`,
    },
    stage: {
      defaultRouteSettings: {
        throttlingBurstLimit: 5000,
        throttlingRateLimit: 10000,
      },
    },
  },
});
```

Probotect0r:
Fixes #6748 

### Background
Specifying `cors: false` on the ApiGatewayv2 construct results in the cors configuration being set to `{}` which actually sets a barebones default configuration instead of completely clearing it. The barebones default config results in all requests to the Api failing cors because the allowed origins and headers list is empty.

### Notes
Because the corsConfiguration parameter on the pulumi api construct is `Output<ApiArgs> | undefined`, I had to wrap the whole creation of the apigateway inside a `.apply()`. 

I tested this code change by deploying it and verifying the cors config value using `aws apigatewayv2 get-api --api-id <my id> --query "CorsConfiguration" --region us-east-2`

### DIFF
--- a/platform/src/components/aws/apigatewayv2.ts
+++ b/platform/src/components/aws/apigatewayv2.ts
@@ -675,7 +675,7 @@ export class ApiGatewayV2 extends Component implements Link.Linkable {
   private constructorName: string;
   private constructorArgs: ApiGatewayV2Args;
   private constructorOpts: ComponentResourceOptions;
-  private api: apigatewayv2.Api;
+  private api: Output<apigatewayv2.Api>;
   private apigDomain?: Output<apigatewayv2.DomainName>;
   private apiMapping?: Output<apigatewayv2.ApiMapping>;
   private logGroup: cloudwatch.LogGroup;
@@ -692,7 +692,6 @@ export class ApiGatewayV2 extends Component implements Link.Linkable {
 
     const accessLog = normalizeAccessLog();
     const domain = normalizeDomain();
-    const cors = normalizeCors();
     const vpc = normalizeVpc();
 
     const vpcLink = createVpcLink();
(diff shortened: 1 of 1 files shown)


Probotect0r:
@vimtor thoughts?

platform/src/components/aws/apigatewayv2.ts
```diff
@@ -757,25 +756,6 @@ export class ApiGatewayV2 extends Component implements Link.Linkable {
       });
     }
 
-    function normalizeCors() {
-      return output(args.cors).apply((cors) => {
-        if (cors === false) return {};
```
vimtor, line 762: wouldn't just returning `undefined` here do the trick?

vimtor: if you did this because typescript complains about pulumi, we can either do `@ts-ignore` or unwrap `cors` from `Input` in the type definition

i lean towards the latter

Probotect0r: Simply returning `undefined` doesn't work, because it needs to be returned from inside the `.apply()`, since the input argument type is `Input<boolean |  Apiv2args...>`. But returning `undefined` from inside the `apply()` results in a return type of `Output<Apiv2CorsConfiguration | undefined>`, which is not assignable to `Input<Apiv2CorsConfiguration> | undefined`.

What are the consequences of changing the `cors` argument type to `boolean | Prettify<ApiGatewayV2CorsArgs>`? The individual fields on `ApiGatewayV2CorsArgs` are still `Input`, so I am guessing it's fine? Still allows someone to pass in an Input for the specific fields?

vimtor: it would break on the very very unlikely scenario that someone is passing an output to `cors`

we should do the `// @ts-ignore` instead to be safe

PR #6767, opened 2026-04-18
fix: use alias for versioned lambda functions

### LINKED ISSUE #6722: Event sources should invoke Lambda through an alias when versioning is enabled

anatolzak:
closes https://github.com/anomalyco/sst/issues/6722 https://github.com/anomalyco/sst/issues/6720

### Summary

Reworks function versioning to always create a persistent `LatestAlias` when `versioning` (or `durable`) is enabled. `targetArn`, `qualifier`, Function URLs, URL permissions, and invoke ARNs now all route through this alias instead of a pinned numeric version.

Previously:
- `targetArn` pointed at `fn.qualifiedArn` (a specific published version number that churns each deploy).
- Durable functions created a one-off `${name}Durable` alias inline, but only for the Function URL — other consumers still got the pinned version ARN.

Now:
- A single `${name}LatestAlias` is created whenever the function is published, and every downstream integration (URL, event sources, API GW, cron, workflow, SDK link, etc.) targets it.
- `fn.qualifier` resolves to the alias name when versioning is on, and `undefined` when it isn't.
- The URL is bound to the alias (no longer to `$LATEST`) as soon as versioning is enabled.

This is groundwork for upcoming rollout support — we need a stable indirection in front of versioned functions before we can implement traffic shifting.

### What I tested

- Workflow **without a URL** — correctly routes through the alias.
- Workflow **with a URL** — URL is served off the alias.
- Function with URL, **no versioning** — no regression, URL still targets the unqualified function.
- Function with URL **and versioning** — alias created, URL targets the alias, permissions scoped to the alias.
- `Resource.<Fn>.qualifier` via the SDK — resolves to the alias name when versioning is enabled, and has no value when it isn't.
- Lambda behind **API Gateway**:
  - without versioning: integration and permissions target the function directly, as before.
  - with versioning: integration points at the alias and the `lambda:InvokeFunction` permission is scoped to the alias.
- **Response streaming** path — `targetResponseStreamingInvokeArn` now rewrites against the alias ARN, and streamed invocations work.
- AWS Router with protection (OAC) + Lambda **no versioning** — no regression, URL still targets the unqualified function.
- AWS Router with protection (OAC) + Lambda **with versioning** — alias created, Router targets the alias, permissions scoped to the alias.
- AWS SQS + Lambda **no versioning** — no regression, queue invoked the unqualified function.
- AWS SQS + Lambda **with versioning** — alias created, SQS triggers the alias.

### Upgrade notes

For any stack that previously enabled `versioning: true` or `durable: true` **with a URL**, the Function URL will be replaced:

- **`versioning: true`**: the existing URL currently points at `$LATEST`. After upgrade it will point at the alias — a new URL will be issued.
- **`durable: true`**: the alias changes from the old inline `${name}Durable` to `${name}LatestAlias`, which also causes a new URL to be issued.

Event sources that consume `fn.targetArn` (buckets, topics, crons, API Gateway integrations, etc.) will see a one-time update on upgrade as the ARN moves from a pinned version to the alias. This isn't new churn — previously, every deploy that changed function code published a new version, which forced those same event sources and permissions to update their target. With the alias, this happens once and then stops: subsequent deploys keep pointing at the same alias ARN, so downstream resources no longer need to be touched on every deploy.

### DIFF
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -1513,6 +1513,11 @@ export interface FunctionArgs {
      */
     function?: Transform<lambda.FunctionArgs>;
     /**
+     * Transform the Lambda Alias resource. This is only created when versioning
+     * is enabled (via `versioning: true` or `durable: true`).
+     */
+    alias?: Transform<lambda.AliasArgs>;
+    /**
      * Transform the IAM Role resource.
      */
     role?: Transform<iam.RoleArgs>;
@@ -2726,12 +2731,15 @@ export class Function extends Component implements Link.Linkable {
         if (!isVersioningEnabled) return;
 
         return new lambda.Alias(
-          `${name}LatestAlias`,
-          {
-            functionName: fn.name,
-            functionVersion: fn.version,
-          },
-          { parent },
+          ...transform(
+            args.transform?.alias,
+            `${name}LatestAlias`,
+            {
+              functionName: fn.name,
+              functionVersion: fn.version,
+            },
+            { parent },
+          ),
         );
       });
     }
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -1726,8 +1726,8 @@ export class Function extends Component implements Link.Linkable {
   private eventInvokeConfig?: lambda.FunctionEventInvokeConfig;
   private alias: Output<lambda.Alias | undefined>;
   private isVersioningEnabled: Output<boolean>;
-  public targetArn: Output<string>;
-  public qualifier: Output<string | undefined>;
+  #targetArn: Output<string>;
+  #qualifier: Output<string | undefined>;
 
   private static readonly encryptionKey = lazy(
     () =>
@@ -1815,8 +1815,8 @@ export class Function extends Component implements Link.Linkable {
     this.urlEndpoint = urlEndpoint;
     this.eventInvokeConfig = eventInvokeConfig;
     this.alias = alias;
-    this.targetArn = targetArn;
-    this.qualifier = qualifier;
+    this.#targetArn = targetArn;
+    this.#qualifier = qualifier;
     this.isVersioningEnabled = isVersioningEnabled;
 
     const buildInput = output({
@@ -3040,6 +3040,25 @@ export class Function extends Component implements Link.Linkable {
    */
   public get arn() {
     return this.function.arn;
+  }
+
+  /**
+   * The ARN to use when wiring this function to an event source.
+   *
+   * When versioning is enabled, this points to the Lambda alias that routes
+   * to the currently published version. Otherwise it falls back to the
+   * function ARN.
+   *
+   * Prefer this over `arn` when passing the function to an event source, so
+   * invocations go through the alias instead of `$LATEST`.
+   */
+  public get targetArn() {
+    return this.#targetArn;
+  }
+
+  /** @internal */
+  public get qualifier() {
+    return this.#qualifier;
   }
 
   /** @internal */
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -1779,7 +1779,7 @@ export class Function extends Component implements Link.Linkable {
     const streaming = normalizeStreaming();
     const logging = normalizeLogging();
     const volume = normalizeVolume();
-    const url = normalizeUrl(args.url);
+    const url = normalizeUrl();
     const copyFiles = normalizeCopyFiles();
     const durable = normalizeDurable();
     const policies = output(args.policies ?? []);
@@ -1988,8 +1988,8 @@ export class Function extends Component implements Link.Linkable {
       }));
     }
 
-    function normalizeUrl(url: FunctionArgs["url"]) {
-      return output(url).apply((url) => {
+    function normalizeUrl() {
+      return output(args.url).apply((url) => {
         if (url === false || url === undefined) return;
         if (url === true) {
           url = {};
@@ -2023,7 +2023,6 @@ export class Function extends Component implements Link.Linkable {
         };
       });
     }
-    type NormalizedUrl = ReturnType<typeof normalizeUrl>;
 
     function normalizeCopyFiles() {
       return output(args.copyFiles ?? []).apply((copyFiles) =>
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -2652,7 +2652,7 @@ export class Function extends Component implements Link.Linkable {
               layers: args.layers,
               tags: args.tags,
               publish: output(args.versioning).apply(
-                (v) => v || Boolean(args.durable),
+                (v) => v ?? Boolean(args.durable),
               ),
               reservedConcurrentExecutions: concurrency?.reserved,
               durableConfig: durable && {
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -1309,6 +1309,12 @@ export interface FunctionArgs {
   /**
    * Enable versioning for the function.
    *
+   * :::caution
+   * If you're wiring this function to an event source by passing its ARN,
+   * use `fn.targetArn` rather than `fn.arn`. Using `fn.arn` invokes
+   * `$LATEST` and bypasses the alias that routes to the published version.
+   * :::
+   *
    * :::note
    * Durable functions enable this by default.
    * :::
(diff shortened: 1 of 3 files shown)


platform/src/components/aws/function.ts
```diff
@@ -2630,7 +2657,7 @@ export class Function extends Component implements Link.Linkable {
               layers: args.layers,
               tags: args.tags,
               publish: output(args.versioning).apply(
-                (v) => v ?? Boolean(args.durable),
+                (v) => v || Boolean(args.durable),
```
vimtor: why did we change this?

platform/src/components/aws/function.ts
```diff
@@ -2002,6 +2028,7 @@ export class Function extends Component implements Link.Linkable {
         };
       });
     }
+    type NormalizedUrl = ReturnType<typeof normalizeUrl>;
```
vimtor: i don't think we need this

platform/src/components/aws/function.ts
```diff
@@ -1967,8 +1993,8 @@ export class Function extends Component implements Link.Linkable {
       }));
     }
 
-    function normalizeUrl() {
-      return output(args.url).apply((url) => {
+    function normalizeUrl(url: FunctionArgs["url"]) {
```
vimtor: the normalize function usually don't take arguments

platform/src/components/aws/function.ts
```diff
@@ -2700,7 +2727,33 @@ export class Function extends Component implements Link.Linkable {
...
+    }
+
+    function createTargetArn(alias: Input<lambda.Alias | undefined>) {
+      return output(alias).apply((alias) => {
+        if (alias) return alias.arn;
+        return fn.arn;
+      });
+    }
+
+    type CreateUrlOpts = {
```
vimtor: inline types like this are very weird

platform/src/components/aws/workflow.ts
```diff
@@ -272,7 +272,12 @@ export interface WorkflowArgs
  * ID generation inside durable operations like `ctx.step()`.
  *
  * :::caution
- * Workflow handlers have versioning enabled. Deploying an update won't update existing running workflows.
```
vimtor, line 275: we should keep the versioned behaviour. i don't understand why we changed this

platform/src/components/aws/function.ts
```diff
@@ -2983,46 +3042,42 @@ export class Function extends Component implements Link.Linkable {
...
+   * function ARN.
+   *
+   * Prefer this over `arn` when passing the function to an event source, so
+   * invocations go through the alias instead of `$LATEST`.
+   */
   public get targetArn() {
-    return this.useQualifiedTarget.apply((useQualifiedTarget) =>
-      useQualifiedTarget ? this.function.qualifiedArn : this.arn,
-    );
+    return this.#targetArn;
```
vimtor: why don't define targetArn here instead of adding this `#targetArn`?

platform/src/components/aws/function.ts
```diff
@@ -1781,7 +1797,13 @@ export class Function extends Component implements Link.Linkable {
     const logGroup = createLogGroup();
     const zipAsset = createZipAsset();
     const fn = createFunction();
-    const urlEndpoint = createUrl();
+    const isVersioningEnabled = fn.publish.apply((publish) => Boolean(publish));
```
vimtor, line 1803: you can also define versioning with `versioning: true`

platform/src/components/aws/function.ts
```diff
@@ -2700,7 +2727,33 @@ export class Function extends Component implements Link.Linkable {
...
+          {
+            functionName: fn.name,
+            functionVersion: fn.version,
+          },
+          { parent },
+        );
+      });
+    }
+
+    function createTargetArn(alias: Input<lambda.Alias | undefined>) {
```
vimtor: this is weird because it's not creating anything. that's why i said to use the getter directly in a comment below

platform/src/components/aws/function.ts
```diff
@@ -2886,6 +2932,15 @@ export class Function extends Component implements Link.Linkable {
       });
     }
 
+    function createLatestUrl() {
+      return qualifier.apply((qualifier) =>
+        createUrl({
```
vimtor: we should inline the `createUrl` function

or keep the `createUrl`, remove this `createLatestUrl` and move the logic to the existing method

vimtor (changes requested):
thanks for your contribution @anatolzak 

i left some comments

platform/src/components/aws/workflow.ts
anatolzak: @vimtor The versioned behavior is unchanged. The main reason durable workflows version the handler is so that an in-flight execution resumes on the same code it started on. That's preserved: the lanbda API captures the resolved version at execution start and pins all resumes/retries to that captured version, regardless of whether the alias is later moved.

What the alias buys us, beyond what a pinned numeric version gives, is decoupling the handler identity from the deployed version. If you ship a bad release and need to roll back, you just point the alias at the previous version, and no consumers have to be updated. Any newly-started executions immediately run against the rolled-back version. With a pinned `qualifiedArn`, every consumer (URL, event sources, API GW, SDK `qualifier`, etc.) would have to be re-wired to roll back.

It also opens the door to weighted aliases: you can split traffic across two versions of a workflow handler for A/B tests or canary rollouts without touching any client. None of that is possible if consumers point at a specific numeric version.

The doc rewrite just makes the mechanism explicit and adds the `$LATEST` escape hatch. No behavior regression.

platform/src/components/aws/function.ts
anatolzak: @vimtor  `fn.publish` is the resolved truth on the underlying `aws.lambda.Function`. Upstream at function.ts:2659, it's set to `args.versioning || Boolean(args.durable)`, so checking `fn.publish` covers both paths in one expression

anatolzak:
@vimtor thanks for the review! I think I addressed all your comments.

A couple of notes on the changes you flagged:

The `createLatestUrl` wrapper around `createUrl` and the dynamic arguments on `normalizeUrl` were leftovers from the rollout PR, which needs to create two URLs (one per traffic-shifted version), so URL normalization and creation had to accept dynamic arguments. Since none of that applies to this PR, I reverted both back to their original signatures.

On using a private `#targetArn` field instead of a getter: I originally went that route to keep the diff minimal, but I switched to getters as you asked. That meant reordering the instance property assignments in the constructor, which I was trying to avoid to keep the diff minimal.

The chain is: `createUrl` now reads `qualifier`, `qualifier` reads `targetArn`, and `targetArn` reads `this.function`. When `targetArn` was a stored private field, `createUrl` could just read the captured value, so the order didn't matter. With a getter, every call walks back through `this.function`, so `this.function` has to be assigned before `createUrl()` runs. The new constructor order reflects that.

Let me know if there's anything else you'd like changed.

anatolzak:
hey @vimtor! thanks again for the review! I've addressed the feedback in the latest commits. Let me know if there's anything else you'd like me to change

PR #6770, opened 2026-04-19
Allow Cloudflare components to override `accountId`

### LINKED ISSUE #6769: Allow overriding of AccountId for Cloudflare resources used under the hood
The Cloudflare provider is a bit weird in the way it expects you to specify the accountId on the resource, but API token on the provider.
SST made the choice to default the accountId to an environment variable for all the resources it provisions, as it is annoying to specify this for every resource, however this makes it difficult for multi-account deployments.
This is compounded as some of the resources can't be accessed/overridden by the transforms property (e.g. the custom domain of a worker).
It would be nice to have the account id as an optional argument to override this default environment variable, or have all resources exposed so we can transform them.

DJDANNY123:
Closes https://github.com/anomalyco/sst/issues/6769

## Context
The Cloudflare provider is a bit weird in the way it expects you to specify the accountId on the resource, but API token on the provider.
SST made the choice to default the accountId to an environment variable for all the resources it provisions, as it is annoying to specify this for every resource, however this makes it difficult for multi-account deployments.
This is compounded as some of the resources can't be accessed/overridden by the transforms property (e.g. the custom domain of a worker).

## This Change
This change simply extends each cloudflare SST component with an optional accountId, which will be used if specified for provisioning/querying all resources within this component.

### DIFF
--- a/platform/src/components/cloudflare/bucket.ts
+++ b/platform/src/components/cloudflare/bucket.ts
@@ -4,8 +4,15 @@ import { Component, Transform, transform } from "../component";
 import { Link } from "../link.js";
 import { binding } from "./binding.js";
 import { DEFAULT_ACCOUNT_ID } from "./account-id";
+import type { Input } from "../input";
 
 export interface BucketArgs {
+  /**
+   * The Cloudflare account ID to use for this R2 Bucket.
+   * Overrides the default account ID set via `CLOUDFLARE_DEFAULT_ACCOUNT_ID`.
+   * @internal
+   */
+  accountId?: Input<string>;
   /**
    * [Transform](/docs/components/#transform) how this component creates its underlying
    * resources.
@@ -73,7 +80,7 @@ export class Bucket extends Component implements Link.Linkable {
           `${name}Bucket`,
           {
             name: "",
-            accountId: DEFAULT_ACCOUNT_ID,
+            accountId: args?.accountId ?? DEFAULT_ACCOUNT_ID,
           },
           { parent },
         ),
--- a/platform/src/components/cloudflare/cron.ts
+++ b/platform/src/components/cloudflare/cron.ts
@@ -81,6 +81,12 @@ export interface CronArgs {
    */
   schedules: Input<string[]>;
   /**
+   * The Cloudflare account ID to use for this Cron and its worker.
+   * Overrides the default account ID set via `CLOUDFLARE_DEFAULT_ACCOUNT_ID`.
+   * @internal
+   */
+  accountId?: Input<string>;
+  /**
    * [Transform](/docs/components/#transform) how this component creates its underlying
    * resources.
    */
@@ -142,7 +148,6 @@ export class Cron extends Component {
     const workerArgs = normalizeWorker();
     const worker = createWorker();
     const trigger = createTrigger();
-
     this.worker = worker;
     this.trigger = trigger;
 
@@ -159,7 +164,7 @@ export class Cron extends Component {
         throw new VisibleError(
           `You must provide a "worker" for the "${name}" Cron component.`,
         );
-      return workerBuilder(`${name}Handler`, workerArgs);
+      return workerBuilder(`${name}Handler`, workerArgs, undefined, undefined, args.accountId);
     }
 
     function createTrigger() {
@@ -169,7 +174,7 @@ export class Cron extends Component {
             args.transform?.trigger,
             `${name}Trigger`,
             {
-              accountId: DEFAULT_ACCOUNT_ID,
+              accountId: args.accountId ?? DEFAULT_ACCOUNT_ID,
               scriptName: worker.script.scriptName,
               schedules: schedules.map((s) => ({ cron: s })),
             },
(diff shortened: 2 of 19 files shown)


vimtor:
thanks for your contribution @DJDANNY123

this week we're adding a lot of Cloudflare components, let's wait until we finish so you can update your PR

in the meantime, i wonder if this is the ideal setup. maybe we should let you define multiple providers, each using a different account. not sure how that would look, though

vimtor:
just realized that what we should do is to find a way to do this:

```ts
const provider = new cloudflare.Provider("OtherAccount", {
  apiToken: "",
})

new sst.cloudflare.Bucket("Bucket", {}, { provider });
```

this would match the aws behaviour

DJDANNY123:
I don't understand your code example, I believe we can already do this with the existing implementation.
The issue is in specifying the accountId which is a required argument for all plain cloudflare components, and we can't specify the accountId in the provider.

Ideally the cloudflare provider would accept an accountId, and use this for relevant resources but this would probably be a fairly major change in the cloudflare Pulumi resource provider.

I think given the current state of the downstream provider, we should mimic how the downstream provider works (but still keep the existing default account behaviour for ease of people who don't need to do multi-account deployments)

```ts
// define a provider with another api token for a different account
const provider = new cloudflare.Provider("AnotherProvider", { apiToken: "" });

// deploy a plain cloudflare resource in the other account
const plainCfResource = new cloudflare.R2Bucket(
  "",
  { name: "MyBucket", accountId: "ANOTHER_ACCOUNT_ID" },
  { provider },
);

// deploy a sst component in another account in the same way
const sstResource = new sst.cloudflare.Worker(
  "MyWorker",
  { handler: "index.ts", accountId: "ANOTHER_ACCOUNT_ID" },
  { provider },
);
```

vimtor:
yeah, we should keep the default account behavior

what i'm saying is that the code snippet you provided might already work, so i'm not sure if this pr is necessary

DJDANNY123:
The snippet I provided would work if every SST component had an optional accountId as part of their arguments, and was applied to all resources provisioned by these components, however this isn't the case currently.
I'm happy to update this PR when you've finished with your spree of adding new cloudflare resources

vimtor:
```ts
// define a provider with another api token for a different account
const provider = new cloudflare.Provider("AnotherProvider", { apiToken: "" });

// deploy a plain cloudflare resource in the other account
const plainCfResource = new cloudflare.R2Bucket(
  "",
  { name: "MyBucket" },
  { provider },
);

// deploy a sst component in another account in the same way
const sstResource = new sst.cloudflare.Worker(
  "MyWorker",
  { handler: "index.ts" },
  { provider },
);
```

we should strive for this and push the accountId selection logic inside each component

does that sound good?

DJDANNY123:
The only way I see to consistently do this is an update to the main cloudflare provider to accept accountId as a provider option.

If we try to inject vanilla cloudflare pulumi constructs with account IDs I feel like it would get messy and deviate from SST being a simple wrapper around plain pulumi, as the types of arguments for these plain constructs would also have to change, but I'm open to hear if there's a way to achieve this I haven't considered.

vimtor:
i guess we could do something like:

```ts
import * as cloudflare from "@pulumi/cloudflare";

const account = cloudflare.getAccountsOutput({}).accounts.apply(accounts => {
    if (accounts.length !== 1) {
        throw new Error(`expected exactly one Cloudflare account, got ${accounts.length}`);
    }
    return accounts[0];
});

export const accountId = account.id;
```

and force people's tokens to be exactly for one account

not sure, let me think about it more

DJDANNY123:
I like the idea, I just don't know if it's much better than the alternative of optionally supplying the account id, as the improved ergonomics may not outweigh the inconsistency of specifying account id, as with the vanilla cloudflare pulumi components, accountId is a required variable.

Unless we were able to modify this as well by overriding all of these vanilla components to omit this from the argument types?

DJDANNY123:
@vimtor have you had any further thoughts on this one?

vimtor:
let's go ahead with your approach but mark the `accountId` as internal and add a doc section in the cloudflare.mdx guide

DJDANNY123:
I've updated this PR to mark each accountId as internal and I've added some documentation of supporting multiple accounts under the credentials section of the cloudflare.mdx.
Wanted to point out I've updated the public D1.get to change the function signature from a string of databaseId to an object allowing for databaseId and accountId which is a breaking change, should I look to support both and deprecate the original?

DJDANNY123:
I've updated this with a proper deprecation of the old D1.get method to avoid breaking change, so happy for this to be reviewed now @vimtor

DJDANNY123:
> let's go ahead with your approach but mark the `accountId` as internal and add a doc section in the cloudflare.mdx guide

@vimtor when you have some time, could you give this a review please?

vimtor (changes requested):
- platform/src/components/cloudflare/static-site.ts:291 and :384: StaticSite does not pass accountId to the KV namespace or router Worker; only asset upload uses it at :368. This can create resources in the default account, then upload assets using the override account.
- platform/src/components/cloudflare/ssr-site.ts:234: SsrSite writes accountId into Wrangler config at :188, but never passes it to the deployed Worker, so the actual Pulumi resources still use DEFAULT_ACCOUNT_ID.
- platform/src/components/cloudflare/worker.ts:769: worker domain aliases/redirects still hardcode DEFAULT_ACCOUNT_ID, while the primary domain uses the override. Aliases/redirects will target the wrong account.
- platform/src/components/cloudflare/hyperdrive.ts:372: Hyperdrive.get still hardcodes DEFAULT_ACCOUNT_ID, unlike the updated D1.get/Kv.get.

Issue #6759, opened 2026-04-17
Step Function should enable `states:RedriveExecution` to redrive failed workflows from AWS console


MengLinMaker:
Currently you cannot redrive failed Step Function since this is not enabled by the SST defined IAM role. I think this should be [enabled by default through adding `states:RedriveExecution` action](https://github.com/anomalyco/sst/blob/b550ff117e6b0b8ac33e2802fc8bec0e18fec9e6/platform/src/components/aws/step-functions.ts#L292-L295).

Additionally a `permissions` property can be added to `StepFunction` component for flexibility - [taking inspiration from the `Function` component](https://github.com/anomalyco/sst/blob/b550ff117e6b0b8ac33e2802fc8bec0e18fec9e6/platform/src/components/aws/function.ts#L638).

`Function` component also allows role alterations through `transform.role`. Though I'm not convinced this is a useful addition. Integration with existing external roles maybe?

Current error when trying to redrive:
```
Error contacting AWS Service. | 
Message from Service: User: 
arn:aws:sts::585946893245:assumed-role/new-homie-production-StepFunctionsScrapePipelineRole-xuwmcrdn/ItoRwPRAqUHSBnwXpssuaJcueXdYVqPJ 
is not authorized to perform: 
states:RedriveExecution on resource: 
arn:aws:states:ap-southeast-2:585946893245:execution:new-homie-production-StepFunctionsScrapePipelineStateMachine-wxordevn/16751907-e18b-326c-8293-d4685b172856:e9c7f9a4-af38-340a-8f3c-dff6febafead 
because no identity-based policy allows the states:RedriveExecution action 
(Service: Sfn, Status Code: 400, Request ID: 667bea2d-a675-4a83-bf19-01b18d9f0ca2) 
(SDK Attempt Count: 1)
```

PR #6779, opened 2026-04-20
feat: allow additional permissions to `StepFunctions` plus defaults #6759

MengLinMaker:
Addresses #6759 as current permissions are to restrictive and cannot be extended.

This PR adds:
- [x] redrive and stop enabled by default.
- [x] ability to extend permissions.
- [ ] ability to attach role (unsure if this is a good idea).
- [ ] add example to test redrive ability.

More details are listed in issue.

### DIFF
--- a/platform/src/components/aws/step-functions.ts
+++ b/platform/src/components/aws/step-functions.ts
@@ -292,6 +292,7 @@ export class StepFunctions extends Component implements Link.Linkable {
                     actions: [
                       "states:StartExecution",
                       "states:DescribeExecution",
+                      "states:RedriveExecution",
                     ],
                     resources: ["*"],
                   },
--- a/platform/src/components/aws/step-functions.ts
+++ b/platform/src/components/aws/step-functions.ts
@@ -291,6 +291,7 @@ export class StepFunctions extends Component implements Link.Linkable {
                   {
                     actions: [
                       "states:StartExecution",
+                      "states:StopExecution",
                       "states:DescribeExecution",
                       "states:RedriveExecution",
                     ],
--- a/platform/src/components/aws/step-functions.ts
+++ b/platform/src/components/aws/step-functions.ts
@@ -4,7 +4,7 @@ import {
   interpolate,
   output,
 } from "@pulumi/pulumi";
-import { Component, Transform, transform } from "../component";
+import { Component, Prettify, Transform, transform } from "../component";
 import { cloudwatch, iam, sfn } from "@pulumi/aws";
 import { Link } from "../link";
 import { permission } from "./permission";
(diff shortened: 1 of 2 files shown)


platform/src/components/aws/step-functions.ts
```diff
@@ -31,6 +31,77 @@ import { physicalName } from "../naming";
 import { functionBuilder } from "./helpers/function-builder";
 import { Function } from "./function.js";
 
+export type PermissionArgs = {
```
MengLinMaker: This is copied from `Function` component. I think we can extract this to `helpers` folder.

vimtor:
i'm not super familiar with step functions, but i believe just adding the stop and redrive permissions should be enough

also, i believe you can already attach permissions by doing something like this:

```ts
sst.aws.StepFunctions.lambdaInvoke({
  name: "LambdaInvoke",
  function: {
    handler: "src/index.handler"
    timeout: "60 seconds",
    permissions: [...]
  }
});
```

MengLinMaker:
@vimtor Yep the stop and redrive permission is enough for now. But it also cannot be extended nicely currently if requirements change.

Unfortunately the code you suggested is for running Lambda inside step Functions and alters the permission of the Lambda only.

Step Function (orchestrator for running workflows) currently doesn't have permission to redrive failures (manually continue from where errors occurred). Large workflows like the several hour workflows I'm frequently running benefit from this feature, especially with spot instances. 

It's possible to override the IAM role via transform in a messy way:

```TypeScript
// Similar IAM role to what SST provides
const StepFunctionsScrapePipelineRole = new aws.iam.Role('StepFunctionsScrapePipelineRole', {
    assumeRolePolicy: aws.iam.assumeRolePolicyForPrincipal({
        Service: 'states.amazonaws.com',
    }),
    inlinePolicies: [
        {
            name: 'inline',
            policy: aws.iam.getPolicyDocumentOutput({
                statements: [
                    {
                        actions: ['events:*'],
                        resources: ['*'],
                    },
                    {
                        actions: [
                            'logs:CreateLogDelivery',
                            'logs:CreateLogStream',
                            'logs:GetLogDelivery',
                            'logs:UpdateLogDelivery',
                            'logs:DeleteLogDelivery',
                            'logs:ListLogDeliveries',
                            'logs:PutLogEvents',
                            'logs:PutResourcePolicy',
                            'logs:DescribeResourcePolicies',
                            'logs:DescribeLogGroups',
                        ],
                        resources: ['*'],
                    },
                    {
                        actions: [
                            'states:StartExecution',
                            'states:DescribeExecution',
                            // I only need this action for redriving
                            'states:RedriveExecution',
                        ],
                        resources: ['*'],
                    },
                    ...StepScrapePipelineDefinition.getRoot().getPermissions(),
                ],
            }).json,
        },
    ],
})
const StepFunctionsScrapePipeline = new sst.aws.StepFunctions('StepFunctionsScrapePipeline', {
    definition: StepScrapePipelineDefinition,
    // What a mess here, but it works
    transform: {
        stateMachine: (args) => {
            args.roleArn = StepFunctionsScrapePipelineRole.arn
            args.definition = pulumi.output(args.definition).apply((definition) => {
                const parsed = JSON.parse(definition)
                const mapState = parsed.States?.[StepMapScrapeLocality.name]
                // Would be good if this could be specified in `sst.aws.StepFunctions.map`
                if (mapState?.Type === 'Map') mapState.ToleratedFailurePercentage = 5
                return JSON.stringify(parsed)
            })
        },
    },
})
```

Another issue is that `ToleratedFailurePercentage` for step functions to run also requires very hacky code:
```
if (mapState?.Type === 'Map') mapState.ToleratedFailurePercentage = 5
```

PR #6785, opened 2026-04-20
Add Durable objects Component

### LINKED ISSUE #6730: Add an example for Cloudflare Durable Objects
Issue #6442  from a couple of months ago had a good discussion around nice patterns for using Cloudflare Durable Objects in conjunction with an SST app.

It would be nice to have it formalized as an official example with recommended patterns for a good local dev experience. This might also expose some areas where a custom SST component/integration could add value later.

This might be somewhat of a duplicate of #6683

bachiitter:
closes #6730


beeirl:
@vimtor can we get this reviewed please

DJDANNY123:
I've had a look at this MR and it's nice, it solves my biggest pain point with durable objects which is managing migrations.
I don't think it handles binding a single durable object namespace to multiple workers, which `links` should probably support if we make it an independent resource that can be associated with multiple other resources.

Worker 1 (host)
```ts
bindings: {
  name: "COUNTER",
  type: "durable_object_namespace",
  className: "CounterObject",
}
migrations: {
  newTag: "v1",
  newSqliteClasses: ["CounterObject"]
}
```

Worker 2 (consumer), no migrations
```ts
bindings: {
  name: "COUNTER",
  type: "durable_object_namespace",
  className: "CounterObject",
// I don't think this would be populated by linking the durable object, but could be overridden with a transform
  scriptName: "Worker1ScriptName", 
}
```
I'd still be in favour of this being merged in it's current state, as for most people this will solve headaches, and binding a single durable object namespace to multiple workers is probably fairly niche.

PR #6792, opened 2026-04-21
fix(aws): Add ap-southeast-6 to SSR site supported regions

### LINKED ISSUE #6791: Missing region ap-southeast-6 missing in ssr-site.ts
AWS launched the ap-southeast-6 region (Auckland, New Zealand) in August 2025. When deploying SST apps to this region, the deployment fails with:

Invalid AWS region: "ap-southeast-6". Please specify a valid AWS region.

Patch needed:

"ap-southeast-6": { lat: -36.8485, lon: 174.7633 }, // Auckland, New Zealand

right here https://github.com/anomalyco/sst/blob/258c7a1d3e216ae84f4230a1c7f95b32d57a2868/platform/src/components/aws/ssr-site.ts#L64

See also https://github.com/pulumi/pulumi-aws/issues/6208

shansmith01:
Add `ap-southeast-6` (Auckland) to the `supportedRegions` map in `ssr-site.ts`.

AWS launched this region in 2025; without an entry, multi-region SSR site routing metadata cannot record coordinates for server Lambdas deployed there, and region validation can fail depending on the Pulumi AWS provider version.

Fixes https://github.com/anomalyco/sst/issues/6791 on the SST side.

Related: https://github.com/pulumi/pulumi-aws/issues/6208 (Pulumi `Region` enum may need updating separately for full support).

### DIFF
--- a/platform/src/components/aws/ssr-site.ts
+++ b/platform/src/components/aws/ssr-site.ts
@@ -61,6 +61,7 @@ const supportedRegions = {
   "ap-southeast-3": { lat: -6.2088, lon: 106.8456 }, // Jakarta, Indonesia
   "ap-southeast-4": { lat: -37.8136, lon: 144.9631 }, // Melbourne, Australia
   "ap-southeast-5": { lat: 3.139, lon: 101.6869 }, // Kuala Lumpur, Malaysia
+  "ap-southeast-6": { lat: -36.8485, lon: 174.7633 }, // Auckland, New Zealand
   "ap-southeast-7": { lat: 13.7563, lon: 100.5018 }, // Bangkok, Thailand
   "ap-south-1": { lat: 19.076, lon: 72.8777 }, // Mumbai, India
   "ap-south-2": { lat: 17.385, lon: 78.4867 }, // Hyderabad, India


vimtor:
we probably need to wait for the pulumi issue to get fixed

have you tried this branch? does it work?

shansmith01:
Only if I patch the pulumi file as well

omercnet (changes requested):
The coordinate entry is correct, but this PR does not yet fix #6791 with the provider version this repository ships. `normalizeRegions()` validates every requested region with `Object.values(Region)` before `supportedRegions` is read, and `platform/package.json` pins `@pulumi/aws` 7.20.0, whose `Region` enum does not contain `ap-southeast-6`. A focused probe against the installed dependency returned `providerAccepts: false`, so users still receive `Invalid AWS region: "ap-southeast-6"` at `ssr-site.ts:1201-1204` and never reach this new coordinate. Pulumi added the enum in `@pulumi/aws` 7.43.0. Please update `@pulumi/aws` to 7.43.0 or newer, including the lockfile, or rebase this change onto a separate provider upgrade; then verify `normalizeRegions()` admits `ap-southeast-6`. The Auckland coordinates themselves are correct.

Issue #6848, opened 2026-04-30
sst.aws.StaticSite errorPage throwing a 502 without index set


nicholasgubbins:
using the current set up to run a static site.

```
export const docs = new sst.aws.StaticSite("Docs", {
    domain: "docs." + domain,
    path: "./packages/docs",
    dev: false,
    build: {
        command: "bun run build",
        output: "./dist",
    },
    errorPage: "/404.html",
});
```

i'm running a vitepress site. sst v4.5.1

if i go to docs.domain/404.html i get the page and can verify it's there

however if i go to any actual 404 route, i get a 502 ERROR Failed to contact the origin.

This makes sense because the errorPage should be throwing only in the instance of a 403 or 404 from the origin, but why is the cloudfront function throwing a 502 from a simple s3 file fetch?

nicholasgubbins:
claude analysis: 

 Now I have a clear picture of the issue. Let me trace through the relevant code path.            
                                                                     
  When errorPage is set in SST, the metadata stored in KV is:                                      
  { "errorResponseCode": 404, "s3": { ... } }                                                      
                                                                                                   
  Note: custom404 is explicitly set to undefined (so it's omitted from JSON) when errorPage is     
  truthy (line 1051 in static-site.ts):                                                            
  custom404: hasErrorPage ? undefined : errorPage,
                                                                                                   
  This means in the function:                                                                      
  - metadata.custom404 → undefined
  - metadata.errorResponseCode → 404                                                               
                                    
  The condition if(metadata.custom404 && !metadata.errorResponseCode) is always false when         
  errorPage is set — the function never rewrites the URL to the error page. Instead it falls       
  through to the S3 fallback (line 2791), which routes the original (non-existent) URI to S3. S3   
  returns a 403, and the function is relying entirely on CloudFront's distribution-level           
  customErrorResponses to intercept that 403 and serve the error page.

  The 502 cause: The customErrorResponses makes a NEW origin request for /404.html, which goes back
   through this viewer request function. At that point metadata is re-read from KV. If that KV read
   fails (transient error, or the key was temporarily absent during redeployment), metadata is     
  undefined, and this line in routeSite throws immediately:

  const baselessUri = metadata.base ? ... : event.request.uri;
  //                  ^^^^^^^^^^^^^ TypeError: Cannot read properties of undefined
                                                                                                   
  This uncaught exception causes CloudFront to emit the 502 — and it's especially likely on the    
[...]

PR #6849, opened 2026-05-01
fix(cf-function): guard against null metadata to prevent 502s

nicholasgubbins:
## Summary

- When the KV store returns `null`/`undefined` for the metadata key — which can happen transiently as the KV entry propagates to edge nodes after a deployment — `routeSite` throws an uncaught `TypeError` accessing `metadata.base`, causing CloudFront to return a 502.
- The same guard is applied to both `static-site.ts` and `ssr-site.ts`.
- The fix fails open: if metadata is unavailable, the request is passed through to the default origin unchanged rather than crashing.

## Root cause

The issue is most visible when `errorPage` is configured on a `StaticSite`. When `errorPage` is set, SST stores `errorResponseCode: 404` in the KV metadata and omits `custom404`, disabling the function-level SPA fallback in favour of distribution-level `customErrorResponses`. This means unmatched requests go through an extra round-trip (original request → S3 403 → CloudFront fetches error page → function invoked again), doubling the exposure to the KV propagation window and making the 502 appear inconsistently depending on which edge node handles the request.

## Test plan

- [ ] Deploy a `StaticSite` with `errorPage` set, request a non-existent path — should return 404 with the error page, not 502
- [ ] Simulate a missing metadata key (e.g. delete the KV entry temporarily) — should pass through to origin rather than 502

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/platform/src/components/aws/ssr-site.ts
+++ b/platform/src/components/aws/ssr-site.ts
@@ -943,8 +943,13 @@ async function handler(event) {
     metadata = JSON.parse(v);
   } catch (e) {}
 
-  const response = await routeSite(kvNamespace, metadata);
-  return response || event.request;
+  if (!metadata) return event.request;
+  try {
+    const response = await routeSite(kvNamespace, metadata);
+    return response || event.request;
+  } catch (e) {
+    return event.request;
+  }
 }`,
           },
           { parent: self },
--- a/platform/src/components/aws/static-site.ts
+++ b/platform/src/components/aws/static-site.ts
@@ -1129,8 +1129,13 @@ async function handler(event) {
     metadata = JSON.parse(v);
   } catch (e) {}
 
-  const response = await routeSite(kvNamespace, metadata);
-  return response || event.request;
+  if (!metadata) return event.request;
+  try {
+    const response = await routeSite(kvNamespace, metadata);
+    return response || event.request;
+  } catch (e) {
+    return event.request;
+  }
 }`,
           },
           { parent: self },


nicholasgubbins:
1. You request a non-existent path                                                                                                        
  2. Function runs → KV lookup succeeds → sets S3 origin → S3 returns 403/404                                                               
  3. CloudFront's customErrorResponses fires → second function invocation for /404.html                                                     
  4. That second invocation hits a cold edge that hasn't cached the routing metadata yet → KV read fails → returns event.request →
  placeholder.sst.dev → 502                                                                                                                 
                                                                                                                                          
  The second invocation is essentially always a "fresh" request from CloudFront's perspective — it's not the same cache key as the original 
  request, so it bypasses whatever warming the edge node had. Valid pages don't go through this second invocation at all, so you only see 
  the cold-edge failure on 404s.

nicholasgubbins:
this is the issue i faced yesterday with #6848

vimtor (commented):
i don’t think this fully fixes the 502. when metadata is missing, returning `event.request` still sends the request to the default origin, which is `placeholder.sst.dev`

for standalone sites, that origin does not resolve, so CloudFront still returns `502 Failed to contact the origin` instead of the intended site/error response

ekaya97:
i monkey patched this successfully in static-site.ts - same logic applies to ssr-site.ts as well.  

Replaced the placeholder origin with S3 + OAC in createDistribution() (starting at line 1168):

**Before:**
```typescript
    function createDistribution() {
      return new Cdn(
        ...transform(
          args.transform?.cdn,
          `${name}Cdn`,
          {
            comment: `${name} site`,
            domain: args.domain,
            origins: [
              {
                originId: "default",
                domainName: "placeholder.sst.dev",
                customOriginConfig: {
                  httpPort: 80,
                  httpsPort: 443,
                  originProtocolPolicy: "https-only",
                  originReadTimeout: 20,
                  originSslProtocols: ["TLSv1.2"],
                },
              },
            ],
```

**After:**
```typescript
    function createDistribution() {
      const oac = new OriginAccessControl(
        `${name}OAC`,
        { name: physicalName(64, name) },
        { parent: self, ignoreChanges: ["name"] },
      );

      return new Cdn(
        ...transform(
          args.transform?.cdn,
          `${name}Cdn`,
          {
            comment: `${name} site`,
            domain: args.domain,
            origins: [
              {
                originId: "default",
                domainName: bucketDomain,
                originAccessControlId: oac.id,
                s3OriginConfig: {
                  originAccessIdentity: "",
                },
              },
            ],
```

marque-b:
Confirmed on SST 4.15.2 with StaticSite + errorPage: "/404.html". Without a fix, missing routes returned 502 on the customErrorResponses sub-request when the viewer-request function fell through to placeholder.sst.dev.

Workaround that worked for me: explicit sst.aws.Bucket + transform.cdn to set the default origin to S3 + OAC (per @ekaya97's approach in #6849). OriginAccessControl had to be created at stack scope, not under StaticSite, due to SST component naming rules.

PR #6854, opened 2026-05-02
Emphasize `npm install` in docs

### LINKED ISSUE #6851: Improve docs to focus more on `npm install`
since SST should be used in node contexts, we probably should give more priority to the docs for doing npm install SST instead of the alternatives

iamujjwalsinha:
## Summary
- emphasize local `npm install sst` for Node projects
- show running SST through the project package manager with `npx sst dev`
- keep the global install path as the fallback for non-Node projects

## Testing
- git diff --check

Closes #6851

### DIFF
--- a/www/src/content/docs/docs/index.mdx
+++ b/www/src/content/docs/docs/index.mdx
@@ -268,19 +268,57 @@ Learn more about our [monorepo setup](/docs/set-up-a-monorepo/).
 
 ## CLI
 
-To make this all work, SST comes with a [CLI](/docs/reference/cli/). For Node projects, install SST locally in your project so the CLI version is tracked with your app.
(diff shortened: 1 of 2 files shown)


www/src/content/docs/docs/index.mdx
```diff
@@ -268,13 +268,19 @@ Learn more about our [monorepo setup](/docs/set-up-a-monorepo/).
...
+To make this all work, SST comes with a [CLI](/docs/reference/cli/). For Node projects, install SST locally in your project so the CLI version is tracked with your app.
 
 ```bash
 npm install sst
 ```
 
-Or if you are not using Node, you can install it globally.
+You can then run the CLI with your package manager.
+
+```bash
```
vimtor: thanks for your contribution!

can we check whether the docs support multiple tabs? maybe the starlight framework we're using already supports this, but i feel like it would be helpful to show how to run sst with pnpm, bun, and yarn

PR #6857, opened 2026-05-04
Encrypt bootstrap SSM parameter to survive org re-encryption policies

### LINKED ISSUE #6856: Some orgs force SSM Parameter encryption which breaks the bootstrap version check
# Encrypt bootstrap SSM parameter to survive AWS organization re-encryption policies

## Problem

SST stores its bootstrap version data in an SSM parameter (`/sst/bootstrap`) as a plain `String` type. Some AWS organizations enforce policies that automatically re-encrypt unencrypted SSM parameters to `SecureString`. Once this happens, SST can no longer read the bootstrap parameter because it calls `GetParameter` with `WithDecryption: false`, which returns the encrypted ciphertext instead of the plaintext JSON.

## Proposed Fix

Two changes in `pkg/project/provider/aws.go` inside the `Bootstrap()` method:

1. **Read**: Change `WithDecryption` from `false` to `true` on the `GetParameter` call for `/sst/bootstrap`.
2. **Write**: Change the `PutParameter` type from `ParameterTypeString` to `ParameterTypeSecureString`.

This matches the pattern already used by the passphrase parameter (`getPassphrase` / `setPassphrase` in the same file).

## Backward Compatibility

This is safe for existing deployments with unencrypted `String` bootstrap parameters:

- **Reading**: The AWS docs explicitly state that `WithDecryption` is ignored for `String` and `StringList` parameter types — it only applies to `SecureString`. So existing unencrypted parameters are returned identically.
  - Reference: https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameter.html
  - *"Return decrypted values for secure string parameters. This flag is ignored for `String` and `StringList` parameter types."*

- **Writing**: The existing parameter is only rewritten when the bootstrap version advances. At that point, `PutParameter` with `Overwrite: true` allows changing the type from `String` to `SecureString` seamlessly.

HeathHopkins:
Closes #6856

## Summary

- Change the `/sst/bootstrap` SSM parameter from `String` to `SecureString` and read it with `WithDecryption: true`
- Prevents breakage when an AWS organization policy automatically re-encrypts unencrypted SSM parameters
- Matches the pattern already used by the passphrase parameter in the same file

## Motivation

Some AWS organizations enforce policies that automatically convert unencrypted SSM parameters to `SecureString`. Once this happens, the bootstrap parameter becomes unreadable because SST calls `GetParameter` with `WithDecryption: false`, which returns encrypted ciphertext instead of the plaintext JSON.

## Changes

Two changes in `pkg/project/provider/aws.go` inside `Bootstrap()`:

1. `GetParameter` for `/sst/bootstrap`: `WithDecryption` changed from `false` to `true`
2. `PutParameter` for `/sst/bootstrap`: `Type` changed from `ParameterTypeString` to `ParameterTypeSecureString`

## Backward Compatibility

This is safe for existing deployments with unencrypted `String` bootstrap parameters:

- **Reading**: From the [AWS docs](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameter.html), `WithDecryption` *"is ignored for `String` and `StringList` parameter types."* Existing unencrypted parameters are returned identically.
- **Writing**: The parameter is only rewritten when the bootstrap version advances. At that point, `PutParameter` with `Overwrite: true` allows changing the type from `String` to `SecureString` seamlessly.

### DIFF
--- a/pkg/project/provider/aws.go
+++ b/pkg/project/provider/aws.go
@@ -154,7 +154,7 @@ func (p *AwsProvider) Bootstrap(region string) (*AwsBootstrapData, error) {
 	slog.Info("fetching bootstrap")
 	result, err := ssmClient.GetParameter(ctx, &ssm.GetParameterInput{
 		Name:           aws.String(SSM_NAME_BOOTSTRAP),
-		WithDecryption: aws.Bool(false),
+		WithDecryption: aws.Bool(true),
 	})
 	if result != nil && result.Parameter.Value != nil {
 		slog.Info("found existing bootstrap", "data", *result.Parameter.Value)
@@ -187,7 +187,7 @@ func (p *AwsProvider) Bootstrap(region string) (*AwsBootstrapData, error) {
 			ctx,
 			&ssm.PutParameterInput{
 				Name:      aws.String(SSM_NAME_BOOTSTRAP),
-				Type:      ssmTypes.ParameterTypeString,
+				Type:      ssmTypes.ParameterTypeSecureString,
 				Overwrite: aws.Bool(true),
 				Value:     aws.String(string(data)),
 			},


vimtor:
i don't think this is backward-compatible

`WithDecryption: true` is safe for reads, but `PutParameter` cannot change an existing parameter from `String` to `SecureString`; AWS returns `HierarchyTypeMismatchException`:  
https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_PutParameter.html#API_PutParameter_Errors

so existing accounts with `/sst/bootstrap` as `String` may fail on the next bootstrap rewrite

also, `SecureString` needs KMS permissions (`kms:Encrypt`/`kms:Decrypt`):  
https://docs.aws.amazon.com/systems-manager/latest/userguide/secure-string-parameter-kms-encryption.html#parameter-policy-kms-encryption

that matters for SST because our least-privilege bootstrap policy in `www/src/content/docs/docs/iam-credentials.mdx` only grants SSM access for `/sst/bootstrap`, so this could break users following our IAM docs unless we update it too

PR #6858, opened 2026-05-04
Publish Python SDK to PyPI as sst-sdk

subssn21:
## Summary

Publishes the SST Python SDK to PyPI as `sst-sdk`, so users can install with `pip install sst-sdk` or `uv add sst-sdk` instead of pointing at the git repo.

## Changes (5 files)

- **`sdk/python/pyproject.toml`** — Rename to `sst-sdk`, add PyPI metadata, hatch build config
- **`sdk/python/README.md`** — Installation, migration guide, usage examples
- **`sdk/python/PUBLISHING.md`** — One-time PyPI trusted publishing setup docs
- **`.github/workflows/release.yml`** — Add Python SDK publish steps (after Rust, before Discord)
- **`.github/workflows/check.yml`** — Pass GITHUB_TOKEN to docs build (fixes rate limit)

## Migration for users

```toml
# Before
[project]
dependencies = ["sst"]

[tool.uv.sources]
sst = { git = "https://github.com/sst/sst", subdirectory = "sdk/python" }

# After
[project]
dependencies = ["sst-sdk"]
```

No code changes — `from sst import Resource` works the same.

## Setup required before first publish

Configure trusted publishing on PyPI for `sst-sdk` (details in `sdk/python/PUBLISHING.md`).


.github/workflows/release.yml
```diff
@@ -105,6 +105,18 @@ jobs:
         env:
           CARGO_REGISTRY_TOKEN: ${{ steps.crates-auth.outputs.token }}
 
+      - name: Release Python SDK
```
vimtor: is "Release" here correct? aren't we building the sdk? not sure

sdk/python/pyproject.toml
```diff
@@ -1,15 +1,41 @@
...
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Topic :: Software Development :: Libraries",
+]
 dependencies = ["pycryptodomex==3.20.0"]
 
+[project.urls]
+Homepage = "https://sst.dev"
+Documentation = "https://sst.dev/docs/reference/sdk/"
+Repository = "https://github.com/sst/sst"
```
vimtor: we should update this URLs to use anomalyco in the org name

sdk/python/pyproject.toml
```diff
@@ -1,15 +1,41 @@
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Topic :: Software Development :: Libraries",
+]
 dependencies = ["pycryptodomex==3.20.0"]
 
+[project.urls]
+Homepage = "https://sst.dev"
+Documentation = "https://sst.dev/docs/reference/sdk/"
```
vimtor: maybe we can add here sdk#python so it links directly?

sdk/python/README.md
```diff
@@ -1,11 +1,76 @@
...
+pip install sst-sdk
+```
+
+Or with uv:
+
+```bash
+uv add sst-sdk
+```
+
+> **Note**: When deploying with SST, the SDK is automatically included — you don't need to install it manually. This package is for local development and testing.
```
vimtor: is this true? i don't see any changes made to the function or python runtime

sdk/python/README.md
```diff
@@ -1,11 +1,76 @@
...
+## Supported Python Versions
+
+- Python 3.9+
+
+## Links
+
+- [SST Documentation](https://sst.dev/docs/)
+- [SDK Reference](https://sst.dev/docs/reference/sdk/)
+- [Python on SST](https://sst.dev/docs/examples/#aws-lambda-python)
+- [GitHub](https://github.com/sst/sst)
```
vimtor: same thing here with the URLs

vimtor (changes requested):

sdk/python/README.md
vimtor: also if this is the case, why would people need to do `uv add sst-sdk`?

.github/workflows/release.yml
subssn21: Not really building as there's nothing to compile or anything. It's just bundling the files and pushing it to pypi. Don't think it matters, whichever you want.

PR #6859, opened 2026-05-06
Add versioned Lambda asset uploads

iloewensen:
## Summary

Adds an opt-in Lambda code asset path controlled by `SST_LAMBDA_CODE_ASSET_VERSIONING`.

When enabled, SST uploads Lambda zip assets to a stable S3 key and updates the existing `BucketObjectv2` using `sourceHash`. Lambda functions then reference the uploaded object version through `s3ObjectVersion`.

## Motivation

Large deployments with many Lambda functions currently create a new S3 object key for every function code change. In testing with 250 functions, this caused significant S3 object churn and larger Pulumi state updates.

Measured results from the investigation:

- Current behavior, 250 functions: ~4m22s
- Versioned Lambda assets, 250 functions: ~2m43s
- No code changes: ~13s

CloudTrail and Pulumi debug logs did not show AWS throttling as the primary bottleneck. The remaining cost appears to be Lambda control-plane update latency across many functions.

## Changes

- Adds `SST_LAMBDA_CODE_ASSET_VERSIONING=1` / `true`.
- Uses stable Lambda asset keys under `lambda/<app>/<stage>/<function>/code.zip`.
- Sets `sourceHash` on the `BucketObjectv2` so Pulumi updates the same object when code changes.
- Passes `s3ObjectVersion` to Lambda so code updates point at the exact uploaded object version.
- Enables versioning on the SST asset bucket during AWS bootstrap.
- Adds lifecycle rules to keep Lambda asset versions bounded.

## Compatibility

Default behavior is unchanged when `SST_LAMBDA_CODE_ASSET_VERSIONING` is not set.

Dev bridge code keeps the existing hashed `assets/dev-bridge-code-<hash>.zip` behavior.


iloewensen:
@vimtor What do you think about using versioned S3 bucket objects for Lambda assets instead of creating and deleting separate bucket objects on each code change? If this makes sense, should we explore this approach further?

iloewensen:
@vimtor I would appreciate your feedback.

PR #6864, opened 2026-05-10
Precompile Python bytecode to reduce Lambda cold starts

subssn21:
## Summary

Precompiles Python bytecode (.pyc) during the build to eliminate runtime compilation on Lambda cold starts. Reduces cold start time by 5-8 seconds for functions with large dependency trees.

## Problem

Without precompiled .pyc files, Lambda must compile every .py file on first import. For a function importing elasticsearch + aiohttp + boto3, this adds 5-8s of pure `compile()` overhead to every cold start.

## Solution

- Run `python -m compileall` with `--invalidation-mode checked-hash` (validates by source hash, not mtime — avoids stale detection after zip extraction)
- Use `-s`/`-p` flags to rewrite source paths to `/var/task/` (matches Lambda runtime)
- Compile deps **once** in the shared cache, then copy .pyc files through to each function
- Compile function source per-function (fast — just your code)
- Stop stripping `__pycache__`/`.pyc` from installed dependencies

## Changes (2 files)

- **`pkg/runtime/python/build.go`** — Add `precompilePythonFiles`, stop deleting .pyc in cleanup, stop skipping `__pycache__` in `copyDependencyPackages`
- **`pkg/runtime/python/build_test.go`** — Update cleanup test expectations
- **`pkg/runtime/python/python.go`** — Stop filtering .pyc/`__pycache__` in `skipContent`

## Performance

| Metric | Before | After |
|--------|--------|-------|
| Cold start (large deps) | ~14s | ~6s |
| Build time (first function) | baseline | +1-2s |
| Build time (subsequent) | baseline | negligible |
| Package size | baseline | +10-20% |

## Testing

- All Go tests pass
- Verified .pyc files have correct `/var/task/` paths embedded
- Verified `checked-hash` invalidation mode in .pyc headers
- Tested real deploy on Lambda — cold start compile() calls eliminated

### DIFF
--- a/pkg/runtime/python/build.go
+++ b/pkg/runtime/python/build.go
@@ -90,6 +90,13 @@ func buildDeploy(ctx context.Context, input *runtime.BuildInput, cacheDir string
 		return nil, fmt.Errorf("dependency installation: %w", err)
 	}
 
+	// Precompile the function's own source files (deps are already compiled in the cache).
+	if !input.IsContainer {
+		if err := precompilePythonFiles(ctx, input, input.Out()); err != nil {
+			slog.Warn("failed to precompile Python source files", "error", err)
+		}
+	}
+
 	output, err := createFinalBuildOutput(input, projectInfo)
 	if err != nil {
 		return nil, fmt.Errorf("build output: %w", err)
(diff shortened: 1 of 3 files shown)


doron98:
How is that different from the [--compile-bytecode](https://docs.astral.sh/uv/reference/cli/#uv-pip-install--compile-bytecode) or [--compile-bytecode](https://docs.astral.sh/uv/reference/cli/#uv-sync--compile-bytecode) option of `uv sync` and `uv pip install` ?

subssn21:
> How is that different from the [--compile-bytecode](https://docs.astral.sh/uv/reference/cli/#uv-pip-install--compile-bytecode) or [--compile-bytecode](https://docs.astral.sh/uv/reference/cli/#uv-sync--compile-bytecode) option of `uv sync` and `uv pip install` ?

uv pip install --compile-bytecode uses timestamp-based invalidation and embeds the local build path in the .pyc files. Both of these break on Lambda:

Timestamp invalidation — after zip extraction, file mtimes change, so Python thinks the .pyc is stale and recompiles from source
Local paths — the .pyc embeds /home/runner/... but Lambda runs from /var/task/, so Python ignores the bytecode
Our compileall call uses --invalidation-mode checked-hash (validates by source hash, survives zip extraction) and -s <build_dir> -p /var/task (rewrites paths to match Lambda). These flags aren't available through uv's --compile-bytecode option.

PR #6865, opened 2026-05-11
Add `enabled` option to bus subscribers

yotamishak:
small dx improvement

i know i can use transform but llms hallucinate when it comes to this, thought it might be useful for my self and others


vimtor:
thanks for your contribution @yotamishak

PR #6866, opened 2026-05-12
Bucket eventbridge notifications

yotamishak:
added support for the eventbridge notification + abstraction on bucket to subscribe to these events directly

main "limitation" on aws side, eventbridge notifications need to be configure together direct sqs/sns/lambda notification in order to prevent overriding bucket configuration

leaving comments on questionable decisions
let me know if anything needs fixing/or drop abstractions

### DIFF
--- a/platform/src/components/aws/bucket.ts
+++ b/platform/src/components/aws/bucket.ts
@@ -733,17 +733,6 @@ export interface BucketNotificationsArgs {
   };
 }
 
-type BucketEventBridgeDetailType =
-  | "Object Created"
-  | "Object Deleted"
-  | "Object Restore Initiated"
-  | "Object Restore Completed"
-  | "Object Restore Expired"
-  | "Object Tags Added"
-  | "Object Tags Deleted"
-  | "Object ACL Updated"
-  | (string & {});
-
 export interface BucketEventBridgeSubscriberArgs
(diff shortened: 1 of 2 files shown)


platform/src/components/aws/bucket.ts
```diff
@@ -532,10 +535,29 @@ export interface BucketArgs {
...
+   *   ]
+   * }
+   * ```
+   */
+  eventBridge?: Input<boolean>;
   /**
    * A list of subscribers that'll be notified when events happen in the bucket.
    */
-  notifications: Input<
+  notifications?: Input<
```
yotamishak, line 560: changed this to optional, allowing only to pass eventbridge true

defaults to empty array uptop

platform/src/components/aws/bucket.ts
```diff
@@ -711,6 +733,54 @@ export interface BucketNotificationsArgs {
...
+      Input<
+        | "Object Created"
+        | "Object Deleted"
+        | "Object Restore Initiated"
+        | "Object Restore Completed"
+        | "Object Restore Expired"
+        | "Object Tags Added"
+        | "Object Tags Deleted"
+        | "Object ACL Updated"
+        | string
```
yotamishak, line 763: should be closed list, but left room for error just incase

PR #6871, opened 2026-05-14
Fix Python SDK source names in examples

### LINKED ISSUE #6868: Python examples are outdated after sst-sdk update
SST version 4.2.7

After recent changes to the Python SDK, `uv` dependencies commands fail with:

```
  × Failed to download and build `sst @
  │ git+https://github.com/sst/sst.git@dev#subdirectory=sdk/python`
  ╰─▶ Package metadata name `sst-sdk` does not match given name `sst`
```

My file pyproject.toml has:

```
dependencies = [
    "sst",
...

[tool.uv.sources]
sst = { git = "https://github.com/sst/sst.git", branch = "dev", subdirectory = "sdk/python" }
```

from this example: https://github.com/anomalyco/sst/blob/dev/examples/aws-python/pyproject.toml 

And thus auto deploy fails.

Migration docs exist in the PyPi package https://pypi.org/project/sst-sdk/ but the official examples have not been updated yet.

asim48-ctrl:
Fixes #6868.

The Python SDK package metadata now uses `sst-sdk`, but several example `pyproject.toml` files still depended on/source-mapped it as `sst`. `uv` rejects that with a package metadata name mismatch.

This updates the affected Python examples so dependency names and `[tool.uv.sources]` keys match `sst-sdk` while keeping the package source path unchanged.

Validation:
- `uv lock --dry-run` in `examples/aws-python`
- `uv lock --dry-run` in `examples/aws-fastapi`
- `git diff --check`

### DIFF
--- a/examples/aws-fastapi/functions/pyproject.toml
+++ b/examples/aws-fastapi/functions/pyproject.toml
@@ -11,4 +11,4 @@ build-backend = "hatchling.build"
 
 [tool.uv.sources]
 core = { workspace = true }
-sst-sdk = { git = "https://github.com/sst/sst.git", branch = "dev", subdirectory = "sdk/python" }
+sst-sdk = { git = "https://github.com/anomalyco/sst.git", branch = "dev", subdirectory = "sdk/python" }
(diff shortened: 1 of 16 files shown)


vimtor:
nice, could you also replace sst/sst with anomalyco/sst? we need to update it in the Python SDK readme as well

Issue #6886, opened 2026-05-23
`sst dev` should watch `.env` files and reload env vars without restart


guidefari:
## Problem

Changes to `.env` or `.env.<stage>` do not take effect while `sst dev` is running.

Today the dev loop is:

1. Edit `.env`
2. Stop `sst dev`
3. Start `sst dev` again
4. Wait for the app to redeploy

It would be useful if SST could watch env files and trigger a redeploy automatically, the same way it already does for source/config changes.

## Example

```ts
export default $config({
  app(input) {
    return {
      name: "my-app",
      removal: input?.stage === "production" ? "retain" : "remove",
      home: "aws",
      watch: {
        env: true,
      },
    };
  },
});
```

Then changing either of these files should trigger a redeploy:

```txt
.env
.env.dev
```

The next Pulumi/SST run should receive the updated env values.

## Proposed API

Add an optional `env` flag to `app.watch`:

```ts
{
  app: {
    watch: {
      env: true,
    },
  },
}
```

Default should probably be `false` to avoid changing existing behaviour.

## Why this would help

This removes a common restart/redeploy cycle during local development.

For apps that use env vars in `sst.config.ts` or component config, tweaking local config/secrets would become part of the normal `sst dev` feedback loop instead of requiring a manual restart.

ekaya97:
SST - by philosophy - doesn't like .env files. For secrets, you use the Secret component, for public keys you use the Linkable component. It's much much much better to handle in multi-stage/multi-dev setups.

I always have a script that auto-syncs from .env files to Secret, that way I still have my .env file for reference (particularly useful if the SST pattern doesn't feel right yet + you don't have to set dozens of keys manually for a new stage). Linkables hot-reload anyway if you drive config via env

if you structure your env variables (Secrets + Linkables) such that you can package them for each consumer, you can simply link them to your component in decomposition - that way you don't have to add or remove an env variable through the entire chain.

PR #6887, opened 2026-05-23
feat: sst dev watches .env files and reloads env vars

guidefari:
Addresses #6886 

## Notes
- Leaving as draft till we're happy with the proposed api/behaviour

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ Pulumi.yaml
 tmp
 
 # coding agents
+.bookgen-notes
 .termai
 .opencode
 .idea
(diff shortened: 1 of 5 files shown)


vimtor:
this should probably be way simpler. by default, `.env` files are ignored because we are ignoring any files that start with a dot

we should just allow them

guidefari:
hey @vimtor, looked into this a bit more. check the log I added, updating my `.env` file fired off the log

```
time=2026-05-31T00:36:09.193+02:00 level=INFO msg="file change deploy decision" service=deployer path=/abcs/gbfm/.env tracked=false
```

The dot-name skip I found is in `shouldSkipDir`, and it returns early for non-directories.

Looks like a redeploy isn't being triggered because `.env` isn't part of `watchedFiles` ?

Issue #6677, opened 2026-04-01
`ApiGatewayWebSocket` re-creates `APIGatewayAccountSetup` and IAM role on every `sst dev`


IshanArya:
## Description

Every time I run `sst dev`, the `ApiGatewayWebSocket` component creates a brand new IAM role and `apigateway.Account` resource, even though one was already created in the previous session:

```
⠼  Creating    APIGatewayPushToCloudWatchLogsRole aws:iam:Role
⠼  Creating    WsApiAPIGatewayAccountSetup aws:apigateway:Account
|  Created     APIGatewayPushToCloudWatchLogsRole aws:iam:Role
|  Created     WsApiAPIGatewayAccountSetup aws:apigateway:Account
```

This happens on every single `sst dev` startup. It doesn't happen with `sst.aws.React` (HTTP API) or `sst.aws.Function` with `url: true` — only with `ApiGatewayWebSocket`.

## Config

Minimal reproduction:

```ts
const wsApi = new sst.aws.ApiGatewayWebSocket("WsApi");

wsApi.route("$connect", { handler: "./functions/ws/connect.handler" });
wsApi.route("$disconnect", { handler: "./functions/ws/disconnect.handler" });
wsApi.route("$default", { handler: "./functions/ws/default.handler" });
```

## Root cause

The helper `setupApiGatewayAccount` in `components/aws/helpers/apigateway-account.ts` does the following:

1. Reads the current API Gateway account-level config via `apigateway.Account.get()`
2. Checks if `cloudwatchRoleArn` is set
3. If empty, creates a **new** `iam.Role` (with a random Pulumi-generated suffix) and a new `apigateway.Account` to set the CloudWatch role ARN

The problem is that Pulumi generates a unique suffix on the IAM role name each session (e.g., `APIGatewayPushToCloudWatchLogsRole-cb7dd1b`, then `-7ec4298` the next time). Between `sst dev` sessions, the account-level `cloudwatchRoleArn` is empty or stale, so the check on line 20 fails and the entire setup runs again — creating yet another orphaned IAM role (with `retainOnDelete: true`, so they pile up in the AWS account).

## Suggested fix

Since the API Gateway account config is a **per-AWS-account, per-region singleton**, the IAM role should have a stable, deterministic name rather than a Pulumi-auto-generated one. For example:

```ts
const role = new iam.Role(
  "APIGatewayPushToCloudWatchLogsRole",
  {
[...]

PR #6911, opened 2026-06-11
Fix duplicate CloudWatch role URN on fresh accounts

### LINKED ISSUE #6199: Duplicate resource URN 'urn:pulumi:stage::app::aws:iam/role:Role::APIGatewayPushToCloudWatchLogsRole'
When I have two API gateways (sst.aws.ApiGatewayV1) in a project I get an error `Duplicate resource URN 'urn:pulumi:stage::app::aws:iam/role:Role::APIGatewayPushToCloudWatchLogsRole'; try giving it a unique name`.

I have one public API gateway and another private API gateway. If comment out either one all is good, but when both are defined SST/Pulumi will fail with error mentioned above.

jasonrowsell:
Fixes #6199, likely #6677 too. I was also getting this error

On a fresh AWS account `cloudwatchRoleArn` is unset, so the guard in `setupApiGatewayAccount` never short-circuits and every gateway registers an `iam.Role` under the same logical name — two or more gateways and the deploy fails with `Duplicate resource URN`. It dies before the ARN is written back, so retries hit the same race. Any account that ever completed a single-gateway deploy short-circuits, which is why this has been hard to reproduce.

Commit 1 registers the role once per process, same as the `LambdaEncryptionKey` dedupe in `function.ts`. No logical names change, so existing stacks see no churn.

Commit 2: since terraform-provider-aws v6, destroying `aws_api_gateway_account` [resets `cloudwatchRoleArn` to null](https://github.com/hashicorp/terraform-provider-aws/blob/v6.33.0/internal/service/apigateway/account.go#L218-L239). The setup only exists in the program while the ARN is unset, so bootstrap never converges: set, reset on the next deploy, bootstrap again with a fresh orphan role — the pile-up in #6677. `retainOnDelete: true` keeps the account-wide setting in place and the cycle converges after the first deploy.

### DIFF
--- a/platform/src/components/aws/helpers/apigateway-account.ts
+++ b/platform/src/components/aws/helpers/apigateway-account.ts
@@ -53,7 +53,7 @@ export function setupApiGatewayAccount(
       {
         cloudwatchRoleArn: useCloudWatchRole(opts).arn,
       },
-      { provider: opts.provider },
+      { retainOnDelete: true, provider: opts.provider },
     );
   });
 }
--- a/platform/src/components/aws/helpers/apigateway-account.ts
+++ b/platform/src/components/aws/helpers/apigateway-account.ts
@@ -6,6 +6,34 @@ import {
 } from "@pulumi/pulumi";
 import { $print } from "../../component";
 
+let cloudWatchRole: iam.Role | undefined;
+
+function useCloudWatchRole(opts: ComponentResourceOptions) {
+  const partition = getPartitionOutput(undefined, opts).partition;
+  cloudWatchRole ??= new iam.Role(
+    `APIGatewayPushToCloudWatchLogsRole`,
+    {
+      assumeRolePolicy: jsonStringify({
+        Version: "2012-10-17",
+        Statement: [
+          {
+            Effect: "Allow",
+            Principal: {
+              Service: "apigateway.amazonaws.com",
+            },
+            Action: "sts:AssumeRole",
+          },
+        ],
+      }),
+      managedPolicyArns: [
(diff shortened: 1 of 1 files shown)


platform/src/components/aws/helpers/apigateway-account.ts
```diff
@@ -20,35 +48,12 @@ export function setupApiGatewayAccount(
...
-    );
-
     return new apigateway.Account(
       `${namePrefix}APIGatewayAccountSetup`,
       {
-        cloudwatchRoleArn: role.arn,
+        cloudwatchRoleArn: useCloudWatchRole(opts).arn,
       },
-      { provider: opts.provider },
+      { retainOnDelete: true, provider: opts.provider },
```
jasonrowsell, line 56: Calling this one out since it's the only behavioral change for existing users: since terraform-provider-aws v6, destroying `aws_api_gateway_account` resets `cloudwatchRoleArn` to null ([account.go](https://github.com/hashicorp/terraform-provider-aws/blob/v6.33.0/internal/service/apigateway/account.go#L218-L239)), and the setting is account/region-wide. Without `retainOnDelete` here, removing one app resets the role for every other stack in the account — and it's also what stops the bootstrap from converging (#6677): the setup only exists in the program while the ARN is unset, so it gets destroyed on the deploy after bootstrap, the destroy resets the ARN, and the next deploy creates another orphan role.

The trade-off is that a full teardown now leaves the account setting in place, pointing at the (already retained) role. That seemed strictly better than resetting shared account state on `sst remove`.

PR #6925, opened 2026-06-24
Add `state.retention` option to prevent big state buckets

vimtor:
## Summary

- add `state.retention` to keep the latest completed state snapshots
- prune matching snapshot, event log, update, and summary history across AWS, Cloudflare, and local homes
- permanently remove pruned S3 versions while preserving current state, secrets, and retained snapshots

## Behavior

```ts
state: {
  retention: 30
}
```

Retention counts completed updates, not days. Omitted retention keeps all history. Pruning runs after an update and failures are non-fatal.

On AWS, retained snapshots become recovery history; noncurrent `app` checkpoints and stale lock versions are removed. On Cloudflare R2 and local homes, stale history objects are deleted explicitly.

## Tests

- `bun run test:cli`
- `go test ./pkg/project/...`
- `bun run typecheck`
- `bun run build:cli`
- `git diff --check`

`bun run build:platform` completed TypeScript bundling, but local Docker reported that its driver does not support multi-platform builds; command exited 0.

Supersedes #6669.

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -47,8 +47,16 @@ type App struct {
 }
 
 type State struct {
-	Purge    bool `json:"purge"`
-	Compress bool `json:"compress"`
+	Purge     bool `json:"purge"`
+	Compress  bool `json:"compress"`
+	Retention *int `json:"retention"`
+}
+
+func validateState(state *State) error {
+	if state != nil && state.Retention != nil && *state.Retention < 1 {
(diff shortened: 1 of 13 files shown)


mclean25:
@vimtor just finished deploying to my test stage I've been using for a long time and it seems to have worked quite well. I got Codex to summarize the before/after and logs in case any of this is useful for you:

CleanShot 2026-06-24 at 16 14 35@2x

Issue #6923, opened 2026-06-22
409 on Lambda Permission, too many concurrent permissions updates with Function URLS


imaginaryCorn:
Hello!

We're hitting an issue with the first time deploy of `sst.aws.React` on function url creation and the permissions created therein:

```
Mon, 22 Jun 2026 04:53:32 GMT   |  Created     ***Site sst:aws:React → ***SiteServerUswest2Function aws:lambda:Function (15.2s)
Mon, 22 Jun 2026 04:53:32 GMT   |  Created     ***Site sst:aws:React (19.8s)
Mon, 22 Jun 2026 04:53:32 GMT   |  Created     ***Site sst:aws:React → ***SiteServerUswest2InvokeFunction aws:lambda:Permission
Mon, 22 Jun 2026 04:53:32 GMT   |  Error       ***Site sst:aws:React → ***SiteServerUswest2Url aws:lambda:FunctionUrl
Mon, 22 Jun 2026 04:53:32 GMT   sdk-v2/provider2.go:572: sdk.helper_schema: adding Lambda Function URL (a-sst-dev-stack-*-
***SiteServerUswest2Function-**) permission operation error Lambda: AddPermission, https response error StatusCode: 409, RequestID: **, 
ResourceConflictException: The function could not be updated due to a concurrent update operation.: provider=aws@7.20.0
Mon, 22 Jun 2026 04:53:32 GMT   |  Created     ***Site sst:aws:React → ***SiteServerUswest2PublicFunctionUrlAccess aws:lambda:Permission
Mon, 22 Jun 2026 04:53:32 GMT   |  Created     ***Site sst:aws:React → ***SitePublicFunctionUrlAccessUswest2 aws:lambda:Permission

```

Appears that permissions creation are done asap in parallel for the Function, so CreateFunctionUrlConfig + [both AddPermission calls mutate](https://github.com/anomalyco/sst/blob/v4.14.3/platform/src/components/aws/function.ts#L2756) the same function's resource policy concurrently, causing 409 on occasion, especially for SST stacks created anew for development environments per PR.

Hjorthdx:
Seeing this on `sst.aws.Nextjs` too. Same 409 on the function URL permission when deploying per-PR preview stacks - intermittent, and re-running usually passes but not always. sst 4.14.3, aws provider 7.22.0.

```
catalog sst:aws:Nextjs → catalogServerEucentral1Url aws:lambda:FunctionUrl
[...]

PR #6933, opened 2026-07-09
Fix intermittent Lambda 409 when adding Function URL permissions

AppleDannyClegg:
# Fix intermittent Lambda 409 when adding Function URL permissions

## Problem

Deploying an SSR site (`Nextjs`, `Astro`, `SolidStart`, …) intermittently fails with:

```
operation error Lambda: AddPermission, https response error StatusCode: 409,
ResourceConflictException: The function could not be updated due to a concurrent
update operation.
```

It's non-deterministic and hits fresh/ephemeral stacks (e.g. per-PR preview
environments) far more often than re-deploys. Retrying the deploy succeeds.

## Cause

AWS Lambda serialises mutating control-plane operations per function: after any
update the function is briefly `LastUpdateStatus: InProgress`, and a second
mutating call in that window is rejected with a 409.

During a deploy two resources mutate the same function and, with no ordering
between them, Pulumi runs them concurrently:

- `aws.lambda.FunctionUrl` — created in `Function.createUrl()`
  (`CreateFunctionUrlConfig`), and
- `aws.lambda.Permission` with `action: "lambda:InvokeFunctionUrl"`.

For the OAC path the permission is created in `SsrSite` and depends only on the
function — not on the function URL. It can't depend on the URL because the
`FunctionUrl` resource isn't exposed on `Function.nodes` (only `role`,
`function`, `logGroup`, `eventInvokeConfig`). So `AddPermission` races
`CreateFunctionUrlConfig` and lands mid-update. Fresh stacks hit it more often
because the function and its URL are created together in a single run.

## Fix

Serialise the permission after the function URL:

- expose the `FunctionUrl` resource on `Function.nodes.url`, and
- add `dependsOn` on it to every `lambda.Permission` that grants URL access —
  both the ones created inside `Function.createUrl()` (public + router modes)
  and the OAC/CloudFront ones created in `SsrSite` (server functions and the
  image optimizer).

`dependsOn` only changes the Pulumi dependency graph, so existing permission
resources are updated in place — no replacement, no downtime.

## Testing

- `tsc --noEmit` passes.
- `platform` test suite: all component tests pass (the only failures are a
  pre-existing Node-version incompatibility in unrelated suites — `alb`,
  `bucket`, `service-alb` — that error on `@types/node` resolution /
  `ERR_TRACE_EVENTS_UNAVAILABLE`, unaffected by this change).
- Recommended before merge: a deploy of an SSR site in `oac` /
  `oac-with-edge-signing` and `none` protection modes (and a multi-region
  deploy) to confirm the 409 no longer occurs and that no permission resources
  are replaced on an in-place update.

## Notes / possible follow-ups

The same class of concurrent-mutation race can in principle affect other
per-function operations that run alongside URL/permission creation (e.g.
provisioned-concurrency config, or an environment update). Those are out of
scope here; happy to look at them separately if maintainers think it's worth it.

### DIFF
--- a/platform/src/components/aws/function.ts
+++ b/platform/src/components/aws/function.ts
@@ -1717,6 +1717,7 @@ export class Function extends Component implements Link.Linkable {
   private role: iam.Role;
   private logGroup: Output<cloudwatch.LogGroup | undefined>;
   private urlEndpoint: Output<string | undefined>;
+  private urlResource: Output<lambda.FunctionUrl | undefined>;
   private eventInvokeConfig?: lambda.FunctionEventInvokeConfig;
 
   private static readonly encryptionKey = lazy(
@@ -1787,7 +1788,7 @@ export class Function extends Component implements Link.Linkable {
     const logGroup = createLogGroup();
     const zipAsset = createZipAsset();
     const fn = createFunction();
-    const urlEndpoint = createUrl();
+    const { endpoint: urlEndpoint, resource: urlResource } = createUrl();
     createProvisioned();
     const eventInvokeConfig = createEventInvokeConfig();
 
@@ -1797,6 +1798,7 @@ export class Function extends Component implements Link.Linkable {
     this.role = role;
     this.logGroup = logGroup;
     this.urlEndpoint = urlEndpoint;
+    this.urlResource = urlResource;
     this.eventInvokeConfig = eventInvokeConfig;
 
     const buildInput = output({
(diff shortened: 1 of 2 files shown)


davidruhmann:
This seems related to https://github.com/anomalyco/sst/issues/6923

omercnet (approved):
Reviewed the related report in #6923, the complete diff against `dev`, the surrounding Function/SSR permission paths, and the pinned AWS provider behavior.

The dependency graph change is correct: every SST-managed `InvokeFunctionUrl`/`invokedViaFunctionUrl` permission now waits for the exact `lambda.FunctionUrl` resource. That lets the provider finish `CreateFunctionUrlConfig` and its built-in public permission calls before separate permission resources can mutate the same Lambda policy. The server and image-optimizer paths preserve their regional provider selection and permission semantics.

Local verification:
- `bun run typecheck:platform`
- `bun run build:platform`
- focused Function tests: 3/3 passed
- component suite: 89/89 tests passed; three test files still report the known environment-level `ERR_TRACE_EVENTS_UNAVAILABLE` unhandled rejection described in the PR
- virtual merge with current `origin/dev` is clean and Prettier-clean

No correctness, security, compatibility, or dependency-graph blockers found.

Hjorthdx (approved):

PR #6938, opened 2026-07-13
Fix sst remove failing on in-use ACM certificate by ordering load Balancer deletion first.

### LINKED ISSUE #6934: `sst remove` fails deleting the ACM certificate for ALB/Service (missing dependency edge)
## Problem

For components that terminate TLS on a load balancer (`Alb`, `Service`, `Service` v1), `sst remove` fails deleting the ACM certificate with `ResourceInUseException`. The certificate is still `InUseBy` the load balancer's network interfaces, and the load balancer hasn't been deleted yet — removal aborts on the certificate before reaching the load balancer. Manually detaching the LB's network interface unblocks it.

## Cause

The certificate is attached to the `Listener`, but the resource that pins it (via ENIs) is the `LoadBalancer` — which has **no reference to the certificate**. So there's no `LoadBalancer → Certificate` edge, the two are unordered in the graph, and since destroy order is the reverse of that graph, the certificate delete races the load balancer delete.

In `alb.ts`, `createSsl()` returns only `.arn` and discards the component; only the `Listener` consumes it. The `LoadBalancer` is built from `subnets` + `securityGroups` only.

(Not an `.apply()` issue — the ARN carries its dependency fine; the load balancer just never references the cert.)

## Not affected

Other `DnsValidatedCertificate` consumers (apigateway v1/v2/websocket, app-sync, cdn, cognito, https-redirect) attach the cert directly to the resource that pins it (`DomainName` / `Distribution` / `UserPoolDomain`), so the edge already exists.

## Proposed Fix

Surface the cert component from `createSsl()` and make the load balancer depend on it (`alb.ts`; same shape for `service.ts` / `service-v1.ts`):

```ts
function createSsl() {
  if (!domain) return { arn: output(undefined) };
  if (domain.cert) return { arn: output(domain.cert) };
  const cert = new DnsValidatedCertificate(`${name}Ssl`,
    { domainName: domain.name, alternativeNames: domain.aliases, dns: domain.dns! },
    { parent: self });
  return { cert, arn: cert.arn };
}

// load balancer opts:
{ parent: self, dependsOn: cert ? [cert] : [] }
```

## Lightweight Verification of proposed fix

Destroy order = reversed dependency graph, so this is checkable from the graph alone.

Without fix: `lb before cert: false | cert before lb: false` → unordered (race).
With `dependsOn: [cert]` on the load balancer: `lb before cert: true` → load balancer removed first.

--
I don't have the time for a proper PR atm, would be great if someone picked this up.

MalekD5:
closes #6934

## Problem Summary
The load balancer's Elastic Network Interface (ENI) release sometimes delays, and when the ACM certificate delete fires while the ENIs still reference the certificate, you get `ResourceInUseException` and the removal aborts before it ever reaches the load balancer. 

This issue is not deterministic by nature, and most of the time we are at the mercy of AWS-side timings so it not as obvious.

## Fix
The way this PR fixes the problem is by explicitly making the ALB depend on the ACM Certificate. Here is the base config I used for testing:
```typescript

export default $config({
  app(input) {
    return { name: "test-lb-cert", home: "aws", removal: "remove" };
  },
  async run() {
    const vpc = new sst.aws.Vpc("Vpc");
    new sst.aws.Alb("Alb", {
      vpc,
      domain: "lb.sst-test.malekd5.com", // AWS hosted zone
      listeners: [{ port: 443, protocol: "https" }],
    });
  },
});
```

Here is the before and after `sst state export` after running deploy:

**Before**
Before Fix Image

As you can notice, no ACM URNs in dependencies, which means your run's success is timing luck.

**After**
image

As you can notice, ACM URNS in dependency array ordered, which means LB delete must fully complete before ACM delete starts every run. And we are no longer under the mercy of AWS-side timing.

### DIFF
--- a/platform/src/components/aws/alb.ts
+++ b/platform/src/components/aws/alb.ts
@@ -269,7 +269,7 @@ export class Alb extends Component implements Link.Linkable {
     const { vpcId, subnets } = normalizeVpc();
     const domain = normalizeDomain();
     const securityGroup = createSecurityGroup();
-    const certificateArn = createSsl();
+    const { cert, arn: certificateArn } = createSsl();
     const loadBalancer = createLoadBalancer();
     createListeners();
     createDnsRecords();
@@ -347,11 +347,14 @@ export class Alb extends Component implements Link.Linkable {
       );
     }
 
-    function createSsl(): Output<string | undefined> {
-      if (!domain) return output(undefined);
-      if (domain.cert) return output(domain.cert);
+    function createSsl(): {
+      cert?: DnsValidatedCertificate;
+      arn: Output<string | undefined>;
+    } {
+      if (!domain) return { arn: output(undefined) };
+      if (domain.cert) return { arn: output(domain.cert) };
 
-      return new DnsValidatedCertificate(
+      const cert = new DnsValidatedCertificate(
         `${name}Ssl`,
         {
           domainName: domain.name,
@@ -359,7 +362,8 @@ export class Alb extends Component implements Link.Linkable {
           dns: domain.dns!,
         },
         { parent: self },
-      ).arn;
+      );
+      return { cert, arn: cert.arn };
     }
 
     function createLoadBalancer() {
@@ -375,7 +379,7 @@ export class Alb extends Component implements Link.Linkable {
             enableCrossZoneLoadBalancing: true,
 
           },
-          { parent: self },
+          { parent: self, dependsOn: cert ? [cert] : [] },
         ),
       );
     }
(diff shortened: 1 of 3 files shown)


omercnet (approved):
Verified the fix across Alb, Service, and Service v1. The explicit load-balancer dependency expands through DnsValidatedCertificate to the ACM certificate and validation resources, so Pulumi's reverse destroy order removes the listener and load balancer before deleting the certificate. Both the direct component dependency used by Alb and the Output-wrapped component dependency used by Service were exercised with Pulumi mock-monitor graph probes. Platform typecheck and build pass. The existing focused Vitest files could not collect because platform/src/global.d.ts imports @types/node as a runtime package; that failure is pre-existing and unrelated to this diff.

PR #6945, opened 2026-07-15
Fix service dev commands with sidecars (Issue #6944)

### LINKED ISSUE #6944: Missing `dev` command in sidecar container causes Mosaic Panic
Hi!

Came across a weird one today. I am running a `Service` with a Datadog sidecar container. I have a toplevel `dev` command configured to run my service locally. However adding the sidecar (which does not have a `dev` command, its only the datadog agent) causes Mosaic to panic since its not able to expand the command on the sidecar:

```
panic: runtime error: slice bounds out of range [1:0]

goroutine 1 [running]:
main.CmdMosaic(0x14000622100)
/home/runner/work/sst/sst/cmd/sst/mosaic.go:120 +0x2740
github.com/sst/sst/v3/cmd/sst/cli.(*Cli).Run(0x14000622100)
/home/runner/work/sst/sst/cmd/sst/cli/cli.go:90 +0x110
main.run()
/home/runner/work/sst/sst/cmd/sst/main.go:147 +0x370
main.main()
/home/runner/work/sst/sst/cmd/sst/main.go:70 +0x4d8
```

Reporting this issue along with my PR to fix it.

mkilp:
Hi!

This PR closes a Bug that causes Mosaic to panic when running a `Service` with sidecar containers that do not have their own `dev` command section set. 

Added Unit tests for the CLI and the platform. Validated the fix against my own stages where I encountered the bug.

LMK if you need any changes!

cheers
marv

Closes #6944

### DIFF
--- a/cmd/sst/mosaic.go
+++ b/cmd/sst/mosaic.go
@@ -111,9 +111,12 @@ func CmdMosaic(c *cli.Cli) error {
 						<-processExited
 						fmt.Println("\n[restarting]")
 					}
-					fields, _ := shellquote.Split(nextEnv.Command)
-					if len(args) > 0 {
-						fields = args
+					fields := args
+					if len(fields) == 0 {
+						fields, err = parseCommand(nextEnv.Command)
+						if err != nil {
+							return err
+						}
 					}
 					cmd = process.Command(
 						fields[0],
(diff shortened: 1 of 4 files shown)


omercnet (changes requested):
Requesting changes for two service-level dev command regressions:

1. `platform/src/components/aws/service.ts:2805-2842`: when `containers` has one entry and both the service-level `dev.command` and that container's `dev.command` are configured, this creates two root `sst:sst:DevCommand` resources named `${name}Dev`. The new block uses `${name}Dev`, and the container path also selects `title = name` when `val.length == 1`. A focused probe records `sst:sst:DevCommand:DuplicateDevServiceDev` twice, which produces a duplicate Pulumi URN in a real run. Please define precedence between these two configurations or give the container command a unique logical name, and add coverage for a one-container `containers` array with both dev configurations.

2. `platform/src/components/aws/service.ts:2812-2815`: the new service-level `DevCommand` passes `args.dev` without a `title`. `ServiceArgs.dev` cannot supply one, so Mosaic falls back to the resource name `${name}Dev`; the existing single-container path explicitly uses the service name. Please pass `dev: { title: name, ...args.dev }` (or equivalent) and assert the resulting dev metadata title so this path keeps the established pane title.

PR #6965, opened 2026-08-10
Make local uv dependencies hermetic in Python container builds

### LINKED ISSUE #6964: Python container deployments fail when a uv local dependency is not directly in the workspace root
## Problem

Currently, deploying a Python function with `python.container: true` from a uv workspace whose local dependency is not immediately inside the workspace root, breaks. For example, a workspace member declared as `../../lib` or `packages/common` causes SST to export that path into `requirements.txt`. The generated Docker image cannot resolve it and fails at:

```text
RUN uv pip install -r requirements.txt --target ${LAMBDA_TASK_ROOT} --system
exit code: 2
```
uv may also report that the workspace member at `/var/task/../../lib` is missing its `pyproject.toml`.

## Impact

This behavior was introduced by #6061 and shipped in v4.13.1.

I believe #6928 is a specific instance of this broader issue.

The same underlying assumption also affects ordinary local workspace dependencies without relative paths when a custom Dockerfile installs `requirements.txt` before the dependency source directories have been copied. So I believe fixing only parent-path as it does in #6929 does not fully remove the artifact-layout coupling.

## Expected behavior

Container artifacts should be self-contained regardless of whether a local uv dependency is below, above, or otherwise outside the SST project directory. The container build should not need to recreate the developer's workspace layout.

## Proposed direction

Build local Python dependencies as PEP 517 source distributions, place them inside the generated artifact, and rewrite local requirement paths to those archives. Docker can then install the archives inside the Lambda Linux image, where native extensions are built for the target platform.

achadha235:
Fixes #6964 and #6928 

## Problem 

Python container artifacts previously preserved local paths from `uv export`. Those paths are valid in the source checkout but not necessarily in the Docker build context. This includes the `../../member` cases reported in #6928 and custom Dockerfiles that install local dependencies before their source is available. This also affects custom Dockerfiles that copy the full SST artifact before installation. Since the artifact is not a copy of the original uv workspace layout, running `uv sync` or regenerating requirements inside the image failed to resolve workspace dependencies.

## Solution 
For container builds, SST now:

- resolves each local requirement in the original workspace;
- builds it as a PEP 517 source distribution under `.sst/packages/`;
- rewrites `requirements.txt` to reference that artifact-local archive; and
- lets uv install the archive inside the Lambda Linux image.

This keeps the generated artifact self-contained and avoids recreating or rewriting the user's uv workspace layout inside Docker.

`materializeContainerRequirements` reads and rewrites the exported requirements file. `rewriteContainerRequirements` resolves and deduplicates local paths while preserving extras and environment markers. `buildContainerSdist` writes each package archive below the artifact root.

## Why this approach 

I believe this is a more robust solution than #6929 because it packages the resolved local dependencies instead of recreating the source workspace inside Docker by copying directories and rewriting relative paths. That keeps the artifact independent of the developers layout and avoids requiring uv to reconstruct the workspace from a partial build context.

## Testing

Tests cover parent-path workspace members, descendant path dependencies, duplicate local dependencies, extras and markers, missing packages, and artifact-path containment.

I've added a test build at my fork [achadha235/sst `yellowbrick-sst-v0.1.0`](https://github.com/achadha235/sst/releases/tag/yellowbrick-sst-v0.1.0) for validating the change against existing Python setups before merge. I've already used this in production with my own project and it seems to be working well. 

@chrislambert @subssn21 I'd appreciate if you could have a look and validate against your own Python SST setups

### DIFF
--- a/pkg/runtime/python/build.go
+++ b/pkg/runtime/python/build.go
@@ -5,6 +5,7 @@ import (
 	"archive/zip"
 	"compress/gzip"
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -78,12 +79,10 @@ func buildDeploy(ctx context.Context, input *runtime.BuildInput, cacheDir string
 		return nil, fmt.Errorf("package discovery: %w", err)
 	}
 
-	var packagesBuilt []string
 	for _, pkg := range localPackages {
 		if err := buildPackage(ctx, input, pkg); err != nil {
 			return nil, fmt.Errorf("build %s: %w", pkg.Name, err)
 		}
-		packagesBuilt = append(packagesBuilt, pkg.Name)
 	}
 
 	if err := installDependenciesForBuild(ctx, input, projectInfo); err != nil {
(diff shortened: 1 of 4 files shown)


achadha235:
hey @vimtor would really appreciate your review on this - thanks!

PR #6967, opened 2026-08-12
APIGatewayV2 construct doesn't actually clear the CORS config when cors: false is set

### LINKED ISSUE #6748: APIGatewayV2 construct doesn't actually clear the CORS config when `cors: false` is set

newmaniese:
AWS treats {} as a bare CORS config; omit corsConfiguration so CORS is fully disabled.

- Fixes #6748
- Supersedes #6749 (simpler approach per review: return `undefined` instead of nested `.apply()`)
- Adds tests that fail on current dev and pass with applied PR.

### DIFF
--- a/platform/src/components/aws/apigatewayv2.ts
+++ b/platform/src/components/aws/apigatewayv2.ts
@@ -775,7 +775,8 @@ export class ApiGatewayV2 extends Component implements Link.Linkable {
 
     function normalizeCors() {
       return output(args.cors).apply((cors) => {
-        if (cors === false) return {};
+        // AWS treats `{}` as an empty CORS config; omit to fully disable.
+        if (cors === false) return undefined;
 
         const defaultCors: types.input.apigatewayv2.ApiCorsConfiguration = {
           allowHeaders: ["*"],
@@ -831,6 +832,7 @@ export class ApiGatewayV2 extends Component implements Link.Linkable {
           `${name}Api`,
           {
             protocolType: "HTTP",
+            // @ts-ignore Output includes undefined when cors is false
             corsConfiguration: cors,
           },
           { parent },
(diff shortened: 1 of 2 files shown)


omercnet (approved):
Verified the fix against #6748 and the relevant Pulumi/AWS update paths. An Output resolving to undefined is omitted during Pulumi resource-property serialization, while removing the provider's cors_configuration block invokes DeleteCorsConfiguration, so this handles both new APIs and updates from the prior empty-block state. The focused CORS tests pass, including default, disabled, and custom configurations; platform typecheck and build also pass.

PR #6990, opened 2026-08-28
fix: resolve custom bundle paths in dev

### LINKED ISSUE #6253: `bundle` option breaks handler resolution in dev mode
## Use Case

I use a custom build pipeline (tsdown/rolldown) instead of SST's built-in esbuild because my project includes native binaries that require special bundling treatment (externals, copy steps). The build outputs to a `dist/` directory which I reference via the `bundle` option.

This works perfectly with `sst deploy`, but breaks in `sst dev`.

## Configuration

```typescript
new sst.aws.Function("MyApi", {
  bundle: "./dist",
  handler: "lambda.handler",
});
```

## Error

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at Object.pathToFileURL (node:url:1024:3)
    at .sst/platform/dist/nodejs-runtime/index.js:33:24
```

## Cause

SST's dev mode nodejs-runtime tries to resolve the handler path using logic that expects source files relative to the project root. When `bundle` is provided, this resolution fails and returns `undefined`, which then crashes `pathToFileURL()`.

Deploy mode works because it just zips the bundle directory without path resolution.

## Workaround

```typescript
new sst.aws.Function("MyApi", {
  ...($dev
    ? { handler: "./src/lambda.handler" }
    : { bundle: "./dist", handler: "lambda.handler" }),
});
```

## Suggested Fix

When `bundle` is set, resolve handler paths relative to the bundle directory, not the project root.

omercnet:
## Related Issues

Fixes #6253

## In a Nutshell

- Resolve custom bundles from the project root
- Cover relative and absolute bundle paths

## Description

Custom bundle paths remained relative during dev builds, while the Node runtime used the same path as both its working directory and part of the handler path. Resolve relative bundles against the SST config directory so dev handler loading matches deploy behavior, while preserving absolute bundle paths.

## Must

- [x] Tests
- [ ] Documentation (not applicable)

## Verification

- `go test ./pkg/runtime`
- `bun run test:cli`
- `bun run build:platform`
- `bun run build:cli`

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -118,6 +118,9 @@ func (c *Collection) Build(ctx context.Context, input *BuildInput) (*BuildOutput
 
 	if input.Bundle != "" {
 		out = input.Bundle
+		if !filepath.IsAbs(out) {
+			out = filepath.Join(path.ResolveRootDir(input.CfgPath), out)
+		}
 		result = &BuildOutput{
 			Handler: input.Handler,
 			Errors:  []string{},
--- a/pkg/runtime/runtime_test.go
+++ b/pkg/runtime/runtime_test.go
@@ -55,6 +55,43 @@ func TestBuildInputOut(t *testing.T) {
 	})
 }
 
+func TestCollectionBuildCustomBundle(t *testing.T) {
+	projectDir := t.TempDir()
+	absoluteBundle := filepath.Join(t.TempDir(), "dist")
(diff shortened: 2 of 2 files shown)


omercnet:
CI is blocked pending maintainer approval of the fork workflow run: https://github.com/anomalyco/sst/actions/runs/33208975206

